PR #2341, opened 2015-01-28
Embedded notes

r888888888:
This is a new feature to embed text in the note boxes to remove the need to hover to read the text. It is conditionally activated on a per-post basis with a new boolean flag.

### DIFF
--- a/app/assets/javascripts/notes.js
+++ b/app/assets/javascripts/notes.js
@@ -5,7 +5,7 @@ Danbooru.Note = {
       $inner_border.addClass("note-box-inner-border");
 
       var opacity = 0;
-      if (Danbooru.meta("post-has-embedded-notes") === "true") {
+      if (Danbooru.Note.embed) {
         opacity = 0.9
       } else {
         opacity = 0.5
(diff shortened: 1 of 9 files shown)


ToksT:
The opacity of notes was raised to 90%, which is fine for reading, but I think it should be brought back down when translation mode is turned on or it's difficult to read the original text.

ToksT:
> It is conditionally activated on a per-post basis with a new boolean flag.

It's possible to detect which notes have text that is too big to fit in the note box, so we could automatically disable it for those specific notes rather than asking the user manually. Although, I guess it might look weird to have some notes with embedded text and some not on the same post.

ToksT:
We could center the text (horizontally+vertically) within the note box with css like this:

```css
div.note-box.with-embedded-text div.note-box-inner-border {
  text-align: center;
  display: table-cell;
  vertical-align: middle;
}
```

Seems to work fine in most situations, but when the text is too big for the note as mentioned earlier it will cause the note itself to expand as well. So we may want to do something about these small notes before implementing text centering.

r888888888:
The text could be truncated in the box.

But IMO if a post is translated with embedded notes in mind, the translator will resize the note accordingly to accommodate the text. All the previous tricks with CSS and font styling should still work.

r888888888:
I've incorporated @ToksT ideas and have deployed to testbooru.

r888888888:
The display: table-cell breaks overflow: hidden so when the note is really small the display breaks, although the resizer still behaves correctly. The overflow works correctly if the display is set to block but then you lose vertical centering.

PR #2342, opened 2015-01-31
Responsive css

SundaySalsa:
Hi!
I saw that danbooru engine has mobile version with separate views. It's nice. There is also one other way to make website friendly to devices with small screens, like mobile phones and tablet. It's fluid design and responsiveness in css. This way, it doesn't require anybody to create extra views and most website features are still available even for those devices!
I wrote the css needed to achieve that
Here is how site looks like now:
Screenshot 1
Screenshot 1
Some pages still look strange, though. For example, the ones that use tables for input fields layout. I decided not to touch them because something might depend on them:
Screenshot 1

### DIFF
--- a/app/assets/javascripts/responsive.js
+++ b/app/assets/javascripts/responsive.js
@@ -1,0 +1,8 @@
+// Whoever did the rest of js files, he/she did it in some other way
+// Hope this is not very wrong
+$(document).ready(function() {
+  $("#maintoggle").click(function(){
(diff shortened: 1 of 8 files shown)


r888888888:
Thanks for writing this. I'll probably get around to testing it more fully after I deploy and dekink embedded notes which should happen within the next week.

r888888888:
This has been merged.

Issue #1831, opened 2013-07-05
Make autocomplete support aliased tags


Type-kun:
So that it would either show something like '3dmg' in the suggestions, or show its alias.

Type-kun:
Reopening since there's more demand for it in the forums - see http://danbooru.donmai.us/forum_topics/11547 , and there is a way to implement it with query remaining fast, costing db one additional column, two additional indexes and one delayed job.

tl;dr:

* add post_count to tag_aliases table;
* fill it via delayed job (it's single update statement, I think) or after-update trigger on tags table. Job will be cheaper, probably.
* add index on tag_aliases.post_count
* add index on tag_aliases.antecedent_name with pattern_ops mode
* fill autocomplete list with this query (subject to change and extensive testing, of course):
```sql
SELECT * FROM
(
  (
    SELECT tags.name, tags.post_count, tags.category, null AS antecedent_name FROM "tags"
    WHERE (name LIKE ? ESCAPE E'\\')
    AND (post_count > 0)
    ORDER BY post_count desc LIMIT 10
  )
  UNION ALL
  (
    SELECT tags.name, tags.post_count, tags.category, tag_aliases.antecedent_name FROM "tag_aliases"
    INNER JOIN tags ON tags.name = tag_aliases.consequent_name
    WHERE ((tag_aliases.antecedent_name LIKE ? ESCAPE E'\\'))
    AND ((tags.consequent_name NOT LIKE ? ESCAPE E'\\'))
    AND  (tags.post_count > 0)
    ORDER BY tag_aliases.post_count desc LIMIT 10
  )
) AS unioned_query
ORDER BY post_count desc LIMIT 10
```
, `?` being a placeholder for query string.

PR #2363, opened 2015-04-04
Include aliased tags in autocomplete

ToksT:
Makes it so autocomplete searches aliases too. See https://danbooru.donmai.us/forum_topics/11547 for details.

Related: #1831

What it looks like:
autocomp example


ToksT:
Note: the cached post_count in the tag_aliases table gets updated once per day with a daily maintenance job. This means the post counts (only for the aliased tags) can be up to 24 hours out of date. There are other ways to do it but I don't think them being 24 hours out of date is much of a problem.

Type-kun:
Is there anything more to add to this? Forum topic mentions coloring antecedent name gray - I don't thing that's necessary, but perhaps wrapping it in `<span class='autocomplete-antecedent'>` might be a good idea to allow some customizing.

Type-kun:
Huh. Was that feature omitted for the last deploy, 2.77, or it somehow doesn't work properly?

ToksT:
> Huh. Was that feature omitted for the last deploy, 2.77, or it somehow doesn't work properly?

It wasn't omitted - if you go to the tag aliases api you can see the post count column filled with the default 0 value.

A possibility I can think of is that the function to update all aliases on the site daily timed out. I forgot to add `without_timeout` since it was so fast on my server, but on the production site it might be slow since there are more aliases.  I'll add it now.

r888888888:
What's probably going on is your dropdown is cached locally. If you open up a javascript console and enter this: $.localStorage.removeAll(); it'll probably reset the cache and you'll see the aliases.

Type-kun:
Nope, clearing local storage doesn't help. Does it work for you?

ToksT:
> What's probably going on is your dropdown is cached locally. If you open up a javascript console and enter this: $.localStorage.removeAll(); it'll probably reset the cache and you'll see the aliases.

Oh right, I forgot about that. It's not the only issue though: the post_count column is still unset at 0 in [the tag_aliases table](http://danbooru.donmai.us/tag_aliases.json). So they're not showing up in autocomplete since 0-post tags are excluded.

ToksT:
The daily job still didn't work after the update.

@r888888888 Any idea why it's not running? If you run `TagAlias.update_cached_post_counts_for_all` directly in the rails console is there an error?

Type-kun:
Ping @r888888888. Newly approved aliases are using this feature just fine, it's clearly the daily update job that's failing.

r888888888:
I just ran it manually. It seems to execute very fast since there's less than 10k aliases.

Type-kun:
Huh. How come it fails as a part of daily jobs, then? Could it be that it's actually the `DailyMaintenance.run` procedure that fails at some earlier task and never gets to it? Is there any exception logging for daily jobs? I see that output is set to `/var/log/whenever.log`, does that catch error stream as well?

ToksT:
> Could it be that it's actually the DailyMaintenance.run procedure that fails at some earlier task and never gets to it?

Tag subscriptions and the json tag cache and the stuff before those all seem to run fine, so by process of elimination it would be `ForumSubscription.process_all!` causing the error I guess?

It seems the `forum_subscriptions` table doesn't specify `not null` for any of its fields so maybe there's a null in there somewhere causing an error.

r888888888:
So the alias autocomplete is working now?

Type-kun:
Yep: http://puu.sh/hNiOv.png

ToksT:
> So the alias autocomplete is working now?

It's working but not updating automatically so values will eventually become outdated.

r888888888:
Yeah looks like the forum subscription mailer was erroring out. Fixed in c31b869a01a49ada43c3c1f253d80177bb794ebe

PR #2392, opened 2015-05-21
make og:image canonical according to spec

bitinn:
Currently `og:image` is a relative url, it should be canonical according to spec http://ogp.me/ and I have set it to be the same as `twitter:image:src`.

Discovered this problem while working on my own project which use opengraph data for link preview.

### DIFF
--- a/app/views/posts/show.html.erb
+++ b/app/views/posts/show.html.erb
@@ -149,7 +149,7 @@
   <meta name="always-resize-images" content="<%= CurrentUser.user.always_resize_images? %>">
   <meta property="og:title" content="<%= @post.presenter.humanized_essential_tag_string %> - <%= Danbooru.config.app_name %>">
   <meta property="og:description" content="<%= @post.presenter.humanized_tag_string %>">
-  <meta property="og:image" content="<%= @post.preview_file_url %>">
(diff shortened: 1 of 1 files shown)


app/views/posts/show.html.erb
```diff
@@ -149,7 +149,7 @@
   <meta name="always-resize-images" content="<%= CurrentUser.user.always_resize_images? %>">
   <meta property="og:title" content="<%= @post.presenter.humanized_essential_tag_string %> - <%= Danbooru.config.app_name %>">
   <meta property="og:description" content="<%= @post.presenter.humanized_tag_string %>">
-  <meta property="og:image" content="<%= @post.preview_file_url %>">
+  <meta property="og:image" content="http://<%= Danbooru.config.hostname %><%= @post.large_file_url %>">
```
r888888888: This should be just file_url.

bitinn: Is `file_url` the original file? It may cause some problem if image is over certain size, say twitter, which limit it to 1MB.

Do you want to change `twitter:image:src` to `file_url` as well? https://github.com/r888888888/danbooru/blob/master/app/views/posts/show.html.erb#L157

r888888888: No, you're right.

Issue #635, opened 2013-02-25
Favorite groups / Private pools


RaisingK:
Some users use the "Favorites" feature to track, say, all the posts they've downloaded, posts to translate, or something other than the stated purpose.  The Pointless Pools thread has seen quite a few cases of pools that were essentially themed-extensions of Favorites.

How about a private pool ("bookmarked posts"?) feature?  You could put a link under Options on post pages that pops up a window like the "Add to pool" link with the ability to add the post to a new/existing ppool or remove it from one.  And you could list/access them all from your user page.

They don't have to be refreshed like subscriptions and they don't factor in to things like favcount, so you could give at least a few to all levels, without the limits applied to subs/favs.

kxz:
There should be some indication of who owns a favorite group on the group's page and the overall listing.

PR #2416, opened 2015-06-25
Favorite groups

ToksT:
implements #635

* Basic members can have 3 groups, gold can have 5, platinum can have 10 (arbitrarily decided)
* Each favorite group can have up to 10,000 posts (arbitrarily decided)
* Favorite groups can be viewed by anyone but only edited/deleted by their creator
* Can add posts to favgroups like pools: either edit the list of ids or type favgroup:[name or id] in the post's tags
* There is a favgroup:[id] search metatag to allow advanced searches. This is implemented with `posts.id IN`, which I would have expected to be slow for large favgroups, but is surprisingly fast. Not slow even with thousands of posts in the favgroup.
* Can browse through favgroups like pools
* There's a link to your own favgroups in the subnavbar, and to other people's via their profile


r888888888:
Only concern is this table blowing up in size but it probably won't be as bad as favorites. For that reason I think being more aggressive with the limiting is probably a good idea, can always raise them later if usage isn't high.

PR #2595, opened 2016-05-03
Fixing IQDB removal after post expunge

### LINKED ISSUE #2594: IQDB index stays the same after expunge
When you expunge a post it's not being removed from IQDB index despite that this functional is written in code.

The problem is in ```remove_iqdb_async```:
```ruby
File.exists?(preview_file_path)
```
This is being called after the file was deleted, so the code after this line never runs.

Fixed by #2595

yurijmi:
Fixes #2594 

Also adds a script for removing expunged posts from IQDB index.

### DIFF
--- a/app/logical/iqdb/remover.rb
+++ b/app/logical/iqdb/remover.rb
@@ -1,0 +1,18 @@
+module Iqdb
+  class Remover
+    def remove!
+      post_ids = []
+
+      IO.popen("iqdb list #{Danbooru.config.iqdb_file}", "w+") do |io|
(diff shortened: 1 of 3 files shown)


r888888888:
I've tried running iqdb list on the server and I think the database at this point is simply too large to scan everything like that. Probably a simpler idea is to just find the post ids that don't exist int he database and delete those from iqdb. Would result in fewer queries also.

r888888888:
alternative script in 27f53fa5db015ca47680e1dda26ad1ef04b27e73

Issue #2644, opened 2016-08-28
Add privileges to user search


Type-kun:
After thinking about last part of #2641, I realized there's no way to list all users with any permission, not just supervoters. It makes sense to include permissions in [user search](https://danbooru.donmai.us/users/search) form. Postgres can handle bitwise operations if permission integer is [cast to bit string](https://www.postgresql.org/docs/9.5/static/functions-bitstring.html), though it looks like built-in indexes don't help with them. Knowing the bit count (or at least the upper bound) is required for that, but it's as simple as getting size of `BOOLEAN_ATTRIBUTES` in user model. Perhaps it's even possible to pack this into ActiveRecord extension along with `has_bit_flags`, but I'm not well versed in Ruby enough to know how to do that.

As far as UI goes, I see two options. 
1. Create a checkbox for every permission. Checking it would search only for users having the permission, un-checking would ignore the permission.
2. Create a select element for every permission, each with three options - blank, set, not set. This would allow to explicitly search for users not having the permission, also would be a bit more intuitive, but requires a bit more work, checking both for set and unset bits.

Also, since I (to my shame) had to google how to perform bit checks, here's a reminder:
Checking if bits B are set in bitstring A: `A & B == B`
Checking if bits B are unset in bitstring A: `A & B == 0`

r888888888:
The users table is small enough and this type of query infrequent enough that the sequential scan is acceptable. Worst case, the query can be offloaded to the replica database.

Type-kun:
I expanded `search` method in user model to support `can_approve_posts`, `can_upload_free` , `is_super_voter` parameters as `true` or `false`.  It's easy to add other binary options when necessary, too. Seems to work fine, but I'm concerned about performance, so maybe deploy it without changing search form first?

Type-kun:
Hmm. First search with new parameters executed about 15 seconds but didn't fail I guess that's about 8+8 seconds on count and search itself - lucky considering the timeout is 9. Subsequent searches executed within 3 seconds because if disc cache. Looks like it's impossible to easily speed this up (aside from writing GiST index in C from scratch - I'm certainly not up to that now; or creating a FBI for every interesting bit and checking them separately, which seems plain awful). It should do well serving as additional search parameter, at least, so I guess it's safe to add parameters to the form. I'll go with `<select>` elements, since I added possibility to exclude permissions too.

r888888888:
Yeah no good way to index that.

[...]

PR #2678, opened 2016-09-17
Index and form for bit preferences

Type-kun:
See #2644 for details.

Array index failed to work with `not (column && ARRAY[])` syntax, so excluded bits are searched almost the same way as before, except omitting the unnecessary bitstring conversion - turns out, postgres can `&` bigints just fine. 

It's possible to change sql function to create array with negative integers for unset bits and positive for set bits. It would allow complete index support, but storage requirements skyrocket and it needs to know max bit count, so I deemed it wasn't worth it.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -762,7 +762,7 @@ class User < ActiveRecord::Base
       bitprefs_include = nil
       bitprefs_exclude = nil
 
-      [:can_approve_posts, :can_upload_free, :is_super_voter].each do |x|
+      [:can_approve_posts, :can_upload_free, :is_super_voter, :is_banned].each do |x|
         if params[x].present?
           attr_idx = BOOLEAN_ATTRIBUTES.index(x.to_s)
           if params[x] == "true"
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -759,7 +759,6 @@ class User < ActiveRecord::Base
         q = q.where("id in (?)", params[:id].split(",").map(&:to_i))
       end
 
-      bitprefs_length = BOOLEAN_ATTRIBUTES.length
       bitprefs_include = nil
       bitprefs_exclude = nil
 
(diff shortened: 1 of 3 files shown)


Type-kun:
After thinking it over, it is possible to index both set and unset bits, but with some limitations. The function can be changed so that only necessary bits are included in the array, positive for set, negative for unset. So, the function definition would look like this:

`bit_position_array(x bigint, positions integer[])`

Then function `user_prefs_indexed_bits(bigint x)` would be created to call `bit_position_array(x, {0, 13, 14, 16})` - it would always return array with 4 integers. The index be will created using `user_prefs_indexed_bits` instead, and it will check both included and excluded bits using `@>` operation.  It should win both in index size and speed. The only major flaw is loss of flexibility -  if more preferences are ever added to search, `user_prefs_indexed_bits` has to be changed accordingly and the index should be rebuilt.

r888888888:
The migration failed:

```
PG::InvalidArgumentForLog: ERROR:  cannot take logarithm of zero
CONTEXT:  SQL function "bit_position_array" statement 1
:       CREATE INDEX index_users_on_bit_prefs_array
       ON users USING gin (bit_position_array(bit_prefs) _int4_ops)
```

I've reverted the PR for now.

Type-kun:
Ah, damn it, zeros. Can I amend the same DB migration, or should I create a new one?

Type-kun:
The function should look like this instead:
```
select 
    case when x > 0 then
      (select array_agg(i) 
        from generate_series(0, floor(log(2, x))::integer) i
       where (x & (1::bigint << i)) > 0)
    else
       ARRAY[]::integer[]
    end
```

It looks like I will have to create a new branch and cherry-pick the commits, since this one is already merged and doesn't allow pull requests. The question about DB migration file still remains, will it work if the file is named the same?

r888888888:
You should be able to create a new pull request that reverts the revert (fcfc29ca6f4f09c25e090b3656fce4cb5f780e0a i believe) and then apply your changes on top of that.

PR #2737, opened 2016-10-25
Link self-upload disclaimer on /uploads; link howto:flag in flag notices.

evazion:
* Add self-uploading disclaimer to uploads page.
* Link to howto:flag in flag notice.


evazion:
Just noticed a typo: "More suitable sites to uploading your own artwork include..."

PR #2753, opened 2016-11-05
Autotag character_(cosplay) -> character, cosplay.

evazion:
Make e.g. `hatsune_miku_(cosplay)` automatically add `hatsune_miku` and `cosplay`.

Perform autotagging after aliasing so that `*_(cosplay)` aliases (e.g. `hestia_(dungeon)_(cosplay)` -> `hestia_(danmachi)_(cosplay)`) can take effect first, just in case that matters (honestly, it probably doesn't, but whatever).

ref: http://danbooru.donmai.us/forum_topics/13367


Type-kun:
Hm, do we have any tags that don't match like that?

evazion:
Not sure. Skimming through [\*\_(\*)\_(cosplay)](http://danbooru.donmai.us/tags?search[order]=count&search[name_matches]=*_(*)_(cosplay)) I don't see any. Not any with more than 5 tags anyway.

PR #2764, opened 2016-11-20
Use raw note body instead of sanitized body for note edit box.

evazion:
Problem: Create a note with body `<i>foo&shy;bar`. Save it, refresh the page, and edit the note. Now you see `<i>foobar</i>` in the edit box even though that is not the actual note body.

The issue is the sanitized body is used instead of the raw body for the edit box. This prevents translators from being able to see and edit the unsanitized note text. So this patch just switches to using the raw note instead.


app/views/notes/_note.html.erb
```diff
@@ -1 +1 @@
-<article data-width="<%= note.width %>" data-height="<%= note.height %>" data-x="<%= note.x %>" data-y="<%= note.y %>" data-id="<%= note.id %>"><%= raw DText.sanitize(note.body) %></article>
+<article data-width="<%= note.width %>" data-height="<%= note.height %>" data-x="<%= note.x %>" data-y="<%= note.y %>" data-id="<%= note.id %>" data-body="<%= note.body %>"><%= raw DText.sanitize(note.body) %></article>
```
r888888888: I'm pretty sure you need to escape note.body here. What if it contains a quotation? I believe as is it'll be rendered as &quot;

evazion: I tested it and as far as I can tell it works. Rails escapes `"lol"` as `&quot;lol&quot;` for the data-body attribute, but the browser unescapes that back to `"lol"` so it still works.

PR #2821, opened 2017-01-06
API limit improvements

evazion:
Fixes a couple bugs and includes API limits in HTTP headers and on your user profile.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -41,6 +41,7 @@ protected
   
   def api_check
     if request.format.to_s =~ /\/json|\/xml/ || params[:controller] == "iqdb"
+      response.headers.merge!(ApiLimiter.limits)
       if ApiLimiter.throttled?(request.request_method)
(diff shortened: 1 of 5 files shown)


BrokenEagle:
Just confirmed my above theory... after switching to Sonohara, it kept going until it 429'd at the 5000 mark.  Then I switched to Hijiribe, and it was able to keep on going.  

To double confirm, I opened browser tabs to both Sonohara and Hijiribe, and tag scripting was blocked on Sonohara (since tag scripting uses the API), but was still allowed on Hijiribe...

evazion:
Yeah, the request counts are stored in memcache, and each server runs its own memcached instance. So they would have to be synchronized. There are already several other places where caches have to be synchronized between servers (updating tag counts and tag aliases are a couple examples), but they do so by creating delayed jobs that run on each server, which means a database write to create the job, which defeats the purpose of using memcache. A db write for each API read wouldn't be very good for performance.

So I dunno how to fix it. Switching to a replicated cache setup with redis or maybe repcached could simplify a lot of cache synchronization issues, but I have no real experience with either.

PR #2840, opened 2017-01-14
Fix #2839: "Original" tag shows up twice in Pixiv source data

### LINKED ISSUE #2839: "Original" tag shows up twice in Pixiv source data
When fetching a post’s source from Pixiv, the `オリジナル` tag shows up twice when it should be there only once.

For example, [post #2599024](https://danbooru.donmai.us/posts/2599024) has these tags on Pixiv:
`漫画   4コマ   オリジナル   おっぱいのついたイケメン`

Clicking “Fetch source data” on Danbooru yields these tags:
`漫画   4コマ   オリジナル   おっぱいのついたイケメン   オリジナル`

So far, I’ve only seen it happen with the オリジナル tag and the extra tag was always last.

evazion:
The root cause is [here](https://github.com/r888888888/danbooru/blob/3ad639521ff70db4198be0d1e44f808503c102c6/app/logical/sources/strategies/pixiv.rb#L314-L316). Pixiv posts can be in an "Original" category, so we fake up a tag in that case, but that can create a duplicate. I deemed it better to uniq the tags for all sources.

### DIFF
--- a/app/logical/sources/strategies/base.rb
+++ b/app/logical/sources/strategies/base.rb
@@ -83,7 +83,7 @@ module Sources
       end
 
       def tags
-        @tags || []
+        @tags.uniq || []
       end
 
       # Should be set to a url for sites that prevent hotlinking, or left nil for sites that don't.


Type-kun:
Can `@tags ` be nil? Judging by precautionary `|| []` it can, and `uniq` will probably throw an error. Better to move `uniq` a line down.

evazion:
Good catch, will fix it later.

Issue #2824, opened 2017-01-08
Bugs for mod actions on post/comment update/delete


Type-kun:
Attempting to edit another user's forum post results in the following message:

> undefined local variable or method `rec' for #<ForumPost:0x007f25e5c59290>
> app/models/forum_post.rb:23:in `block in <class:ForumPost>'
> app/controllers/forum_posts_controller.rb:50:in `update'

It's due to `rec` being undefined outside of lambda expression.

Type-kun:
Mostly fixed, but deleting a comment causes a following message to be written:

> comment #... updated by ...

instead of "deleted by". Not sure why.

evazion:
I think the `after_destroy` callbacks never actually run, since deleting a comment just sets the `is_deleted` flag instead of destroying the object.

Type-kun:
Hm, true. Same happens to forum posts now that I have checked.

r888888888:
fixed

PR #2841, opened 2017-01-14
Revert to: "Another partial fix for issue #2824, which also affected …

evazion:
ref: http://danbooru.donmai.us/forum_topics/9127?page=152#forum_post_125635

Partially reverts f2a5d45db091e04e696f959c2590614c513304e4 to 9da5e67a65f29c84344ecf4fac03834ed61d80f7.


Type-kun:
It's might be better to use `|rec|` as `after_update` block argument too, to not rely on model fields, but the `updater_id` doesn't exist on user_feedback, and link to user is worth keeping :3

Issue #2863, opened 2017-01-27
Upload tag change report


Type-kun:
There was a discussion in [mod chat](https://safebooru.donmai.us/forum_topics/13341?page=3#forum_post_126076) about a tool for monitoring given user's tagging habits, to see what they added to their uploads and what they haven't. I proposed the following:

1) Take user's uploads, order by ID descending; show ID and maybe thumbnail in the first column
2) For each upload, traverse versions, see what tags were added by user, merge them in a single record, and show it as a second column
3) Take current version, diff against the second column we just calculated, and show the result (tags added and removed by other users) as a third column.

In addition to being useful to moderators, it might make for a great tag discovery tool, and also allow users to relatively easily identify posts reported by danboorubot.

r888888888:
Are there still any outstanding issues with this?

Type-kun:
It's not linked anywhere yet. The link should either go near profile upload count (`Uploads 7669 (Tag Change Report)`) or in profile subnavbar. 

Other than that, it seems to be loading a tad bit slowly, probably because of usual N+1 issue, but it's not that much of the problem so far.

evazion:
#2947 fixes a couple N+1 queries problems. There's one left in that each time `changes` is called, it loads the previous version to calculate the rating/source/parent changes. Archives would need `rating` / `previous_rating` fields instead of `rating` / `rating_changed` to fix this.

Type-kun:
>There's one left in that each time  changes  is called, it loads the previous version to calculate the rating/source/parent changes.

Hm. Can't rails make use of versions preloaded with `includes(versions: [:post])`? It re-loads already loaded rows after all. Or, it would be uglier, but we can expose a method in `version` model to calculate changes to another version given as parameter, then call that method both from `changes` and from report.

evazion:
> Hm. Can't rails make use of versions preloaded with includes(versions: [:post])?

Well, it hadn't been because the way `PostArchive#previous` was defined, it didn't use associations so preloading didn't help it. #2948 fixes it, but it's pretty hacky and subtle.

PSA: I'm using the RailsPanel chrome extension for finding N+1 problems. It makes seeing what's happening on the SQL level much easier. I have it running on http://devbooru.evazion.ml:3000/, so if you have the extension installed you can use it there.

PR #2864, opened 2017-01-27
Implement #2863: basic upload tags change report

Type-kun:
Quick implementation for  #2863. This is a naive approach: I created a new read-only `UploadTagReport` model by extending `Post` and adding methods for comparing versions. This allows for easy pagination, but I have a sneaking suspicion it doesn't really fit Rails' MVC paradigm, and it should have gone into `logical` instead, with methods like `search` to return collection of some other class having `Post` as an instance variable; but I was completely baffled about applying pagination to such things. @r888888888 ?

Ideally, to make the report more useful, there should also be a way to exclude posts where all tags were added by uploader. This fits numbered pagination even less, unfortunately, and I'm not familiar enough with sequential paginator to make it work on something like that. Besides, it's too hard to determine the condition at SQL level, meaning it'll somehow have to apply pagination AFTER loading enough rows. Unfortunately, I have no idea how to do this, either.

### DIFF
--- a/app/controllers/upload_tags_report_controller.rb
+++ b/app/controllers/upload_tags_report_controller.rb
@@ -1,0 +1,9 @@
+class UploadTagsReportController < ApplicationController
+  respond_to :html, :xml, :json
+
+  def show
+    @user = User.find(params[:id])
+    @upload_reports = UploadTagsReport.for_user(params[:id]).order("id desc").paginate(params[:page], :limit => params[:limit])
+    respond_with(@upload_reports)
+  end
+end
--- a/app/helpers/upload_tags_report_helper.rb
+++ b/app/helpers/upload_tags_report_helper.rb
@@ -1,0 +1,12 @@
+module UploadTagsReportHelper
+  def diff_to_current(report)
+    html = []
+    report.added_tags_array.each do |tag|
+      html << '<ins>+' + link_to(wordbreakify(tag), posts_path(:tags => tag)) + '</ins>'
+    end
+    report.removed_tags_array.each do |tag|
+      html << '<del>-' + link_to(wordbreakify(tag), posts_path(:tags => tag)) + '</del>'
+    end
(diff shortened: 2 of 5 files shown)


Type-kun:
Also missing: any links to this report. I'm not sure where to put them, since it requires a user ID. Also, the name itself is too bulky.

evazion:
> Ideally, to make the report more useful, there should also be a way to exclude posts where all tags were added by uploader.

I think this would do it:

```ruby
# uploads where there is an edit from someone other than the uploader.
uploader.posts.where("EXISTS (SELECT 1 FROM post_versions WHERE post_id = posts.id and updater_id != ?)", uploader.id)
```

This won't work if/when `post_versions` is moved to a separate database for the archives service though. 

> Besides, it's too hard to determine the condition at SQL level, meaning it'll somehow have to apply pagination AFTER loading enough rows. 

Well, you could filter out such posts in the view after they're loaded. That's what's done in other places that can't do their filtering in SQL. It does lead to incomplete/empty pages though (see: `order:score_asc`, or `loli` while logged out).

Type-kun:
>I think this would do it:

Hm, this wouldn't work for the cases when there's a version which was reverted later, but would cover most cases. Though archive service makes this impossible to do, yes.

>Well, you could filter out such posts in the view after they're loaded. 

Would probably be pretty useless in this case. I imagine someone would use this option when there's not many uploads where tags were changed, leading to almost empty pages. Technically, we have sequential paginator module that tells your model to load N records after a certain ID. Maybe ActiveRecord model can be coded to somehow fetch and filter records from Relation to get enough rows for `limit`?

r888888888:
Shouldn't this be linked from the user profile?

Type-kun:
@r888888888 yeah, it certainly should, I just didn't bother to implement this further until the main part was given the OK. I wonder where would be the best place to put it, though.

PR #2877, opened 2017-02-04
Fix remaining test failures

evazion:
This fixes the test suite so that it should pass cleanly now. See commits for details. Also it cleans out some cruft from `test_helper.rb`.

98b0b2c bears some explanation. Basically, ever since VCR was removed I've been getting random connection reset errors coming from the mechanize scraper for pixiv. From what I understand, mechanize doesn't cope well with the server unexpectedly closing a keep alive connection. This just disables keep alive entirely, which is not great but was the only thing that worked for me.

ref:

http://www.rubydoc.info/gems/mechanize/Mechanize#retry_change_requests%3D-instance_method
github.com/sparklemotion/mechanize/issues/123


evazion:
Updated to also update all testing gems to their latest versions.

PR #2958, opened 2017-04-03
Fix #2955: Reapproval error for unapproved post 

### LINKED ISSUE #2955: Reapproval error for unapproved post
From https://danbooru.donmai.us/forum_topics/9127?page=170#forum_post_129179:

> Attempting to approve post #2679797 gives me the error that I've already approved it... kind of strange, since it's never been approved (blue queue).

Apparently the post threw a `You have previously approved this post` error even though the post is not approved (status is still pending and approver_id is null). This could only happen if a `PostApproval` record exists, which isn't supposed to be possible when the post hasn't yet been approved.

I think the problem is that `Post#approve!` doesn't run inside a transaction, so if the `PostApproval` creation succeeds but updating the post itself fails (due to a timeout), then the `PostApproval` won't be rolled back.

evazion:
Refactors `Post#approve!` to move all the approval validation and post updating logic to `PostApproval`. This way, when saying `post.approvals.create(user: approver)`, the approval becomes atomic, since rails runs callbacks inside transactions automatically. This should fix #2955. 

(wrapping `approve!` in a `Post.transaction` block would have sufficed, but I figured moving the approval code to to `PostApproval` would clean up `Post` a bit).


evazion:
Rebased, force pushed.

PR #2989, opened 2017-04-23
Default /notes to listing notes, not posts.

evazion:
Discussion: https://danbooru.donmai.us/forum_topics/13946.

This would default the `/notes` index to showing notes instead of posts. The `group_by=post` view is removed and replaced with a link to `order:note`. `order:note` is more usable anyway, since it has the sidebar.

Removing `group_by=post` would be a breaking change to the API, but I doubt anyone actually uses it.

Finally, this fixes a couple N+1 issues and switches the `/notes` and `/note_versions` listings to the numbered paginator. I don't see any performance issues with the numbered paginator so the sequential paginator doesn't appear necessary.


r888888888:
Interesting. I figured the notes page would be more popular but it doesn't show up at all in analytics.

PR #3009, opened 2017-04-29
Fix #1868: Dmail search improvements: subject, searchbox

### LINKED ISSUE #1868: Dmail search improvements: subject, searchbox
- Ability to search subject/title only, not just the body
- Search box in subnavbar--message search or subject+message.

evazion:
Fixes #1868. Adds an option for searching dmails by subject line.

Also inlines the search form on `/dmails` instead of it being a separate page at `/dmails/search`. This is more convenient and is the style used in most other places.


r888888888:
title isn't indexed but given that the scope is limited to one user i don't think it will cause problems.

PR #3017, opened 2017-05-03
Fix links to DeviantArt sources

evazion:
Discussion: https://danbooru.donmai.us/forum_topics/9127?page=179#forum_post_130592:

Current DeviantArt sources as linked as http://misumi_illustration.deviantart.com/gallery/#/db7ibsi in the sidebar. This relies on a javascript redirect on DeviantArt's end which doesn't work for everyone. This links the source as http://misumi-illustration.deviantart.com/art/Lethargic-friends-677740338 instead.

The comments in 1baffb3 are just documentation, the only change in the regexes was adding a `prnt` subdomain.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -387,11 +387,27 @@ class Post < ActiveRecord::Base
         base_36_id = base_10_id.to_s(36)
         "http://twitpic.com/#{base_36_id}"
 
-      when %r{\Ahttps?://(?:fc|th|pre|orig|img)\d{2}\.deviantart\.net/.+/[a-z0-9_]*_by_([a-z0-9_]+)-d([a-z0-9]+)\.}i
-        "http://#{$1}.deviantart.com/gallery/#/d#{$2}"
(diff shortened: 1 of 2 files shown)


BrokenEagle:
Is DeviantArt consistent with how they format the artist and title in their image links so that the transforms that you apply will always work...?

evazion:
1baffb31ff04f8bb12e3af6c71dd0e88a3dd5041 lists as many distinct link formats as I could find. AFAICT the general format is either `${title}_by_${artist}-d${base36_id}` or `${md5}-d${base36_id}`. Neither the artist nor the title seem to matter for the redirect, you can say http://blah.deviantart.com/art/blah-677740338 and it still works. So even if we don't format the title exactly the same way dA does, the link should still work.

In cases where the artist or the base36 id aren't present, the url is left as-is.

BrokenEagle:
Ah, good, I was considering running some kind of script to do a random sampling of images to determine if they matched the above post/image scheme.  But as long as only the last portion of the URL matters (deviant ID), then it should be good to go.

PR #3067, opened 2017-05-19
Fix #3065: <div> id incorrect for certain controllers

### LINKED ISSUE #3065: <div> id incorrect for certain controllers
I was trying to develop some CSS for various controllers, and I noticed that some were using the incorrect <div> id.  The ones I could find like this were **wiki_page_versions** and **pool_versions**, which used **c-wiki-pages** and **c-pools** instead of **c-wiki-page-versions** and **c-pool-versions**.

Changing the above might also require updating the CSS since the current version pages might be relying on `#c-pools` or `#c-wiki-pages`.

evazion:
Fixes #3065.


BrokenEagle:
Made a note on issue #3065, but there are a few more pages that need fixing.  I'm not sure I've gotten everything yet, so I'll probably do a full examination hopefully be end of tomorrow.

evazion:
The upload tags page was `#c-post-versions` so it could use the `<ins>`/`<del>` CSS that is used by the post versions and artist versions pages. I refactored everything to use a `.diff-list` class instead.

PR #3082, opened 2017-05-23
Add `data-is-favorited` attribute to post thumbnails.

evazion:
Adds a `data-is-favorited` attribute to post thumbnails. This is to make it possible to use custom CSS to mark posts that you have already favorited.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -954,7 +954,7 @@ class Post < ActiveRecord::Base
     end
 
     def favorited_by?(user_id)
-      fav_string =~ /(?:\A| )fav:#{user_id}(?:\Z| )/
+      fav_string.split.include?("fav:#{user_id}")
     end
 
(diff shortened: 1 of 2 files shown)


app/models/post.rb
```diff
@@ -954,7 +954,7 @@ def clean_fav_string!
     end
 
     def favorited_by?(user_id)
-      fav_string =~ /(?:\A| )fav:#{user_id}(?:\Z| )/
+      fav_string.split.include?("fav:#{user_id}")
```
r888888888: Is this actually faster? I imagine splitting would use up a lot more memory.

evazion: I don't know, I didn't consider the performance. I had wanted to avoid saying `!!(fav_string =~ /(?:\A| )fav:#{user_id}(?:\Z| )/)` to cast it to a boolean, since that's kind of ugly, but I'll change it back.

PR #3092, opened 2017-05-26
Optimize related tags calculation

evazion:
`RelatedTagCalculator#calculate_from_sample` is called when clicking the General / Artist / Copyright / Characters buttons during tag editing. Currently it's slow because it counts up the tags in ruby, and because it makes repeated calls to memcache in the inner loop when filtering tags by category.

This changes it to do the counting and filtering entirely in SQL using a GROUP BY tag / COUNT(*) aggregation. For most tags calculating the related tags takes 200-250ms with a sample size of 300. This also makes it more feasible to increase the sample size.


r888888888:
Woah.

PR #3145, opened 2017-06-12
Add "flagger:" and "appealer:" metatags (fixes #3142)

### LINKED ISSUE #3142: Implement `flag:` and `appeal:` metatags
This would allow to quickly search for post that were flagged or appealed by users, combined with other tags. Looks doable, this shouldn't be much harder than `noteupdater` for example.

Syntax: 
* `flag:<username>`, `flag:any`, `flag:none`
* `appeal:<username>`, `appeal:any`, `appeal:none`

First, searching for the specific usernames. For appeals, it's the same `posts.id IN (?)` with simple `PostAppeal.unscoped.where(creator_id = ...)` would work.  For flags, only non-automatic flags should be selected, so it's probably better to use `PostFlag.unscoped.search({:creator_id => ..., :category = "normal"}).reorder("")` to make use of already coded-in restrictions.

As for `any` and `none`, it's a bit tricky. It can be done with the same `IN` and `NOT IN` queries without specifying `creator_id`, but it should be checked to see if postgres optimizes it properly. Otherwise, it should be possible to do using `where (' EXISTS (' + PostFlag.unscoped.search({category = "normal"}).where('post_id = posts.id').select('1').to_sql + ')')` or something.

Type-kun:
Went with "EXISTS" option. I'm pretty sure it would be faster for "none" cases; "any" cases should be tested on large datasets if possible, to see if `IN` would be better instead.

### DIFF
--- a/app/helpers/post_flags_helper.rb
+++ b/app/helpers/post_flags_helper.rb
@@ -7,8 +7,11 @@ module PostFlagsHelper
       html << '<li>'
       html << DText.parse_inline(flag.reason).html_safe
 
-      if CurrentUser.is_moderator?
-        html << " - #{link_to_user(flag.creator)} (#{link_to_ip(flag.creator_ip_addr)})"
+      if CurrentUser.can_view_flagger?(flag.creator_id)
+        html << " - #{link_to_user(flag.creator)}"
+        if CurrentUser.is_moderator?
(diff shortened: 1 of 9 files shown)


evazion:
There should be a privilege check that `flagger:<name>` can only be used by people with permissions to view flagger names (either Mod+ or people searching their own flags).

PR #3210, opened 2017-07-09
Fix #3208: Fix translated tag suggestions for Pixiv.

### LINKED ISSUE #3208: Fix translated tag suggestions for Pixiv
The combination of #3174, #3033, and #2915 has led to bad translated tags being suggested for Pixiv. Namely, `photoshop` and `sai` are suggested because of #3033 and #3174, but our `photoshop` tag has a very different meaning from Pixiv's. Another example is `Fate/GrandOrder` suggesting `fate`.

Translated tags should work like this:

* First, search undeleted wikis by other names.
* If a match is found but the tag is empty, stop. This way ambiguous tags can be blackholed.
* If no match, search for the tag with the same name.
* If no match, split tag on delimiters and repeat above.

evazion:
Fixes #3208:

* Only suggest the Danbooru tag with the same name if there is no matching wiki other name. Example: if we have the Pixiv tag `Fate` and the Danbooru tag `fate_(series)` with other name `fate`, suggest that, not the Danbooru tag `fate`.

* Don't suggest tags that are empty or whose wiki is deleted.

* Only split tags on `/` if there are no other matches, and only do this for Pixiv.

* For Pixiv, only include traditional media tags in the tag list, not digital media (Photoshop, SAI, etc).

* Add some tests.

### DIFF
--- a/app/logical/pixiv_api_client.rb
+++ b/app/logical/pixiv_api_client.rb
@@ -3,6 +3,21 @@ class PixivApiClient
   CLIENT_ID = "bYGKuGVw91e0NMfPGp44euvGt59s"
   CLIENT_SECRET = "HP3RmkgAmEGro0gn1x9ioawQE8WMfvLXDz3ZqxpK"
 
+  # Tools to not include in the tags list. We don't tag digital media, so
+  # including these results in bad translated tags suggestions.
+  TOOLS_BLACKLIST = %w[
+    Photoshop Illustrator Fireworks Flash Painter PaintShopPro pixiv\ Sketch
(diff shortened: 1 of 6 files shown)


BrokenEagle:
Would it be better to use a tool whitelist instead of a blacklist...?  There would still be an issue of updating the list as required, but I'm guessing that the list of traditional media won't grow as fast as the list of digital media tools.

evazion:
> Would it be better to use a tool whitelist instead of a blacklist...? There would still be an issue of updating the list as required, but I'm guessing that the list of traditional media won't grow as fast as the list of digital media tools.

I wanted to err on the side of allowing new tools by default. If new tools are added, we're more likely to notice it if they're allowed by default. As long as they don't cause problems, which I think is unlikely, nothing needs to be updated.

Issue #1210, opened 2013-04-06
Rapidly favoriting or unfavoriting posts results in inaccurate favorite counts and scores


Sal-N:
http://danbooru.donmai.us/users/337872

No favorite since I removed all, but the favorite count stay the same for a month and won't change to zero. Cache issue?

PR #3244, opened 2017-07-28
Fix #2062: Update user _count fields when expunging posts/wikis

### LINKED ISSUE #2062: Update user _count fields when expunging posts/wikis
The fields `post_upload_count` and `favorite_count` will fail to be decremented when one of the user's uploads/favorites is expunged. This can lead to those fields being higher than they actually should be.

Similarly, `wiki_page_version_count` might become inaccurate if a wiki page is permanently deleted.

evazion:
Fixes #2062: 

* Updates the `post_upload_count`, `post_update_count`, `note_update_count`, and `favorite_count` fields when a post is expunged.

* Removes `Favorite.destroy_all`. AFAICT the comment is inaccurate, deleting favorites from the child table by hand is not necessary.

* Removes some code related to cleaning favorite counts. This was a workaround for #1210 that I think is no longer necessary.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -36,7 +36,7 @@ class Post < ApplicationRecord
 
   belongs_to :updater, :class_name => "User"
   belongs_to :approver, :class_name => "User"
-  belongs_to :uploader, :class_name => "User"
+  belongs_to :uploader, :class_name => "User", :counter_cache => "post_upload_count"
   belongs_to :parent, :class_name => "Post"
   has_one :upload, :dependent => :destroy
   has_one :artist_commentary, :dependent => :destroy
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -1025,7 +1025,9 @@ class Post < ApplicationRecord
     end
 
     def remove_from_favorites
-      Favorite.destroy_all(post_id: self.id)
+      favorites.find_each do |fav|
+        remove_favorite!(fav.user)
+      end
     end
 
     def remove_from_fav_groups
--- a/app/models/favorite.rb
+++ b/app/models/favorite.rb
@@ -4,19 +4,6 @@ class Favorite < ApplicationRecord
   scope :for_user, lambda {|user_id| where("user_id % 100 = #{user_id.to_i % 100} and user_id = #{user_id.to_i}")}
   attr_accessible :user_id, :post_id
 
-  # this is necessary because there's no trigger for deleting favorites
-  def self.destroy_all(user_id: nil, post_id: nil)
-    if user_id && post_id
-      connection.execute("delete from favorites_#{user_id % 100} where user_id = #{user_id} and post_id = #{post_id}")
-    elsif user_id
-      connection.execute("delete from favorites_#{user_id % 100} where user_id = #{user_id}")
-    elsif post_id
-      0.upto(99) do |uid|
-        connection.execute("delete from favorites_#{uid} where post_id = #{post_id}")
-      end
-    end
-  end
-
   def self.add(post:, user:)
     Favorite.transaction do
       User.where(:id => user.id).select("id").lock("FOR UPDATE NOWAIT").first
@@ -40,7 +27,7 @@ class Favorite < ApplicationRecord
       User.where(:id => user.id).select("id").lock("FOR UPDATE NOWAIT").first
 
       return unless Favorite.for_user(user.id).where(:user_id => user.id, :post_id => post_id).exists?
-      Favorite.destroy_all(user_id: user.id, post_id: post_id)
+      Favorite.for_user(user.id).delete_all(post_id: post_id)
       Post.where(:id => post_id).update_all("fav_count = fav_count - 1")
       post.delete_user_from_fav_string(user.id) if post
       User.where(:id => user.id).update_all("favorite_count = favorite_count - 1")
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -147,8 +147,6 @@ class Note < ApplicationRecord
 
   def create_version
     return unless versioned_attributes_changed?
-    User.where(id: CurrentUser.id).update_all("note_update_count = note_update_count + 1")
-    CurrentUser.reload
 
     if merge_version?
       merge_version
--- a/app/models/note_version.rb
+++ b/app/models/note_version.rb
@@ -1,6 +1,6 @@
 class NoteVersion < ApplicationRecord
   before_validation :initialize_updater
-  belongs_to :updater, :class_name => "User"
+  belongs_to :updater, :class_name => "User", :counter_cache => "note_update_count"
   scope :for_user, lambda {|user_id| where("updater_id = ?", user_id)}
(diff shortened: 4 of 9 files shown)


r888888888:
The reason I changed the `dependent: :destroy` on `Favorite` is it was deleting favorites by looking up the id:

```
irb(main):002:0> Post.last.favorites.destroy_all
  Post Load (2.4ms)  SELECT  "posts".* FROM "posts"  ORDER BY "posts"."id" DESC LIMIT 1
  Favorite Load (5.6ms)  SELECT "favorites".* FROM "favorites" WHERE "favorites"."post_id" = $1  [["post_id", 100]]
   (0.1ms)  BEGIN
  SQL (6.8ms)  DELETE FROM "favorites" WHERE "favorites"."id" = $1  [["id", 43]]
   (2.3ms)  COMMIT
```

This is a degenerate query because it'll perform a sequential scan on every table. `favorites.id` is not indexed since it wouldn't have been of much use.

It looks like using `dependent: :delete` will at least use the foreign key so it matches what `Favorite.destroy_all` was doing. This will likely cause the counter cache to not get updated, however.

Postgres still isn't that smart about how to query on `user_id`:

```sql
explain select * from favorites where user_id = 1 limit 10;

 Limit  (cost=0.00..9.66 rows=10 width=12)
   ->  Append  (cost=0.00..17026.56 rows=17618 width=12)
         ->  Seq Scan on favorites  (cost=0.00..0.00 rows=1 width=12)
               Filter: (user_id = 1)
         ->  Index Scan using index_favorites_0_on_user_id on favorites_0  (cost=0.42..185.98 rows=60 width=12)
               Index Cond: (user_id = 1)
         ->  Bitmap Heap Scan on favorites_1  (cost=227.96..2126.60 rows=11811 width=12)
               Recheck Cond: (user_id = 1)
               ->  Bitmap Index Scan on index_favorites_1_on_user_id  (cost=0.00..225.00 rows=11811 width=0)
                     Index Cond: (user_id = 1)
         ->  Index Scan using index_favorites_10_on_user_id on favorites_10  (cost=0.42..134.79 rows=54 width=12)
               Index Cond: (user_id = 1)
         ->  Index Scan using index_favorites_11_on_user_id on favorites_11  (cost=0.42..110.41 rows=53 width=12)
               Index Cond: (user_id = 1)
         ->  Index Scan using index_favorites_12_on_user_id on favorites_12  (cost=0.42..163.70 rows=66 width=12)
               Index Cond: (user_id = 1)
...
etc.
```

The way around this is adding the mod check:

```sql
explain select * from favorites where user_id = 1 and user_id % 100 = 1 limit 10;

 Limit  (cost=0.00..363.79 rows=10 width=12)
   ->  Append  (cost=0.00..2182.71 rows=60 width=12)
         ->  Seq Scan on favorites  (cost=0.00..0.00 rows=1 width=12)
               Filter: ((user_id = 1) AND ((user_id % 100) = 1))
         ->  Bitmap Heap Scan on favorites_1  (cost=225.02..2182.71 rows=59 width=12)
               Recheck Cond: (user_id = 1)
               Filter: ((user_id % 100) = 1)
               ->  Bitmap Index Scan on index_favorites_1_on_user_id  (cost=0.00..225.00 rows=11811 width=0)
                     Index Cond: (user_id = 1)
(9 rows)
```

So I think any reference to `Favorite.destroy_all` has to be changed to `Favorite.delete_all` or equivalent, and the one line in [app/models/favorite.rb:28](https://github.com/r888888888/danbooru/pull/3244/files#diff-c497e09dbda237435869964d9137a197R28) has to be changed to take the user id mod into account. Switching to delete (I assume) means any counter caches have to be updated manually.

evazion:
Rebased. Changed it in 5da8ea8 to just call `remove_favorite!` on every favorite. Removing favs one-by-one like this is not as efficient as removing all of them at once with a `delete_all`, but it ensures `favorite_counts` are updated properly and for expungement purposes, it shouldn't matter much if it's not optimal, it just needs to not time out.

> The reason I changed the dependent: :destroy on Favorite is it was deleting favorites by looking up the id:

> This is a degenerate query because it'll perform a sequential scan on every table. favorites.id is not indexed since it wouldn't have been of much use.

I see, I get it now. But `favorites.id` is used in `ordfav:<name>` searches, and that does become slow for people with sufficiently many favorites. See [ordfav:Claverhouse](http://danbooru.donmai.us/posts?tags=ordfav:Claverhouse) for a worst case scenario. So I think indexing it would be beneficial.

> Postgres still isn't that smart about how to query on user_id:
> The way around this is adding the mod check:

I tested this out and on my installation at least, not doing the mod check is actually faster than doing it, even despite the uglier query plan:

```sql
danbooru2=# select count(*) from favorites;
  count   
----------
 42779818

danbooru2=# \timing
Timing is on.

-- simplified query for `ordfav:Claverhouse limit:100`, no user_id % 100 constraint
-- https://explain.depesz.com/s/gH1
danbooru2=# explain (analyze, verbose, buffers) select * from posts p join favorites f on f.post_id = p.id where f.user_id = 72775 order by f.id desc limit 100;                                                                               
Time: 7611.468 ms

-- with user_id % 100 constraint
-- https://explain.depesz.com/s/MAth
danbooru2=# explain (analyze, verbose, buffers) select * from posts p join favorites f on f.post_id = p.id where f.user_id = 72775 and f.user_id % 100 = 75 order by f.id desc limit 100;                                                      
Time: 8489.397 ms

-- with favorites_${user_id % 100}
-- https://explain.depesz.com/s/S3Jg
danbooru2=# explain (analyze, verbose, buffers) select * from posts p join favorites_75 f on f.post_id = p.id where f.user_id = 72775 order by f.id desc limit 100;                                                                            
Time: 6758.963 ms
```

r888888888:
i have no plans on adding an index to one of the biggest tables in the database. especially since the only thing it speeds up is ordfav.

those query results are interesting. for what it's worth running those queries on the live database, the first one averages 6-7s, second one is 4-5s, last one is 6-7s. i'm not sure why we're seeing such a big discrepancy; the amount of concurrency going on may affect it. i'll try rerunning them at a different time.

app/models/post.rb
```diff
@@ -1025,7 +1025,9 @@ def favorite_groups(active_id=nil)
     end
 
     def remove_from_favorites
-      Favorite.destroy_all(post_id: self.id)
+      favorites.find_each do |fav|
+        remove_favorite!(fav.user)
```
r888888888: this probably needs to be moved to delayedjob. right now this occurs in a transaction in a synchronous http request. it also calls `Favorite.remove` which does some locking itself. although i'm not sure if that would break anything if the post itself no longer exists. it should be fine i think?

r888888888:
I'm not sure what the best way to update this PR was so I forked it [here](https://github.com/r888888888/danbooru/pull/3299). I believe you can commit to that so if you have any questions or comments please update that one.

evazion:
That's fine, I've been busy lately and haven't had the chance to follow up on this.

PR #3268, opened 2017-08-13
update nginx config

tbennett6421:
On Ubuntu 16.04 LTS, the break in the config causes nginx to fail to restart


tbennett6421:
2017/08/12 20:30:27 [warn] 19202#19202: invalid value "http_5\
02" in /etc/nginx/sites-enabled/danbooru.conf:53

PR #3353, opened 2017-11-07
Migrated tag logic into config file and added fifth tag category meta

### LINKED ISSUE #3292: Add fifth category tag type for metatags
Brought up in the [Danbooru Forum](http://danbooru.donmai.us/forum_topics/14372).  Related to #2358, but a lot more straightforward.

Basically, there was a strong desire expressed for a fifth category of tags which would encompass all metatags (but not those used with a semicolon ":").  See [Tag Group:Metatags](http://danbooru.donmai.us/wiki_pages/87398) for a mostly complete list on what tags would qualify.

The advantage of grouping tags like this is that it moves those tags out of the sea of blue (all general tags) from those that don't care about them, while highlighting them to those that do want to see them.

The thought was to have this category be placed beneath the **General** category in the tag list, which would be right above the **Information** section, thus grouping all of the different meta-information about an image in the same general vicinity.

BrokenEagle:
Fixes #3292.  Most of the configuration logic for the tag categories was first moved out of all of the affected files and into the configuration file.  This will facilitate greater control over tag categories for anyone that uses a Danbooru instance, to include renaming, adding, or removing categories.

Some of the logic was left in still though, such as when one category affected another, or something only affected a single category.

One such case was for the new **meta** tag category.  An additional restriction of Builder+ only was applied to that as per Type-kun's suggestion, but since it only affected that category, it was left in the corresponding model and controller files.  However, it could always be moved out to the configuration file as well if necessary.

### DIFF
--- a/app/assets/javascripts/related_tag.js
+++ b/app/assets/javascripts/related_tag.js
@@ -12,10 +12,9 @@
 
   Danbooru.RelatedTag.initialize_buttons = function() {
     this.common_bind("#related-tags-button", "");
-    this.common_bind("#related-general-button", "general");
-    this.common_bind("#related-artists-button", "artist");
-    this.common_bind("#related-characters-button", "character");
-    this.common_bind("#related-copyrights-button", "copyright");
+    $.each(JSON.parse(Danbooru.meta("related-tag-button-list")), function(i,category) {
+      Danbooru.RelatedTag.common_bind("#related-" + category + "-button", category);
+    });
     $("#find-artist-button").click(Danbooru.RelatedTag.find_artist);
   }
 
--- a/app/assets/javascripts/related_tag.js
+++ b/app/assets/javascripts/related_tag.js
@@ -2,7 +2,7 @@
   Danbooru.RelatedTag = {};
 
   Danbooru.RelatedTag.initialize_all = function() {
-    if ($("#c-posts").length || $("#c-uploads").length) {
+    if ($("#c-posts #a-show").length || $("#c-uploads #a-new").length) {
       this.initialize_buttons();
       $("#related-tags-container").hide();
       $("#artist-tags-container").hide();
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -2776,7 +2776,8 @@ CREATE TABLE posts (
     pixiv_id integer,
     last_commented_at timestamp without time zone,
     has_active_children boolean DEFAULT false,
-    bit_flags bigint DEFAULT 0 NOT NULL
+    bit_flags bigint DEFAULT 0 NOT NULL,
+    tag_count_meta integer DEFAULT 0 NOT NULL
 );
 
 
@@ -7517,4 +7518,6 @@ INSERT INTO schema_migrations (version) VALUES ('20170613200356');
 INSERT INTO schema_migrations (version) VALUES ('20170709190409');
 
 INSERT INTO schema_migrations (version) VALUES ('20170914200122');
+
+INSERT INTO schema_migrations (version) VALUES ('20171106075030');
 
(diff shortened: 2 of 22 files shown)


BrokenEagle:
Screenshot, just for reference...

image

app/assets/javascripts/related_tag.js
```diff
@@ -12,10 +12,9 @@
 
   Danbooru.RelatedTag.initialize_buttons = function() {
     this.common_bind("#related-tags-button", "");
-    this.common_bind("#related-general-button", "general");
-    this.common_bind("#related-artists-button", "artist");
-    this.common_bind("#related-characters-button", "character");
-    this.common_bind("#related-copyrights-button", "copyright");
+    $.each(JSON.parse(Danbooru.meta("related-tag-button-list")), function(i,category) {
+      Danbooru.RelatedTag.common_bind(`#related-${category}-button`, category);
```
evazion: [Template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) (`\`#related-${category}-button\``) are a ES6 feature and aren't supported in IE11. This will cause a Javascript parse error in that browser.

BrokenEagle:
On suggestion from @evazion and concurrence from @Type-kun on Discord, the logic locking tag edits to or from **Meta** for Builder+ only was removed.  As evazion pointed out, most of the **Meta** category tags will be over a count of 50, locking them to Builder+ only.  If there are any issues with low-count **Meta** tags, such as *tagme** which is regularly pruned, then a Moderator+ can always lock that tag.

BrokenEagle:
I changed the Javascript removing the template as suggested by evazion above.

db/structure.sql
```diff
@@ -2776,7 +2776,8 @@ CREATE TABLE posts (
     pixiv_id integer,
     last_commented_at timestamp without time zone,
     has_active_children boolean DEFAULT false,
-    bit_flags bigint DEFAULT 0 NOT NULL
+    bit_flags bigint DEFAULT 0 NOT NULL,
+    tag_count_meta integer DEFAULT 0 NOT NULL
```
r888888888: This may need an index but let's see how the queries work out in practice first.

r888888888:
I've deployed the branch on testbooru.

Not sure if a dedicated button for related tags is needed for meta tags.

BrokenEagle:
Based upon a recommendation from Type-kun, I created a new **TagCategory** class in ***logical***, which has the benefit of decluttering the config file of non-config info in addition to making the variable references much shorter and easier to read.  I also changed the name in the config info from "relatedtag" to "relatedbutton" to better reflect what it signfies, i.e. the name of the button.

evazion also brought up the point that the html `<meta>`'s wouldn't be necessary if we used a *.js.erb.  Additionally, the scss files could also use *.scss.erb in order to move the tag config dependency out of them and into the config file.  Is that something we should do as a one-off here and there, or should we collate all such instances into a single **erb** wrapper for both **js** and **scss** files...?

db/structure.sql
BrokenEagle: I didn't see an index on the other tag category counts, so I'm guessing that there's no issue...?

r888888888:
I wasn't aware the assets could be templated. That could actually be very useful.

I think in general it's a good idea to move away from the meta tag paradigm as part of a larger move to minimize client-side processing.

BrokenEagle:
Found and fixed a bug with the related tag script.  It was running when it shouldn't on the **post/index**, **upload/index**, and **upload/show**.  The above caused it to error out when it couldn't find the `<meta>` it was looking for, i.e. the related tag button names.  It's now restricted to running only on **post/show** and **upload/new**.

BrokenEagle:
Found and fixed a bug with the related tag script.  It was running when it shouldn't on the **post/index**, **upload/index**, and **upload/show**.  The above caused it to error out when it couldn't find the `<meta>` it was looking for, i.e. the related tag button names.  It's now restricted to running only on **post/show** and **upload/new**.

r888888888:
This is now deployed. Can we set up or update a wiki page to explain what it is?

Issue #3352, opened 2017-11-04
Add additional validation to post creates/updates


BrokenEagle:
I've been prototyping some additional tag input validations with a userscript in [topic #1234](http://danbooru.donmai.us/forum_topics/14474) ([GitHub code link](https://github.com/BrokenEagle/JavaScripts/blob/stable/validatetaginput.user.js)).  I don't believe there are any additional considerations at this point, so it's probably safe to move forward with this.

Basically on tag inputs for uploads, post edits, and quick post edits, a validation is run on all added/removed tags to check for different error conditions.

- [x] 1. Tag add validation - Checks for tags that are empty, and that will be created/populated
- [x] 2. Tag remove validation -  Checks hierarchy relationships that will cause a tag removal to be readded
- [x] 3. Rating validation - Quicker and less destructive way to check that the rating exists
- [x] 4. First party source validation - Sources from Pixiv, Nicoseiga, Nijie et al should always have an artist tag.  An exception to this should be for official company sources as indicated by the presence of the **official_art** tag.
- [x] 5. Copyright tag validation - Posts should either have a copyright tag, the **original** tag, **copyright_request**

Preedit and postedit tag calculation (psuedocode):
```
preedittags = tags that already exist on post (for uploads this will be the null set)
postedittags = tags that exist at form submission
postedittags = transformdowncase(postedittags) #Normalize tags
postedittags = filtermetatags(postedittags)
```

### 1. Tag add validator

This is more of a warning condition, as it could potentially indicate a mispelling or mistagging.  Adding this check would reduce the vast number of mistags that get accidentally added.  One overide on the tag add validation would be if the tag is prepended with one of the tagcategory tags, e.g. `general:`, `gen:`, as this is seen as more of a deliberate action.

Added tags calculation (psuedocode):
```
addedtags = filtercategorytags(postedittags)  #The overide as mentioned before
addedtags = filternegativetags(addedtags)
addedtags = addedtags - preedittags  #Set difference
negativetags = getnegativetags(postedittags)  #Filters nonnegative and removes minus "-" prepend
addedtags = addedtags - negativetags  #Allows users to correct an error without having to hunt for it

```

For the validation, it checks all tag adds using the `/tags` controller with the `search[hide_empty]=true` option to find tags that don't exist or have a `post_count` of 0.  Additionally, it checks for any existing aliases using the `/tag_aliases` controller with `search[antecedent_name]=tagname` and also `search[active]=true`.

[...]

PR #3357, opened 2017-11-09
Post editing: add warning when creating new tags (#3352)

evazion:
This is a proof of concept demonstrating how the soft validations from #3352 can be done fully server-side. This adds two type of warnings:

* When new general tags are created, a `Created 2 new tags: <tag1>, <tag2>, ...` notice is displayed. This warns users about typos.
* When an artist tag is created and it doesn't have an artist entry, a `Artist <name> requires an artist entry. Create new artist entry.` notice is displayed. This reminds users to create artist entries.

The approach is to use standard Rails validations to perform these checks during the post saving process, but to store the warnings inside `post.warnings` instead of `post.errors`. This way we can add soft errors that don't prevent the object from saving. Then in the controller we simply set `flash[:notice] = @post.warnings.full_messages.join("; ")`, and Danbooru will display a warning notice after the page reloads.

The downside to this approach is that the user doesn't see the warning until after the edit completes and the page refreshes. The upside is that the implementation is simple and requires no extra Javascript.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -110,6 +110,10 @@ private
   def respond_with_post_after_update(post)
     respond_with(post) do |format|
       format.html do
+        if post.warnings.any?
+          flash[:notice] = post.warnings.full_messages.join(".\n \n")
+        end
+
         if post.errors.any?
           @error_message = post.errors.full_messages.join("; ")
           render :template => "static/error", :status => 500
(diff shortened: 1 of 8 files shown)


evazion:
Force pushed an update. Added the remaining warnings:

* Warn when a new upload doesn't have a copyright tag or the `copyright_request` tag.
* Warn when a new upload from a first party source (Pixiv, Twitter, etc) doesn't have an artist tag.
* Warn when a tag could not be removed due to implications, automatic tags, or `*_(cosplay)` tags. The error message here could be improved.

Issue #83, opened 2011-09-17
You are logged out after closing your browser.


Pyrolight:
Closing (in my case firefox) causes you to be logged out.

Issue #82, opened 2011-09-16
Notes do not hide when clicking image


Beefki:
Very useful for when the note itself is intruding on the image.

Issue #81, opened 2011-09-16
Only original size has working transparency 


Pyrolight:
Currently only the original size has the transparency working aka post 24.

Should and could it be applied to "Large" as well?

Issue #80, opened 2011-09-16
Clicking Pixiv source address at the information panel


ghostrigger:
unless this is a new feature, i noticed that clicking the pixiv source address at the left-side panel results to a direct image link outright. in the current production danbooru, there's a sort of a "landing" intermediate page: medium mode for single images and manga mode for stacked images.

PR #3362, opened 2017-11-13
Moved input existence validations to client (#3352)

BrokenEagle:
This adds in the third validation from #3352.  As explained on that issue, doing it this way is quicker and more responsive to the user as all the information is already available at the client.  

I took out the `error_messages_for` since the view ERB and Javascript code will cover all of those cases, but I left the original validation error pushes in the uploads model because the user will get those if they try to upload from the API.  The button is left as an `<input>` with a `value="Submit"` to take advantage of existing CSS that styles that button.  The `name="commit"` was left out as it's not sent anyways as a result of using the _trigger_ action with "submit", which is done for the **Return** key anyways so there shouldn't be an issue.

### DIFF
--- a/app/assets/javascripts/uploads.js
+++ b/app/assets/javascripts/uploads.js
@@ -12,6 +12,7 @@
       this.initialize_info_bookmarklet();
       this.initialize_similar();
       this.initialize_shortcuts();
+      this.initialize_submit();
       $("#related-tags-button").trigger("click");
       $("#find-artist-button").trigger("click");
 
@@ -24,6 +25,28 @@
     if ($("#iqdb-similar").length) {
       this.initialize_iqdb_source();
     }
+  }
+
+  Danbooru.Upload.initialize_submit = function() {
+    $("#form").submit(function(e) {
+      var error_messages = [];
+      if (($("#upload_file").val() === "") && ($("#upload_source").val() === "")) {
+        error_messages.push("Must choose file or specify source");
+      }
+      if (!$("#upload_rating_s")[0].checked && !$("#upload_rating_q")[0].checked && !$("#upload_rating_e")[0].checked &&
+          ($("#upload_tag_string").val().search(/\brating:[sqe]/) < 0)) {
+        error_messages.push("Must specify a rating");
+      }
+      if (error_messages.length === 0) {
+        $("#submit-button")[0].setAttribute("disabled","true");
+        $("#submit-button")[0].setAttribute("value","Submitting...");
+        $("#client-errors").hide();
+      } else {
+        $("#client-errors")[0].innerHTML = "<strong>Error</strong>: " + error_messages.join(", ");
+        $("#client-errors").show();
+        e.preventDefault();
+      }
+    });
   }
 
   Danbooru.Upload.initialize_shortcuts = function() {
--- a/app/views/uploads/new.html.erb
+++ b/app/views/uploads/new.html.erb
@@ -15,7 +15,7 @@
       <%= render "post", :post => @post %>
       <%= render "sources/info", :source => @source %>
 
-      <%= error_messages_for "upload" %>
+      <div id="client-errors" class="error-messages ui-state-error ui-corner-all" style="display:none"></div>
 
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/uploads.js
```diff
@@ -26,6 +27,27 @@
     }
   }
 
+  Danbooru.Upload.initialize_submit = function() {
+    $("#submit-button").click(function(e) {
+      var error_messages = [];
+      if (($("#upload_file").val() === "") && ($("#upload_source").val() === "")) {
+        error_messages.push("Must choose file or specify source");
+      }
+      if (!$("#upload_rating_s")[0].checked && !$("#upload_rating_q")[0].checked && !$("#upload_rating_e")[0].checked) {
```
evazion: It's possible to specify the rating in the tag string using the `rating:` metatag. This case would have to be checked too.

app/assets/javascripts/uploads.js
```diff
@@ -26,6 +27,27 @@
     }
   }
 
+  Danbooru.Upload.initialize_submit = function() {
+    $("#submit-button").click(function(e) {
```
evazion: This won't trigger if you submit the upload by pressing the enter key instead of clicking the submit button.

This could be fixed by binding to the form submit event instead (`$("#form").submit(function (e) ...)`) then saying `e.preventDefault()` in the error branch (prevent the submission from going through on error, rather than submitting it on success).

BrokenEagle:
Made the changes as suggested by evazion.  Reverted the submit button back to the way it was, but kept the ID to make it more uniquely addressable, plus left out the `disable_with => "Submitting..."` since it needs to be reusable.  Once a request has been validated, the Javascript disables the button instead.

r888888888:
I've rebased and deployed to Testbooru.

BrokenEagle:
I just checked, and I don't see it on Testbooru.

r888888888:
Sorry, I messed up the branch. The correct version should be deployed now.

BrokenEagle:
I ran a series of tests on Testbooru, using [uploads 75-82](http://testbooru.donmai.us/uploads).  I somewhat labeled them with tags, but will clarify them below.  Before submitting the request for each one, I tested each iteration with both the rating and file/source not present, only one of each present, then both present.  Only the ones where both were present went through as expected and created upload records.  For all but the last test, I used bogus sources or files to avoid generating too many new posts.

- Safe radio button, bad source used:  [upload #75](http://testbooru.donmai.us/uploads/75.xml)
- Questionable radio button, bad file used:  [upload #76](http://testbooru.donmai.us/uploads/76.xml)
- Explicit radio button, bad source used: [upload #77](http://testbooru.donmai.us/uploads/77.xml)
- Safe rating tag, bad source used: [upload #78](http://testbooru.donmai.us/uploads/78.xml)
- Questionable rating tag, bad source used: [upload #79](http://testbooru.donmai.us/uploads/79.xml)
- Explicit rating tag, bad file format: [upload #80](http://testbooru.donmai.us/uploads/80.xml)
- Safe rating tag, bad file format, legit source: [upload #81](http://testbooru.donmai.us/uploads/81.xml)
- Safe rating tag, legit source: [upload #82](http://testbooru.donmai.us/uploads/82.xml)

That last one went through successfully and created [post #53](http://testbooru.donmai.us/posts/53).

After thinking about it, I did one more test with the Javascript off.

- No Javascript, safe radio button, bad source used:  [upload #83](http://testbooru.donmai.us/uploads/83.xml)

The above went through fine, however when testing the various fail conditions, the program wouldn't process the upload as expected as the server-side validations are still in effect, but it also wouldn't give any error conditions when it failed since that was removed from the ERB view.

The question brought up by this is, should we support users that don't have Javascript enabled for the site?  Or should program behavior always be tailored to the expectation that Javascript is running?

r888888888:
I think it's safe to assume users, particularly ones who upload, will have Javascript enabled. I'll do one final rebase and deploy to Testbooru again.

BrokenEagle:
I tested it again, and the validations still work as expected.  There weren't as many upload records though since I started blocking the IP before every series of tests, which works well at testing the Javascript if it doesn't need to query the server and you don't want to actually create a record.

Issue #3365, opened 2017-11-14
Incorrect tag category counts


evazion:
https://danbooru.donmai.us/posts/2924654.json

Some (all?) posts after #3353 was merged have incorrect `tag_count_*` fields. See above for an example; it says the post has 167 gentags, 4 artists, 4 characters, 4 copyrights, and 12 meta tags.

evazion:
I *think* 5f39a8b fixes it. There were a few test failures in `test/unit/post_test.rb` that now pass, at least.

Leaving this open because the tag counts on affected posts still need to be fixed. I think just resaving the posts will be sufficient.

BrokenEagle:
Ah, snap... don't know how I missed that. I thought I had rechecked each function individually?

What's the best way to go about resaving all of the posts?

evazion:
Closing, I think the tag counts are correct now.

PR #3369, opened 2017-11-14
Added fix script to update the correct tagcount for all posts

BrokenEagle:
This is for #3365.  After speaking it over with @evazion, we realized that fixing the tagcounts client-side would take forever, and that it would be much better to do it server-side.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -624,10 +624,10 @@ class Post < ApplicationRecord
       set_tag_count(category,self.send("tag_count_#{category}") + 1)
     end
 
-    def set_tag_counts
+    def set_tag_counts(disable_cache = true)
       self.tag_count = 0
       TagCategory.categories.each {|x| set_tag_count(x,0)}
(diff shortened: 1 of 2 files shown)


BrokenEagle:
Doing some initial runs on this, it was taking about 50 ms per post to update, which would end up being ~48 hours for 3M posts.  After working it out with evazion, we added a couple of optimizations to the `Post.fix_post_counts` function, thereby dropping the time down to 20 ms.  Further refining was done by restricting it to posts that were updated after 11/12/2017, and posts with tags that were changed to the **Meta** tag category.

r888888888:
I think we need to enable the category cache because as is it's taking forever. Something like check_translation with around 20k posts takes like an hour. The n+1 factor here is crazy: a single SQL query for each tag for each post.

I think as long as we lower the priority of Tag#update_category_post_counts should be sufficient. The category cache will always be updated locally, and the cache update job will always run before update_category_post_counts on the other server.

Issue #3392, opened 2017-11-21
Add safe mode account setting


evazion:
Currently you have to go to Safebooru to browse in safe mode. There should be an account-level setting to enable safe mode, so that you can use it on the main domain too.

PR #3399, opened 2017-11-23
Changed safe mode error message for Gold+ users

BrokenEagle:
It seems about once every few months a user asks on the forum why they can't view certain posts after they have upgraded to Gold+.  [Topic #14724](http://danbooru.donmai.us/forum_topics/14724) is just the latest, but if you look back, you can see it happens pretty often, probably moreso than gets reported.

This issue occurs mostly because the error message is confusing as pointed out by @kittey on the above topic.  It tells Gold+ users to upgrade to Gold when it hides content in Safe Mode.  Going by his recommendation to fix this issue, I changed the error message to redirect Gold+ users to Danbooru or to adjust their user settings due to #3392.  The wiki that was linked doesn't exist yet (**Help:User Settings**), but such a wiki probably should exist and I plan on creating one.  A hard link was added to the Danbooru domain, but if it's an issue it can always be moved to the config file (along with several other embedded hard links).

### DIFF
--- a/app/logical/post_sets/post.rb
+++ b/app/logical/post_sets/post.rb
@@ -87,11 +87,15 @@ module PostSets
     end
 
     def banned_posts
-      posts.select(&:is_banned?)
+      posts.select { |p| p.banblocked? }
     end
 
     def censored_posts
-      hidden_posts - banned_posts
+      posts.select { |p| p.levelblocked? && !p.banblocked? }
+    end
+
+    def safe_posts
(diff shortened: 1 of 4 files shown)


BrokenEagle:
Per recommendation from @kittey due to [forum #140223](http://danbooru.donmai.us/forum_posts/140223), I added a safe-mode error message to the post show.  I also made the ordering explicit, from banned, to Gold+, to safe mode.  Additionally, I made sure that the groups are mutually exclusive.

PR #3432, opened 2017-12-18
Move lib/danbooru, lib/danbooru_image_resizer to app/logical.

evazion:
Currently the paginator code and the image resizer code reside under `lib/`. This moves them to `logical/`.

This fixes an issue with code reloading during development not working. The paginator code was only included in `ApplicationRecord` once during bootup, which meant that it wouldn't get reloaded by Rails after you edited a file. This led to pages throwing `undefined method 'paginate'` errors after you edited a file.


evazion:
Rebased and force pushed to fix merge conflicts.

Issue #2424, opened 2015-06-29
Private favorite groups


ToksT:
Add a checkbox when creating/editing a favgroup for whether it should be publicly visible or only to you.

evazion:
This was fixed by #3436.

PR #3436, opened 2017-12-19
Give option to make favorite groups public

BrokenEagle:
Builds onto #3433 by giving the option to make favorite groups public.  It also changes how the index search parameters are handled, showing only the public favorite groups instead of throwing a permission error.


evazion:
#2424 also suggested this. IMO, though, favgroup visibility should be determined by the public/private flag alone. I suggest that the "privacy mode" account setting should only affect favorites, not favgroups.

As it stands, to make a single favgroup private, you'd have to opt-in to privacy mode, then opt-out of all your other favgroups being private. Also, it's not possible to make a single favgroup private without making your favorites private as well.

PR #3456, opened 2017-12-24
Fix 3455: Convert tag category Javascript/CSS assets to ERB

### LINKED ISSUE #3455: Shorthand versions of tag types not recognized by search autocomplete
From the [Danbooru forum](http://danbooru.donmai.us/forum_posts/140890):

> kuuderes_shadow said:
> 
> Can this be changed so that sticking copy: or char: etc. at the start is ignored?
> 
> I noticed when typing copy:tensei the top results are:
> [[megami_tensei]]
> [[tenseiani]]
> [[mushoku tensei]]
> [[copy x]]
> [[edo tensei]]
> [[comic mugen tensei]]
> [[tense]]
> 
> when there are 13 tags beginning with "tensei" (probably 14 by the time you read this), of which 11 (12) are copyright ones (none of which appear in the dropdown at all, although both the non-copyright ones do) which the person might actually be attempting to enter. [[Tensei shitara slime datta ken]] is the top result by the time you reach "tensei_" but when doing copy: at the start it doesn't even appear on the dropdown until "copy:tensei_shita" and doesn't get above [[shin megami tensei]] until you reach "copy:tensei_shitara_sl"

BrokenEagle:
Fixes #3455.  ~~Also, I didn't consider worth it to add the very short category names of **ch** and **co**.~~

### DIFF
--- a/app/assets/javascripts/autocomplete.js.erb
+++ b/app/assets/javascripts/autocomplete.js.erb
@@ -9,7 +9,6 @@
   Danbooru.Autocomplete.initialize_all = function() {
     if (Danbooru.meta("enable-auto-complete") === "true") {
       Danbooru.Autocomplete.enable_local_storage = this.test_local_storage();
-      this.update_static_metatags();
       this.initialize_tag_autocomplete();
       this.initialize_mention_autocomplete();
       this.prune_local_storage();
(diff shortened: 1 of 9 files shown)


BrokenEagle:
I went back and decided to just bite the bullet and move applicable assets involving tag categories over to ERBs as discussed in #3353.  This has the added benefit of adding in **ch** and **co** for the tag categories.

For the **posts.scss.erb** in particular I encountered an error which required the reshuffling of affected assets.

``File to import not found or unreadable``

^That was the error encountered after renaming the **posts.scss** file to **posts.scss.erb**.  After a bit of searching, it seems like wildcards don't work for *.scss.erb files.

github.com/rails/sass-rails/issues/139

The current solution is still to address such ERB files directly in the **application.scss** file.  Therefore the **posts.scss.erb** file was moved out of the **stylesheets/specific** directory and into the **stylesheets/erb** directory.

Issue #2833, opened 2017-01-11
Standalone IQDB page


evazion:
Discussion: http://danbooru.donmai.us/forum_topics/13630

> Since Danbooru has it's own IQDB engine available to registered users, it'd be nice if there was a standalone page like http://iqdb.org where this functionality could be accessed, outside of the Uploads page or the Find Similar function.

BrokenEagle:
I added a pull for the standalone page that handles the case of post ID or URL as an input.  It would still be nice however to add support for local files.

I see a few ways to do this:

1. Convert the file to a thumbnail and host it somewhere
a. Temporarily on Danbooru
b. Temporarily on another 3rd party site
2. Modify IQDBS so that it can take file input from Danbooru
a. Add support for POST endpoint and multipart file encoding

1.a might be the simplest to implement, but it depends on how much space is available for thumbnails.  I can't imagine that these thumbnails would need to be stored for more than an hour though.

evazion:
This exists at https://danbooru.donmai.us/iqdb_queries and it now supports files.

PR #3479, opened 2017-12-30
Initial add of IQDB standalone page

BrokenEagle:
A partial fix of #2833.  I say partial since it would be nice to eventually add support for local files, and so I'd like to keep that issue open at least for now.

I converted the endpoint to a singular resource since it doesn't really need the actions that resources do, and I added an additional named action since it didn't make sense to have the standalone page on the new or edit action.


r888888888:
So the thing that held me back from implementing this is I was worried about people abusing it in scripts and whatnot and DDOSing the server, which is not beefy at all. But I guess we can find out if it's actually a problem in practice. It may require some sort of rate limiting in the future.

PR #3521, opened 2018-01-27
Add multiple show views and more available options on index views

BrokenEagle:
@nonamethanks brought up the idea of having direct links for many of the controllers, since there currently isn't a good way of getting the ID of an instance without going to the API.  This would be useful in pointing out an individual instance, either in a comment, on the forum or in a Dmail.  It's true that ``search[id]`` could be used for some of those, but the show link seems a bit more clean in its usage.

There are also multiple items that aren't shown on the index views, but that could be shown with the additional space of a show view.  Additionally, more available options were added to the show, index and secondary links.

In the course of the above, I discovered that BURs don't set the forum post ID when they run the script, so I fixed that.  If needed, I can submit that as a separate pull.

Additionally, I discovered that although the DText is rendered as inline in the notices, it wasn't being rendered as inline in the DText preview or in the index view.  This could go as part of another pull if needed, although the following commit of this pull is dependent on it.

I made flags and appeals their own commit since there was a lot more work involved.  I first started with the general idea of just showing more of the flag/appeal information like I had done with the other controllers, but then nonamethanks (an approver) brought up the idea of wanting a more visible approve button for flags/appeals.  Instead of changing the post show page, I thought it might be more useful to try making the appeal and flag pages more specialized towards a (re)approval decision.

1. Flag/appeal information
2. Most of the pertinent post information without any unnecessary functions like voting or favoriting
3. Meta tags which would show information outside the post
4. Artist tags which would show their banned status and also provide an easy link to compare against their other works
5. Large approve button for approvers
6. The image at large size (not original)
7. All child/parent relations to include deleted posts
8. Flag and appeal history

(1) went at the top since it's rendered inline, which could potentially make it overflow any HTML container. (2)-(5) went on the left side to present all information. (6) went in the center to make the image the focus. (7) went on the side to avoid disrupting the focus on (6).  (8) went on the bottom because of its table format.

image

Post appeals look almost identical, with the difference being the appeal information, and the color of the reason being the same as used for appeals on the post notice.  Both of these views will most likely go through a few iterations, but I feel that the above is a good starting point.


BrokenEagle:
Based on a recommendation from @nonamethanks, I went back and added percentages to the column widths for the mod actions table for the first commit.  This is because long messages, such as a long ban message, can crowd out the User and Date columns.

http://danbooru.donmai.us/mod_actions?page=b223459

PR #3524, opened 2018-01-29
Fix #3522: Enable HSTS

### LINKED ISSUE #3522: Enable HSTS
Discussed in [topic #14902](https://danbooru.donmai.us/forum_topics/14902) and previously in [topic #12916](https://danbooru.donmai.us/forum_topics/12916). In summary, I propose this:

- [ ] Enable HSTS for danbooru.donmai.us and safebooru.donmai.us (but not for sonohara.donmai.us or hijiribe.donmai.us, unless their SSL certs are fixed).
- [x] Add an http://insecure.donmai.us domain for those who want to continue using HTTP.

A significant consequence of this is that, since nowadays browsers default to HTTP/2 when using SSL if the site supports it (which Danbooru does), this will cause most traffic to switch to HTTP/2. This should be a good thing, but it may be worth monitoring.

evazion:
Fixes #3522. Enabling HSTS is just a matter of setting `config.force_ssl = true` in `config/application.rb`. Everything else is just cleanup to make that work.

Note that the Nginx config will have to be updated to set `X-Forwarded-Proto` correctly, as per eeecf49.


evazion:
```ruby
def force_ssl?
  Rails.env.production? && server_host.in?(["danbooru.donmai.us", "safebooru.donmai.us"])
end
```

Just realized this has a problem. `server_host` will always be `hijiribe.donmai.us` or `sonohara.donmai.us`, so this won't trigger. We need to do something like this instead:

```ruby
config.force_ssl = true
config.ssl_options = {
  redirect: {
    exclude: ->(request) { request.host.in?(["insecure.donmai.us", "sonohara.donmai.us", "hijiribe.donmai.us"]) }
  }
}
```

...but the `exclude` option only works on Rails 5. This may have to wait until then.

r888888888:
What I can do is purchase SSL certs for sonohara and hijiribe. It's annoying because these days I would just use letsencrypt but using that together with conventional certs is probably a headache.

r888888888:
https://sonohara.donmai.us and https://hijiribe.donmai.us and http://insecure.donmai.us are now functioning.

reiyasona:
@r888888888 
In the near future, [Let's Encrypt will support wildcard certificates](https://letsencrypt.org/upcoming-features/). Let them sign a certificate for `*.donmai.us`, and you're done.

evazion:
@r888888888 fixed the merge conflicts, it should be ready to go now.

PR #3525, opened 2018-01-29
Fix index default order

BrokenEagle:
I noticed that with the latest deployment that the bans index was [showing weird orderings](http://danbooru.donmai.us/bans), not being ordered by ID or expiration.  After looking into it, I found out that it was most likely broken by #3495 when it was changed to early return if the parameters are blank.  Also, the order logic was divided between either being specified in the controller or in the model.

After speaking with @evazion, we thought it would be best to standardize all of the models.  Since orders are frequently specified in the model as part of the **search** function, it would be easiest to move there.  When there was an **order** block, then that was moved to the end of the **search** function, and if not then the final action was to apply the default order to the query.  Also, when one of the order terms matched the default order then it was eliminated, such as ``order[time]`` for wiki pages.

There is a default for all models of ID DESC, however each model can specify its own default.  Additionally, we thought that having a custom ordering for records would be useful on occasion, so that was added to all models as well.

### DIFF
--- a/app/models/artist.rb
+++ b/app/models/artist.rb
@@ -536,7 +536,6 @@ class Artist < ApplicationRecord
 
     def search(params)
       q = super
-      params = {} if params.blank?
 
       case params[:name]
       when /^http/
@@ -581,18 +580,6 @@ class Artist < ApplicationRecord
         q = q.url_matches(params[:url_matches])
       end
 
-      params[:order] ||= params.delete(:sort)
-      case params[:order]
-      when "name"
-        q = q.order("artists.name")
-      when "updated_at"
-        q = q.order("artists.updated_at desc")
-      when "post_count"
-        q = q.includes(:tag).order("tags.post_count desc nulls last").references(:tags)
-      else
-        q = q.order("artists.id desc")
-      end
-
       if params[:is_active] == "true"
         q = q.active
       elsif params[:is_active] == "false"
@@ -622,6 +609,18 @@ class Artist < ApplicationRecord
         q = q.joins(:tag).where("tags.post_count > 0")
       elsif params[:has_tag] == "false"
         q = q.includes(:tag).where("tags.name IS NULL OR tags.post_count <= 0").references(:tags)
+      end
+
+      params[:order] ||= params.delete(:sort)
+      case params[:order]
+      when "name"
+        q = q.order("artists.name")
+      when "updated_at"
+        q = q.order("artists.updated_at desc")
+      when "post_count"
+        q = q.includes(:tag).order("tags.post_count desc nulls last").order("artists.name").references(:tags)
+      else
+        q = q.apply_default_order(params)
       end
 
       q
--- a/app/models/artist_version.rb
+++ b/app/models/artist_version.rb
@@ -15,7 +15,6 @@ class ArtistVersion < ApplicationRecord
 
     def search(params)
       q = super
-      return q if params.blank?
 
       if params[:name].present?
         q = q.where("name like ? escape E'\\\\'", params[:name].to_escaped_for_sql_like)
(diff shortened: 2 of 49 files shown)


evazion:
The ordering logic is a little confusing, it looks like the `search` methods return before the default order is applied, when actually it was called already in `ApplicationRecord#search` and we're returning early to avoid it being applied twice.

I think this could be avoided by handling the custom order inside `default_order`, something like this:

```ruby
# q = q.default_order(params, default: { updated_at: desc })
def default_order(params, default: { id: :desc })
  case params[:order]
  when "custom"
    parse_ids = Tag.parse_helper(params[:id])
    if parse_ids[0] == :in
      find_ordered(parse_ids[1])
    end
  else
    order(default)
  end
end
```

That way there wouldn't be a need to do `return q if params.blank?` everywhere. We could go through the full method normally and the custom order would be applied inside `default_order`.

app/models/artist.rb
```diff
@@ -624,6 +612,18 @@ def search(params)
...
       end
 
+      params[:order] ||= params.delete(:sort)
+      case params[:order]
+      when "name"
+        q = q.order("artists.name")
+      when "updated_at"
+        q = q.order("artists.updated_at desc")
+      when "post_count"
+        q = q.includes(:tag).order("tags.post_count desc nulls last").references(:tags)
```
evazion: A pre-existing bug: this should be `order("tags.post_count desc nulls last, artists.name")`. Ordering by post count alone means the order is undefined when artists have the same post count.

app/models/ban.rb
```diff
@@ -58,7 +58,7 @@ def self.search(params)
     when "expires_at_desc"
       q = q.order("bans.expires_at desc")
     else
-      q = q.order("bans.id desc")
+      q = q.default_order
```
evazion: I removed the `return q if params.blank?` from this method in 5ea61c69212d56bbf301ba87cd689d312ea09802, it would have to be added back.

app/models/artist_version.rb
```diff
@@ -52,6 +45,13 @@ def search(params)
         q = q.where("is_banned = false")
       end
 
+      params[:order] ||= params.delete(:sort)
+      if params[:order] == "name"
+        q = q.order("name")
```
evazion: This needs to be `q.order("name, id desc")`. Previously the `.order("id desc")` clause in the controller effectively did this, so removing that means it needs to be added here.

app/models/pool.rb
```diff
@@ -99,6 +91,18 @@ def search(params)
...
       end
 
+      params[:order] ||= params.delete(:sort)
+      case params[:order]
+      when "name"
+        q = q.order("pools.name")
+      when "created_at"
+        q = q.order("pools.created_at desc")
+      when "post_count"
+        q = q.order("pools.post_count desc")
```
evazion: Removing `order("updated_at desc")` from the controller means it needs to be added here: `order("pools.name, pools.updated_at desc")` and `order("pools.post_count desc, pools.updated_at desc")`.

BrokenEagle:
I made several of the changes that evazion suggested, including removing all of the ``return q if params.blank?`` as they are no longer required, although I left the ``default_order`` functions as is and just created an ``apply_default_order`` function.

Also I forgot to bring it up in my previous comment, but I moved all of the default orderings that used ``created_at`` to ``ID`` instead, as they should be equivalent.  However, are there any performance reasons why one may have been used over the other, and if so, should some of the models be changed to use ``created_at: :desc``?

evazion:
Ordering by `created_at` will be slower if the table doesn't have a `created_at` index (which most should have, but still). Also, the ordering will be undefined in the unlikely event that two records have the exact same `created_at` timestamp. So ordering by ID should be better.

PR #3529, opened 2018-02-01
implement post keepers

r888888888:
Hide uploader information (still visible in post history) and add a new field for keeper (most frequent tagger)


r888888888:
In practice this changes attribution 5% of the time. I tried various weighting schemes but it always devolved into ridiculous situations where adding one tag could be weighed higher than adding 20.

PR #3531, opened 2018-02-02
Rails 5.1

r888888888:
Upgrade to Rails 5.1. Main work is phasing out protected_attributes and switching to strong params.


evazion:
I was doing some work on converting everything to strong params myself this week. I'll make a pull request for what I've got done.

evazion:
Did some work and got it running, but there are a lot of test failures. Here's my branch: https://github.com/evazion/danbooru/compare/fix-strong-params...wip-rails-5.1.

It might be easier to upgrade to 5.0 first. Turns out there are a lot of little things that were first deprecated in 5.0 then removed in 5.1.

r888888888:
Ugh this is turning into a shitshow. Functional tests were deprecated so everything has to be rewritten as an integration test, but that means we can't stub in session or cookies anymore, and the syntax for everything is different. I'm slogging through it but it'll be awhile.

r888888888:
It's finally done!? All tests passing.

PR #3540, opened 2018-02-07
Revert "implement post keepers"

r888888888:
Reverts r888888888/danbooru#3529


r888888888:
I've realized the async nature of post archiving causes problems.

PR #3551, opened 2018-02-17
Fix #3528: Add method to prevent image-hosting CloudFlare sites from altering the image

### LINKED ISSUE #3528: Add method to prevent image-hosting CloudFlare sites from altering the image
It was recently noted that Artstation will send different images depending on whether an image is a cache hit or miss.  Artstation uses CloudFlare, and @evazion noted that that service uses a function called polish which strips an images metadata and also compresses the image.

https://blog.cloudflare.com/introducing-polish-automatic-image-optimizati/

One solution to this is to add a randomized dummy parameter to the URL to prevent a cache hit.  Another alternative would be to set the outgoing headers using cache control.

``Cache-Control: no-cache, no-store, no-transform``

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

Artstation is pretty bad about the above, and there are a plethora of MD5 mismatches between Danbooru and Artstation because of it.  It's possible that it also affects other sites as well.  Most users would have no means around the above, which is why it should be the server that handles it.

evazion:
Fixes #3528. Adds a random URL param when downloading files to bypass any transparent caches that may  potentially alter the file.

### DIFF
--- a/app/logical/downloads/file.rb
+++ b/app/logical/downloads/file.rb
@@ -26,20 +26,23 @@ module Downloads
     end
 
     def size
-      @source, headers, @data = before_download(@source, @data)
+      url, headers, _ = before_download(@source, @data)
       options = { timeout: 3, headers: headers }.deep_merge(Danbooru.config.httparty_options)
-      res = HTTParty.head(@source, options)
+      res = HTTParty.head(url, options)
       res.content_length
     end
 
     def download!
+      url, headers, @data = before_download(@source, @data)
+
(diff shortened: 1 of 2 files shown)


evazion:
@r888888888 this needs to be reverted. It turns out it breaks Twitter:

https://pbs.twimg.com/media/DWliR4nVwAE-Cus.jpg:orig (fine)
https://pbs.twimg.com/media/DWliR4nVwAE-Cus.jpg:orig (404)

PR #3577, opened 2018-03-21
Introduce storage manager concept

### LINKED ISSUE #3569: Post replacement breaks posts whose images are hosted on raikou2.donmai.us
When post replacement is done on an old post whose sample/fullsize are hosted on https://raikou2.donmai.us/, the new image is stored as new posts store images, but the post still tries to use the old https://raikou2.donmai.us/ URLs with the new MD5.

See [post #872149](https://danbooru.donmai.us/posts/872149) for one such casualty.

evazion:
This introduces a concept of storage managers for dealing with the storage of uploaded files. This incidentally fixes #3569, but the main goal is to clean up how archived files are dealt with.

This PR makes it possible to configure how uploads are stored: locally, via SFTP, or via S3. This also makes it possible to fully store all files externally, so that the webservers can be independent of the asset servers.

The basic idea is to introduce a `StorageManager` class with `store` and `delete` methods. Subclasses implement three different storage methods: local filesystem, SFTP, or S3. Each method supports flat or hierarchical directory structures. The SFTP method supports multiple hosts, which allows you to store files to both sonohara and hijiribe at the same time.

There is a fourth "hybrid" method, which lets you choose the method to use based on the post's ID and type (preview, sample, or original file). In this way it's possible to support complex configurations like Danbooru's, where files are stored across different servers with different URL structures. [Here](https://github.com/r888888888/danbooru/commit/b0c7d9c18527be41ab79b8db9c30c5522d1f684e#diff-ca389c899d0120072c1b5b187c4c5cbdR244) is how you would replicate Danbooru's current setup.

The basic idea itself is simple enough, but integrating it with the upload process turned out to be somewhat involved. This was mainly because of assumptions throughout the code that files are stored locally on the webserver. In short:

* Autotagging of animated GIFs/PNGs was moved from during post editing to during upload. This assumed that files were available on the local webserver, which we don't want to assume. Rechecking the file on every edit isn't very efficient anyway.

* Ugoira .webm conversion was made synchronous instead of asynchronous. Doing it asynchronously created multiple problems: it assumed the .zip file was available locally, it could break backups and iqdb submissions (since the conversion may not be finished yet when these things run), and it required hacky file existence checks in various places to detect whether the conversion was finished and the images were ready to use.

* Backups were made synchronous instead of asynchronous. Again, this assumed that uploaded files were present on the local webserver. It also meant that if the job worker went down, uploads would be accepted without being backed up.

* The signature for `Post.delete_files` had to be changed. Important: this means existing file deletion jobs need to be cancelled ASAP. These jobs wouldn't work right anyway, since in many cases they're trying to delete files that have since been archived.

* `/posts.json` now returns absolute URLs in all cases. Right now it returns absolute URLs for some posts and relative URLs for others, which clients have to detect and fix in order to download files. This change may break scrapers. However, it's not hard to fix and makes the API easier to deal with.

* The upload process was reworked to pass around temp files, instead of writing preview/samples directly to the local filesystem. This simplified some things.

* URL building was moved to `StorageManager`, since the URL structure depends on how files are stored.

* The backup system was changed to use the storage management system. This means you can choose between backing up files to a local directory, to SFTP, or to S3.

Making backups and ugoira conversions synchronous may slow down uploads. Ultimately I think it'd be better to make the whole upload process asynchronous, but that's outside the scope of this.


r888888888:
Is this good to merge now?

evazion:
Yeah, I'm finished with it, should be good to go.

Issue #102, opened 2011-10-02
Problems with some characters in tags


jjjoe:
Searching for a tag with an ampersand or semicolon in it doesn't work:
http://testbooru.donmai.us/posts?tags=tiger_&_bunny actually searches for just tiger_
http://testbooru.donmai.us/posts?tags=panty_&_stocking_with_garterbelt searches for just panty_
http://testbooru.donmai.us/posts?tags=steins;gate searches for steins.

If you enter a tag with a semicolon in it, it's converted to a space:
http://testbooru.donmai.us/posts/25

Tags with colons, slashes, or tildes work correctly:
http://testbooru.donmai.us/posts?tags=magic:_the_gathering
http://testbooru.donmai.us/posts?tags=.hack//roots
http://testbooru.donmai.us/posts?tags=nyoro~n

Issue #101, opened 2011-10-01
comment limitations


ghostrigger:
just going to report that the "1-week old members cannot post yet" and "one comment per hour"  seem to be not implemented. this might be prone to abuse.

PR #3584, opened 2018-03-28
Fix #3582: Switch from ImageMagick to libvips

### LINKED ISSUE #3476: Samples and thumbnails discard color profiles
As brought up in [topic #14833](https://danbooru.donmai.us/forum_topics/14833), samples and thumbnails of images with non-sRGB color profiles have wrong colors. Example: [post #2970342](https://danbooru.donmai.us/posts/2970342)

The cause is that Danbooru’s resizer discards all color profiles. https://github.com/r888888888/danbooru/blob/1ca2497015c03e99ff9809bcd77e61942eec9127/app/logical/danbooru_image_resizer.rb#L18

The solution would be to unconditionally convert to sRGB before stripping. This works even if the input file has no profile because sRGB is assumed in that case anyway. With command line ImageMagick, this does the trick:

`convert input.png -resize 850x -profile /usr/share/color/icc/ghostscript/srgb.icc -strip sample.jpg`

Unfortunately, I don’t know how to do that with RMagic in Ruby, so hopefully one of our Ruby wizards can figure it out.

On my PC, the only source for the sRGB color profile is GhostScript. If that’s not installed on Danbooru’s servers, the `icc-profiles-free` package installs the sRGB profile as `/usr/share/color/icc/sRGB.icc`.

This is unrelated to issue #2870, which may not even apply to recently created samples and thumbnails anymore.

### LINKED ISSUE #3570: ImageMagickError: width or height exceeds limit
From https://danbooru.donmai.us/forum_topics/15066:

> Fisheye placebo (pool #10558) posted for the first time in 3 years. Only issue is, the hight limits have been changed since it last posted so I can't upload a 800x21916 anymore (it's not quite that tall this time). Is there a way around that, or is two images at 10753px and 9912px the only way?

When uploading, ImageMagick raises an error if the image dimensions exceed some default limit. Not sure what a better limit would be, but it would at least better to set known limits rather than relying on implicit defaults. Uploading should also check the dimensions before resizing and fail gracefully if the image is too big to resize.

The limit can be adjusted by editing ImageMagick's `policy.xml` config file (see also https://www.imagemagick.org/discourse-server/viewtopic.php?t=31438).

### LINKED ISSUE #3582: Switch from ImageMagick to libvips
[libvips](https://github.com/jcupitt/libvips) is an alternative library for resizing images. It has several advantages over ImageMagick:

* It boasts better speed and lower memory usage: https://github.com/jcupitt/libvips/wiki/Speed-and-memory-use.

* It has a well-documented and well-maintained ruby binding ([ruby-vips](https://github.com/jcupitt/ruby-vips)). [rmagick](https://github.com/rmagick/rmagick) by contrast has been dead since 2016. rmagick is still stuck on ImageMagick version 6, while the latest version is 7. This makes it difficult to even install on some distros (namely Arch).

* Thumbnailing is easier (basically takes two lines to do everything we do) and it handles color profiles correctly (#3476).

evazion:
Fixes #3582, #3570, and #3476:

* Replaces ImageMagick with libvips. libvips must be installed on your system; on Debian the package is `libvips-dev`.

* Fixes #3570, in that libvips won't error out on huge images the way ImageMagick does. There still needs to be limits on uploading excessively large images though. Sufficiently large images can take multiple gigabytes of memory to resize.

* Fixes #3476. libvips correctly converts non-sRGB images to sRGB when resizing. The sRGB profile used is a public domain one taken from [ArygllCMS](https://www.argyllcms.com/).

* There's an option to resize images in a linear colorspace. This would be more accurate, but it's significantly slower, so it's not used. See [testbooru #101](http://testbooru.donmai.us/posts/101) and [testbooru #102](http://testbooru.donmai.us/posts/102) for a couple examples of the difference this makes.


r888888888:
So far the only downside of this has been the crazy amount of dependencies it sucks in. Even on homebrew it pulls in like 20 packages.

evazion:
Oh, sorry, didn't realize that. I guess Debian builds with all the optional dependencies enabled.

evazion:
Just noticed a problem. This requires libvips-8.5 (released a year ago), but Debian still packages libvips-8.4. There's an Ubuntu PPA with it [here](https://launchpad.net/~dhor/+archive/ubuntu/myway), not sure if it would work though.

r888888888:
So [there is a way](https://wiki.debian.org/CreatePackageFromPPA) but installing a package manually like this is kinda dicey, especially if there are missing dependencies. It also complicates the install script. Unfortunately it doesn't look like 8.5 is coming in Buster.

Would it be simpler to fork the gem and try to get it working on 8.4?

PR #3655, opened 2018-04-19
Implement BUR pruning

r888888888:
* update migrations to include version
* implement voting on forum posts
* prune old tag changes


BrokenEagle:
I don't know if it'd be possible, but it'd also be nice to know how people vote like how it's currently done on GitHub.  The current scheme has that visibility, so there shouldn't be any reason not to keep it.  The only difference this new scheme will change is that it will be easier to vote and not perturb those that may want to vote but also don't want to create a forum post just for that.

PR #3666, opened 2018-04-26
Fix #3664: Highlight non-gentags in post tooltip

### LINKED ISSUE #3664: Highlight non-gentags in post tooltip
Something that has been suggested in the forums (https://danbooru.donmai.us/forum_topics/15163) and that would be pretty neat to have.

The idea is to make these tags show up in separate lines above the main tag list in the tooltip so that it's possible to tell with a glance, basically like it's already done in each post's page.

I know some of us spend a lot of time when tagging chracters from copyrights we don't know by opening each post and looking at the character tag list, so easier tagging would also be a reason to implement this.

evazion:
Fixes #3664. This goes a step further and replaces the default browser tooltips with HTML tooltips. This allows tags to be colorized and clickable too.

This works as follows:

* After hovering over the thumbnail for enough time, we make a call to `/posts/{id}?variant=tooltip` to get the tooltip html and insert it into the tooltip. This template uses Rails' [variants system](http://guides.rubyonrails.org/4_1_release_notes.html#action-pack-variants).
* [qtip2](http://qtip2.com) is used to generate the tooltips. This gives us a look and feel that's close to native browser tooltips.
* To generate the tag list, we use `split_tag_list_html`, which is the same code that generates the sidebar tag list. 442964b and f299858 were refactorings to make this possible. A side effect of this is that tag lists on the /comments page and in the modqueue are now sorted by category too (previously they were just alphabetical).

Screenshot:

awesomescreenshot-2018-04-26t04-04-51-291z

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -6,6 +6,7 @@
 //= require jquery.storageapi.js
 //= require jquery.dropdown.min.js
 //= require jquery.hammer.js
+//= require jquery.qtip.js
 //= require ugoira_player.js
 //= require stupidtable.js
 //= require rails.js
--- a/app/assets/javascripts/post_tooltips.js.erb
+++ b/app/assets/javascripts/post_tooltips.js.erb
@@ -1,0 +1,51 @@
+Danbooru.PostTooltip = {};
+
+Danbooru.PostTooltip.render_tooltip = function (event, qtip) {
+  var post_id = $(event.target).parents("[data-id]").data("id");
+
+  return $.get("/posts/" + post_id, { variant: "tooltip" }).done(function () {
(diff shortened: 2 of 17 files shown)


r888888888:
this might be a contentious change. i suspect 300ms is too little.

evazion:
Just noticed this broke after the jquery migrate script was removed:

```
jQuery.Deferred exception: r.getClientRects is not a function TypeError: r.getClientRects is not a function
```

[This suggests](https://stackoverflow.com/questions/37914869/jquery-ui-error-f-getclientrects-is-not-a-function) that we need to upgrade to jquery-ui 1.12.

evazion:
Pushed a couple tweaks: increased the delay to 450ms, adjusted the height, and condensed the tag list. This should feel a bit better.

PR #3693, opened 2018-05-01
get travis ci working

r888888888:


r888888888:
This actually kinda works now, although some tests are failing.

PR #3716, opened 2018-05-16
Artist urls can be prepended with a '-' to mark as inactive

r888888888:


evazion:
There's an xss vuln here in `var $del = $("<del/>").html(desc);`:

https://github.com/r888888888/danbooru/blob/fcd80b6043830e7234cd37588a665efed1cb4fba/app/assets/javascripts/related_tag.js.erb#L233-L240

Exploit: make an artist entry with these URLs then edit a post with source `http://a.b/`:

* `http://a.b/`
* `-http://a.b/<script>alert(1)</script>`.

Also, inactive URLs aren't prepended with '-' in the edit form, so inactive URLs will be made active again if someone edits the artist and doesn't add '-' signs manually.

PR #3717, opened 2018-05-16
eliminate usage of localstorage to cache tag autocomplete results (fixes #3543)

### LINKED ISSUE #3543: Use IndexedDB as alternate caching mechanism
The [prototype autocomplete script developed for Danbooru](https://danbooru.donmai.us/forum_topics/14701) has been stable for at least a month, and there are still no issues with size even after 3 months of usage with no pruning, even though it currently sits at above 10M characters (20MB).  This is quite the improvement over the space-limited LocalStorage with a current set maximum size of 2.5M characters (5MB).  Additionally, the performance characteristics for IndexedDB were very close to that of LocalStorage, usually on the order of 10 ms versus 1 ms.  Therefore, I'd like to recommend transitioning Danbooru to supporting IndexedDB on those browsers that support it.

I've been using [LocalForage](http://localforage.github.io/localForage/) in my script as it offers several advantages.  The first advantage being is that it offers a fallback scheme for those browsers that don't support IndexedDB, the first fallback being LocalStorage and the second fallback being no storage.  It does offer WebSQL, but my own testing saw poor latency with that scheme, usually on the order of 200 ms.  Another advantage is that it offers a similar simple API to that which is currently offered by LocalStorage, making it easy to program with.  Most of the functions are non-blocking and asynchronous, however it does supply the usage of either promises or callbacks for this.

The following are several other items of notice in regards to the prototype script.

1. Many more API calls are being stored
2. SessionStorage used to mitigate time penalty of IndexedDB
3. Variable expiration durations of 1 week to 1 month based on the count and/or variability of the data
4. Full validation of client-side data

Although it wasn't implemented yet, it would also be recommended to perform a regular pruning action on IndexedDB for expired entries, although since it's not as critical, this could be set to be performed once a day instead.  Also, the current pruning code for LocalStorage would still be used for those browsers that have LocalStorage as a fallback.

r888888888:


r888888888:
I don't think there's any need to cache artist/wiki/user/etc.

BrokenEagle:
There are a few pros/cons to adding support for artist/wiki/user/etc.

## Pros

1. Cuts down on network requests and therefore load to the server

No need to explain the above.

2. Avoids repeat requests for various corrections, other patterns

Someone types out an item, finds a mistake, deletes several characters, then types out some more.  The previous process could be repeated several times.  Under the current scheme, each backtrack would need to send the same network request that has already been sent.

Also at least for myself, I search for many of the same things.  Having a cache makes those common patterns of searching become faster over time.

3. Reduces network latency experienced by the user

This can be even more noticeable and detrimental when Danbooru is being very slow.

## Cons

1. Data can be stale

Though for much of the different types, the data on average doesn't change much, especially not over the period of several days.

2. More code to maintain

There's really no getting around this one.  It's all about weighing the benefits versus the costs.

## Final

The above are just my thoughts though.  I'd be interested to hear what other people think.

r888888888:
I'll try it. Deployed to testbooru.

evazion:
There's a typo in the users controller, it should be `@users.to_json` not `@user.to_json`.

Caching in this way would mean all API calls get cached, not just autocomplete calls. If you're doing something like scraping `/artists.json` regularly you could get stale responses. That's why when I tested this I made the cache duration an url param, so that the caller can control how API responses are cached.

PR #3746, opened 2018-06-16
Post Flagger Report

r888888888:
A report to list anyone who might be targeting a user for flagging, assuming that a 50% target rate or more (with 95% confidence) is abnormal.


Ion288:
Im flagging a lot of old single page comics that have Albert as uploader. I understand they are mostly uploaded by anonymous. Will this be a problem?

chinatsu-chan:
I'm concerned posts I flagged which were "resolved" but then reflagged by another user and deleted will still damage ratios.

r888888888:
Those old single page comics are so far back in Danbooru's history that they probably don't get seen in most people's searches. I wouldn't worry about flagging them. If they bother you, then downvote them.

I don't see why reflags should alter the calculation. The fact remains that you're targeting a user.

PR #3790, opened 2018-07-28
Retirement system for underutilized aliases+implications

r888888888:


r888888888:
For the record this is the result of the dry run: https://gist.github.com/r888888888/fedf8cbc204fdd8a8c82715cbd64eca7

PR #3805, opened 2018-08-08
Refactor sources

r888888888:
This is an attempt to unify + refactor the `RewriteStrategies` and `Sources::Strategies` code paths. The current naming scheme of rewrites, referers, and normalized urls is confusing. There should be only two canonical urls at play: a url to the binary file, and a url to the HTML page.


evazion:
One thing about the source field: the value it should be set to after upload isn't always the page url. The preferred source depends on the site: for sites like Pixiv or DeviantArt it's the image url, for sites like Twitter or Tumblr it's the page url.

Basically, if a site has a rewrite in `Post#normalized_source`, then the source should be left as the image url, otherwise it should be set to the page url.

r888888888:
On the plus side I've figured out a way to obtain metadata for Fanbox and Novel images. On the down side, I'm deprecating profile images/backgrounds because these don't seem to have API endpoints, and they haven't been uploaded at all recently.

test/unit/sources/deviantart_test.rb
```diff
@@ -31,25 +29,24 @@ def setup
...
     context "The source for a DeviantArt image url" do
       should "fetch the source data" do
-        @site = Sources::Site.new("https://pre00.deviantart.net/b5e6/th/pre/f/2016/265/3/5/legend_of_galactic_heroes_by_hideyoshi-daihpha.jpg")
+        @site = Sources::Strategies.find("https://pre00.deviantart.net/b5e6/th/pre/f/2016/265/3/5/legend_of_galactic_heroes_by_hideyoshi-daihpha.jpg")
 
         assert_equal("hideyoshi", @site.artist_name)
-        assert_equal("https://hideyoshi.deviantart.com", @site.profile_url)
-        assert_equal("https://orig00.deviantart.net/9e1f/f/2016/265/3/5/legend_of_galactic_heroes_by_hideyoshi-daihpha.jpg", @site.image_url)
+        assert_equal("https://www.deviantart.com/hideyoshi", @site.profile_url)
+        assert_equal("https://pre00.deviantart.net/b5e6/th/pre/f/2016/265/3/5/legend_of_galactic_heroes_by_hideyoshi-daihpha.jpg", @site.image_url)
```
evazion: Here the `orig00.deviantart.net` url was the correct image url to use. `pre00.deviantart.net/b5e6/th/pre/` is a sample url.

r888888888:
The true source for 1417494 is `http://i2.pixiv.net/img46/img/fm59/35673338.jpg`. The moniker is embedded in the URL, although that isn't necessarily the identifier used on Danbooru.

The true source for 3226836 is `https://i.pximg.net/img-original/img/2018/08/05/18/00/06/70034302_p0.jpg`. There's no reference to the moniker.

Based on the usage of the `pixiv_id` field on `posts` I think the general trend has been to reference works by their `illust_id` or equivalent. It is possible to change the older URLs to reference the illust id if consistency is the goal.

PR #3847, opened 2018-08-28
Intelligent autocomplete

r888888888:
This adds, among other things, fuzzy matching and prefix matching.

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -10,6 +10,7 @@ class TagsController < ApplicationController
 
   def index
     @tags = Tag.search(search_params).paginate(params[:page], :limit => params[:limit], :search_count => params[:search])
+
     respond_with(@tags) do |format|
       format.xml do
         render :xml => @tags.to_xml(:root => "tags")
@@ -18,7 +19,13 @@ class TagsController < ApplicationController
   end
 
   def autocomplete
-    @tags = Tag.names_matches_with_aliases(params[:search][:name_matches])
+    if CurrentUser.is_builder?
+      # limit rollout
+      @tags = TagAutocomplete.search(params[:search][:name_matches])
+    else
+      @tags = Tag.names_matches_with_aliases(params[:search][:name_matches])
+    end
+
     expires_in params[:expiry].to_i.days if params[:expiry]
 
     respond_with(@tags) do |format|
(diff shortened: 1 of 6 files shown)


chinatsu-chan:
My experience with this so far has been very positive

evazion:
I've experimented with ranking exact matches first before, like in `count_sort`. In my experience it didn't work too well. Tags like [white](https://danbooru.donmai.us/posts?tags=white) or [black](https://danbooru.donmai.us/posts?tags=black) get in the way of a lot of searches. When you're typing something like `animal_ears`, you hit a bunch of small tags along the way: [a](https://danbooru.donmai.us/posts?tags=a), [an](https://danbooru.donmai.us/posts?tags=an), [anima](https://danbooru.donmai.us/posts?tags=anima), [animal](https://danbooru.donmai.us/posts?tags=animal). Danbooru has a very long tail of small tags that often get in the way.

The prefix anchoring approach looks similar in concept to [Jaro-Winkler](https://en.wikipedia.org/wiki/Jaro%E2%80%93Winkler_distance#Jaro.E2.80.93Winkler_Similarity), which also tries to rank matches higher when they start with a common prefix.

r888888888:
The prefix search will let you enter something like `osf` and it will match `okita_souji_(fate)`, in effect enabling aliases for many multi-word tags. SublimeText does something similar and I've always found it extremely useful for navigating between files.

evazion:
Ah, in that case I'd suggest a couple things:

* Consider treating `-`, `:`, and `;` as word boundaries too (think: `side-tie_bikini`, `re:zero`, `steins;gate`).
* Consider requiring a slash to trigger it: `/osf`. That would avoid issues with short strings accidentally triggering it (think: ass, hat, bow, day, sky). It would also allow existing shortcut aliases to be replaced with this system.
* Some abbreviations like `/be` may be ambiguous, so allow saying something like `/br.e` for `brown_eyes` and `/bl.e` for `blue_eyes`.

r888888888:
1 is simple.

I would like to see how this pans out in practice before considering 2. The prefix search is disabled once the string reaches 5 characters (maybe it should be 4, but that decreases the utility). This may end up being necessary anyway for performance reasons.

3 would probably make the regexp more complicated but it may be worth investigating next.

PR #3852, opened 2018-08-29
Twitter: fix handling of direct image urls without a referer url.

evazion:
This fixes the Twitter strategy to not error out when given a direct image url but not a referer url. This can happen when uploading a direct image url from Twitter without the bookmarklet.

### DIFF
--- a/app/logical/sources/strategies/twitter.rb
+++ b/app/logical/sources/strategies/twitter.rb
@@ -24,6 +24,8 @@ module Sources::Strategies
     def image_urls
       if url =~ /(#{ASSET}[^:]+)/
         return [$1 + ":orig" ]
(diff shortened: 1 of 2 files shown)


app/logical/sources/strategies/twitter.rb
```diff
@@ -47,49 +47,59 @@ def page_url
     end
 
     def profile_url
+      return "" if api_response.blank?
```
r888888888: This should probably go on line 55. This strategy will sometimes be used by ArtistUrl to normalize profile urls, so sometimes `url` will be the profile url.

evazion: Should be fixed in 6c94047. The artist url test for this now passes.

PR #3886, opened 2018-09-11
cache api clients

r888888888:


app/logical/httparty_cache.rb
```diff
@@ -0,0 +1,11 @@
+module HttpartyCache
+  extend self
+
+  def get(url, headers: {}, params: {}, base_uri: nil, format: :html, expiry: 60)
+    Cache.get("cachedget:#{Cache.hash(url)}", expiry) do
```
lllusion3469: Wouldn't you want to consider (hash) the `params` (and maybe the headers) as well, not just the `url`?
e.g. [here](https://github.com/r888888888/danbooru/blob/b4b03ed7732d17332e568f7dfbbb147f576b8bde/app/logical/tumblr_api_client.rb#L4)

or do you think it's reasonable to assume that nobody's going to upload 2 tumblr post (going with tumblr as an example again) from the same artist within 60s?

or am I just reading the code wrong?

[another example](https://github.com/r888888888/danbooru/blob/b4b03ed7732d17332e568f7dfbbb147f576b8bde/app/logical/deviant_art_api_client.rb#L46), this time affecting every da upload, I think

Unrelated note: isn't [this](https://github.com/r888888888/danbooru/blob/b4b03ed7732d17332e568f7dfbbb147f576b8bde/app/logical/sources/strategies/deviant_art.rb#L249) kind of pointless now?

Just tested it and that's exactly what happens:
```
$ bundle exec rails console
[1] pry(main)> da = DeviantArtApiClient.new(Danbooru.config.deviantart_client_id, Danbooru.config.deviantart_client_secret, Danbooru.config.httparty_options)
pry(main)> da.metadata(["BA286AEF-D6E2-8765-490E-4720153E4F44"])
... [too long]
[3] pry(main)> da.metadata(["5D9B17AB-DBA1-684D-BB0F-6CDBF12FE218"]) # different uuid
... [removing irrelevant parts]
  [{:deviationid=>"BA286AEF-D6E2-8765-490E-4720153E4F44",  # [same uuid as the first request above]
...
```

r888888888: You're right. I'll have to expand the key.

Issue #3900, opened 2018-09-14
Allow to search for urls associated with artists using wildcards


nonamethanks:
There's currently several of us who employ bots to keep artist pages up to date. It would greatly simplify our job (and improve the artist interface) if there was a way to search for artist urls by domain or wildcard.

The best example for this would be if someone wanted to scan all pixiv/twitter/insert-domain-here urls and mark the dead ones as deleted - there's currently no way to do so via the api without looping through the entire artist database. Same thing if a new kind of permanent url is discovered (such as was the case for the twitter intent url scheme) and one has to go through the tens of thousands of entries to check where it's applicable.

The api page says that the only way to currently search for artist urls is:
```
search[name] This field has multiple uses depending on what the query starts with:
    http  Search for artist with this URL.
```
As far as I could test it only works with exact urls, and it falls apart once one uses wildcards.

There's also search[url_matches] apparently, though it's undocumented in the wiki page, but that too only works with perfect matches.

BrokenEagle:
Not sure if this would be possible as well, but having access to regular expressions for URL matching could also prove useful.

If either that or the wildcard matching as per the OP is a database expensive operation, then maybe just limit it to Builder+.

BrokenEagle:
Probably a text input for any/all of the above.  Either overload the **URL** field with the extra capability by detecting "*"s, or providing a separate input for that.  The domain name search would require its own input though, but could be overloaded to also handle wildcards.  I don't see the need to search for root-level domains (net, com, org), so check for the presence of a period.

Since the artist index is getting kind of full, maybe move all of the search options to their own dedicated search page.

As a similar but separate issue, perhaps finally deprecate the ``search[name]`` input and provide individual text inputs for all of those options.  It's been at least a year since those individual options have been made available (d1188d8184053ccfe849147a9de408f996c7d324).

evazion:
I'm thinking this for API searches:

```
# artists containing an active pixiv url:
http://danbooru.donmai.us/artists?search[artist_url][url_regex]=pixiv&search[artist_url][is_active]=true
```

and overload the URL field for manual searches. WIP: https://github.com/r888888888/danbooru/compare/master...evazion:fix-3900

BrokenEagle:
I like the above idea for regexes, both URL and/or domain.  If it's limited to the API, then only those that "probably" know what they're doing would have access to it.  No need to confuse around 99.9... percent of the user base.

r888888888:
[...]

PR #3907, opened 2018-09-16
#3900: Allow to search for urls associated with artists using wildcards

evazion:
Partial fix for #3900: 

* Adds an `/artist_urls` index page. This is mainly intended for api use, but there's a basic html page too.

* Adds `search[url_matches]=...` and `search[normalized_url_matches]=...` params on the `/artist_urls` page. These params are overloaded such that if the input looks like `/pixiv/` it does a regex search, otherwise if it looks like `*pixiv*` it does a wildcard search, otherwise it does an artist finder search:

```
# regex search
/artist_urls?search[normalized_url_matches]=/pixiv/

# wildcard search
/artist_urls?search[normalized_url_matches]=*pixiv*

# artist finder search - normalize to the profile url and search for the profile url instead.
/artist_urls?search[normalized_url_matches]=https://www.pixiv.net/member_illust.php%3Fmode=medium%26illust_id=46503128
=> /artist_urls?search[normalized_url_matches]=http://www.pixiv.net/member.php%3Fid=9948/
```

* Adds the following search params:

```
/artist_urls?search[url]=...            # matches exactly
/artist_urls?search[url_eq]=...         # matches exactly
/artist_urls?search[url_not_eq]=...     # does not match exactly
/artist_urls?search[url_like]=...       # matches wildcard (case sensitive)
/artist_urls?search[url_ilike]=...      # matches wildcard (case insensitive)
/artist_urls?search[url_not_like]=...   # does not match wildcard (case sensitive)
/artist_urls?search[url_not_ilike]=...  # does not match wildcard (case insensitive)
/artist_urls?search[url_regex]=...      # matches regex
/artist_urls?search[url_not_regex]=...  # does not match regex

/artist_urls?search[normalized_url]=... # all the same options as above
```

These options are provided for API use. Some of these things are more useful than others, but I went ahead and provided them all for completeness.

The method that handles this (`ApplicationRecord#search_text_attribute`) is designed to be generic, to make it possible to support these kind of searches in other places on the site too.

### DIFF
--- a/app/views/artist_urls/index.html.erb
+++ b/app/views/artist_urls/index.html.erb
@@ -1,0 +1,44 @@
+<div id="c-artist-urls">
+  <div id="a-index">
+    <%= simple_form_for(:search, url: artist_urls_path, method: :get, defaults: { required: false }, html: { class: "inline-form" }) do |f| %>
+      <%= f.input :is_active, label: "Active?", collection: [["Yes", true], ["No", false]], include_blank: true, selected: params[:search][:is_active] %>
+      <%= f.input :order, collection: [["ID", "id"], ["Created", "created_at"], ["Updated", "updated_at"]], selected: params[:search][:order] %>
+      <%= f.submit "Search" %>
+    <% end %>
+
+    <table class="striped" width="100%">
+      <thead>
+        <tr>
+          <th>ID</th>
+          <th>Artist Name</th>
+          <th>URL</th>
+          <th>Normalized URL</th>
+          <th>Active?</th>
+          <th>Created</th>
+          <th>Updated</th>
+        </tr>
+      </thead>
+      <tbody>
+        <% @artist_urls.each do |artist_url| %>
+          <tr>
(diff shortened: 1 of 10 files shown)


app/views/artist_urls/index.html.erb
```diff
@@ -0,0 +1,44 @@
...
+        <% @artist_urls.each do |artist_url| %>
+          <tr>
+            <%= tag.td artist_url.id %>
+            <%= tag.td link_to(artist_url.artist.name, artist_url.artist) %>
+            <%= tag.td external_link_to(artist_url.url.to_s) %>
+            <%= tag.td external_link_to(artist_url.normalized_url) %>
+            <%= tag.td artist_url.is_active.to_s %>
+            <%= tag.td artist_url.created_at %>
+            <%= tag.td artist_url.updated_at %>
+          </tr>
```
BrokenEagle: Should we add some controls here that allow the marking of a URL as inactive, like what is found on the Post/Show page?

Issue #3947, opened 2018-10-06
Flagged/pending posts aren’t being deleted after 3 days


crs00:

evazion:
It's hard to tell whether daily maintenance is erroring out or not running at all. #3951 adds logging to help debug this.

Issue #3413, opened 2017-11-30
Certain daily maintenance tasks not running


evazion:
Certain daily maintenance tasks don't seem to be running:

* Post disapprovals aren't being pruned: http://danbooru.donmai.us/posts?tags=status%3Adeleted+age:>40d

* Negative tag counts aren't being fixed: https://danbooru.donmai.us/tags?limit=500&page=1388&search[hide_empty]=no&search[order]=count.

* Cached post counts on tag aliases aren't being updated. Some aliases, `saber_extra -> nero_claudius_(fate)` for example, don't show up during autocomplete because their cached post count is zero. See http://danbooru.donmai.us/tag_aliases/15127.json.

* The [/cache/tags.json](http://danbooru.donmai.us/cache/tags.json) dump appears out of date. The most recent tag is from 2017-10-09.

I think generating `/cache/tags.json` may be the culprit. If there's some exception writing that file then the remaining tasks may not run.

BrokenEagle:
Might it be useful to write a mod action for the daily maintenance task for cases like this?  The list of functions in it could be wrapped in an exception handler that writes a mod action upon an error.  With it more visible like this, it would allow users to more easily catch when there is a disruption in the process instead of having to rely on secondary or tertiary effects.

evazion:
Any kind of logging would help. Another idea would be to run tasks as delayed jobs, that way failures would show up in the delayed jobs index. It would also mean that a failure in one task wouldn't prevent others from running.

r888888888:
Running it manually seems to have worked, including all the steps before it up to the upload prune. Maybe there was a bug that was fixed in the recent deployment.

PR #3951, opened 2018-10-08
Improve daily maintenance error handling

evazion:
Related to #3947 and #3413. This arranges for exceptions during hourly/daily/weekly maintenance tasks to be logged to NewRelic.

This works by changing `config/schedule.rb` to run `rake maintenance:daily` instead of running `DailyMaintenence.new.daily` directly. NewRelic will automatically instrument rake tasks, as long as you add this to your `config/newrelic.yml` to enable it:

```
  # https://docs.newrelic.com/docs/agents/ruby-agent/background-jobs/rake-instrumentation
  rake:
    tasks: ["maintenance:.*"]
```

After that, maintenance runs will show up in NewRelic Insights like this:

image

And exceptions will show up like this:

image

### DIFF
--- a/app/logical/maintenance.rb
+++ b/app/logical/maintenance.rb
@@ -1,5 +1,12 @@
-class DailyMaintenance
-  def run
+module Maintenance
+  module_function
+
+  def hourly
+    UploadErrorChecker.new.check!
+    DelayedJobErrorChecker.new.check!
+  end
+
+  def daily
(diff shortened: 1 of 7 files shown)


r888888888:
This needs to be rebased. I got rid of the hourly maintenance.

evazion:
@r888888888 this has been rebased.

Issue #3987, opened 2018-11-13
Convert more columns to arrays


evazion:
As with #3979, here are some other columns that should be stored as arrays instead of strings:

* `artists.other_names`
* `artist_versions.other_names`
* `artist_versions.url_string`
* `wiki_pages.other_names`
* `wiki_page_versions.other_names`
* `tag_implications.descendant_names`

BrokenEagle:
What about many of the post columns?

- ``posts.tag_string``
- ``posts.tag_string_general``
- ....all other post tag strings...
If the above were done, then all of the **tag_count** columns could be removed and replaced with a function that just returns the size of the array.

- ``posts.pool_string``
Would allow the creation of a **pool_count** variable, as well as ``pools:>=4``.

- ``posts.fav_string``
The **fav_count** column could be removed and replaced with a function.

evazion:
Those definitely should be arrays too. Those will be more involved though, so I plan on tackling them last.

I think it would technically be possible to drop the pool_string and fav_string columns entirely. This would save a lot of space in the posts table, but it would involve some other tradeoffs.

PR #3989, opened 2018-11-14
Wiki pages: convert other_names column to array (#3987)

evazion:
Addresses #3987 for wiki pages. Converts `wiki_pages.other_names` and `wiki_page_versions.other_names` to arrays.

### DIFF
--- a/app/controllers/pools_controller.rb
+++ b/app/controllers/pools_controller.rb
@@ -94,7 +94,7 @@ class PoolsController < ApplicationController
   private
 
   def pool_params
-    permitted_params = %i[name description category is_active post_ids]
+    permitted_params = %i[name description category is_active post_ids post_ids_string]
     params.require(:pool).permit(*permitted_params, post_ids: [])
   end
 end
--- a/app/controllers/wiki_pages_controller.rb
+++ b/app/controllers/wiki_pages_controller.rb
@@ -5,7 +5,7 @@ class WikiPagesController < ApplicationController
   before_action :normalize_search_params, :only => [:index]
   
   def new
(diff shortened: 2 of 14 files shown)


evazion:
Converted the remaining columns to arrays (artist_versions.other_names, artist_versions.url_string, artists.other_names, tag_implications.descendant_names).

This incidentally fixes a bug with searching for other names on the artists page. Before if you searched the name `foo*` it incorrectly matched against the entire other_names string; now it matches against each individual other name.

evazion:
API changes:

* /wiki_pages.json: `other_names` is now an array.
* /tag_implications.json: `descendant_names` is now an array.
* /artist_versions.json: `other_names` is now an array; `url_string` is now 'urls' and is an array.
* /artists.json: `other_names` is now an array.
* PUT /artists/1234.json: the `other_names_comma` parameter was removed. Instead use `/artists/1234.json?other_names=one two three` and separate names by spaces.

BrokenEagle:
For that last one, does that mean that we are now eliminating artist other names with spaces in them? I've encountered those in the past. Are we supposed to convert those spaces to underscores now?

evazion:
Yes, when setting other_names via the API, spaces should be replaced with underscores. For example:

* old way: `PUT /artists/1234.json?artist[other_names_comma]=first name, second name`
* new way: `PUT /artists/1234.json?artist[other_names]=first_name second_name`

The new way actually works right now, it was just never documented.

Internally, the other_names field has always been separated by spaces, not commas, which you can see with something like https://danbooru.donmai.us/artists/178741.xml. So using commas as the separator here was weird and not very consistent with the rest of the site.

PR #4000, opened 2018-12-03
add dynamic tag counts on upload form

r888888888:


evazion:
Does this need to be tied to autocomplete? It won't trigger if the user disables autocomplete in their account settings.

app/javascript/src/javascripts/posts.js.erb
```diff
@@ -598,6 +602,27 @@ Post.initialize_saved_searches = function() {
...
+    }
+  }
+  $("#tags-container .count").html(string);
+  if (count < 25) {
+    $("#tags-container .options #face").removeClass().addClass("far fa-frown");
+  } else if (count < 35) {
+    $("#tags-container .options #face").removeClass().addClass("far fa-meh");
+  } else {
+    $("#tags-container .options #face").removeClass().addClass("far fa-smile");
+  }
```
konovalov-nk: Just a suggestion. Since we have `webpacker` and already using ES6+ syntax in some places, then we could write a little bit less code

```es6
let class_emoji = "smile";
if (count < 35) {
  class_emoji = count < 25 ? "frown" : "meh";
}
$("#tags-container .options #face").removeClass().addClass(`far fa-${class_emoji}`);
```

app/javascript/src/javascripts/posts.js.erb
```diff
@@ -35,10 +35,14 @@ Post.initialize_all = function() {
     }
   }
 
-  if ($("#image").length) {
+  if (true || $("#image").length) {
```
r888888888: todo: delete this

PR #4007, opened 2018-12-12
Fix #4004: Add additional order by metatags for posts

### LINKED ISSUE #4004: Add additional order by metatags for posts
Brought up in [topic #15787](https://danbooru.donmai.us/forum_topics/15787) with a modicum of support.

Mentioned in the topic is order post by comment count. After thinking about it a bit, a couple other order metatags that might prove interesting would be order by pool count and order by note count. Others that might prove both interesting and useful would be flag count, appeal count, and approval count. A stretch goal would be tagger count (i.e. how many users have contributed to the tags on a post), however since that involves the post versions table, that may not be very resource friendly.

So all told:

- ``order:comment_count``
- ``order:pool_count``
- ``order:note_count``
- ``order:flag_count``
- ``order:appeal_count``
- ``order:approval_count``
- ``order:tagger_count``

I'm not sure the reverse order would be interesting/useful for any of the above orderings though.

evazion:
Fixes #4004. This adds numerous metatags for searching by various kinds of counts:

* `comments`, `active_comments`, `deleted_comments`
* `notes`, `active_notes`, `deleted_notes`
* `flags`, `resolved_flags`, `unresolved_flags`
* `children`, `active_children`, `deleted_children`
* `pools`, `active_pools`, `deleted_pools`, `series_pools`, `collection_pools`
* `appeals`
* `approvals`
* `replacements`

All metatags can be used both for ordering (e.g. `order:notes`) and for searching (`notes:>5`). Ordering supports both ascending and descending directions (`order:notes_asc`, `order:notes_desc`). Negated searches aren't supported (`-notes:5` doesn't work).

`*_count` can be used as a synonym for any of these tags. For example: `order:note_count`, `child_count:>5`.

Most of these searches will be timeout-prone unless constrained by other tags, or by id or date ranges (`id:>3000000`, `age:<1y`). But they will work for smaller searches, and there are already other metatags that tend to timeout unless constrained (`source:*blah*`, `order:score`), so I don't think it makes things worse.

This is what one of these searches boils down to:

```ruby
Post.tag_match("comments:>5 order:notes")

Post.with_comment_stats.with_note_stats.where("comment_count > 5").order("note_count DESC, id 
DESC")

SELECT
  "posts".*
FROM (
  SELECT
    posts.*,
    COUNT(comments.id) AS comment_count,
    COUNT(comments.id) FILTER (WHERE comments.is_deleted = TRUE) AS deleted_comment_count,
    COUNT(comments.id) FILTER (WHERE comments.is_deleted = FALSE) AS active_comment_count,
    COUNT(notes.id) AS note_count,
    COUNT(notes.id) FILTER (WHERE notes.is_active = TRUE)  AS active_note_count,
    COUNT(notes.id) FILTER (WHERE notes.is_active = FALSE) AS deleted_note_count
  FROM "posts"
  LEFT OUTER JOIN "comments" ON "comments"."post_id" = "posts"."id"
  LEFT OUTER JOIN "notes" ON "notes"."post_id" = "posts"."id"
  GROUP BY "posts"."id"
) posts
WHERE
  (posts.comment_count > 5)
ORDER BY "posts"."note_count" DESC, "posts"."id" DESC
LIMIT 20
OFFSET 0
```

### DIFF
--- a/app/javascript/src/javascripts/autocomplete.js.erb
+++ b/app/javascript/src/javascripts/autocomplete.js.erb
@@ -5,6 +5,7 @@ let Autocomplete = {};
 
 Autocomplete.METATAGS = <%= Tag::METATAGS.to_json.html_safe %>;
 Autocomplete.TAG_CATEGORIES = <%= TagCategory.mapping.to_json.html_safe %>;
+Autocomplete.ORDER_METATAGS = <%= Tag::ORDER_METATAGS.to_json.html_safe %>;
 
 Autocomplete.TAG_PREFIXES = "-|~|" + Object.keys(Autocomplete.TAG_CATEGORIES).map(category => category + ":").join("|");
 Autocomplete.TAG_PREFIXES_REGEX = new RegExp("^(" + Autocomplete.TAG_PREFIXES + ")(.*)$", "i");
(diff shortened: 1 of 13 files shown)


evazion:
@r888888888 have you had a change to review this? Admittedly it's going to timeout for large searches, but I think it should be usable for small searches.

I could switch to caching the counts in a materialized view, which would make it very fast, at the cost of results not always being up to date.

PR #4076, opened 2019-04-14
Facilitate autocomplete queries by tag category

### LINKED ISSUE #4062: Tag completion should limit results based on category prefix
I don't know the name for these tags that begin with char:, art:, artist:, etc., however if one is used in the search bar or in the tag field, results should be limited to that tag category. So if you use one of these prefixes and begin typing, the suggested tags should be of only that category of tag.

BrokenEagle:
Fixes #4062. As discussed on Discord with @evazion, when in a tag edit interface, it makes sense to not do a category search as the tag edit box can be used to change a tag's category.  As such, it fixes the post query so that tag category prefixes can now be used.

Additionally, it fixes an issue where queries with just the tag category prefix was sending blank queries. As part of the fix, it also now checks for valid tag category IDs when using the **Tag** ``search`` function.

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -19,11 +19,12 @@ class TagsController < ApplicationController
   end
 
   def autocomplete
+    category = (params[:search][:category].present? ? params[:search][:category] : -1)
     if CurrentUser.is_builder?
       # limit rollout
-      @tags = TagAutocomplete.search(params[:search][:name_matches])
+      @tags = TagAutocomplete.search(params[:search][:name_matches], category)
     else
-      @tags = Tag.names_matches_with_aliases(params[:search][:name_matches])
+      @tags = Tag.names_matches_with_aliases(params[:search][:name_matches], category)
     end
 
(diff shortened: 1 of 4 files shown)


BrokenEagle:
After thinking about it a bit more, it'd probably be better to make the search category parameter optional so that it would be backwards compatible with userscripts, such as my own IndexedAutocomplete script.  I updated it just now with that change.

r888888888:
Is this still relevant? I must have missed the notification.

BrokenEagle:
Are you asking if this capability is still needed or if it's still good to merge?

If it's the former, then that depends on the user. I'm not sure I've ever had a situation where I would need to restrict autocomplete results by category, as I usually just know the name of the tag I'm searching for. Pinging @ampersham since he made the issue this is based off of.

If it's the latter, then it will take a while for me to get my instance current and working again to test it out. Or it could be deployed to Testbooru, which would give other users a chance to trial it out and comment on its use or need.

BrokenEagle:
Is this needed anymore, or should it be closed?

Issue #4079, opened 2019-04-22
Allow deleted forum items to be seen with the is_deleted parameter


BrokenEagle:
Deleted comments are shown to all users when the **is_deleted** parameter is used.  The same should be true for forum posts and forum topics.  I can't think of any reason why they shouldn't be seen, as long as they're not the moderator-only topics/posts.

PR #4242, opened 2020-01-04
Add API data to show/index views

### LINKED ISSUE #4182: Add API data to HTML
On many pages data has to be scraped from rendered IDs or classes in order to get information needed like in a userscript. This is very tenuous as any change to site HTML structure can break things.

This could be done simply by adding a hidden HTML structure that contains the JSON text. Then ensure that the ID is on the rendered HTML elements. For tables, this would probably be best on the `<tr>` elements with something like `data-id="INSTANCEID"`.

BrokenEagle:
Fixes #4182. To convert to the table format, some logic was taken out of the index controllers, such as the min level and deleted checks on forums. The min level doesn't matter since it's already being handled by the model. For the deleted check, it was moved to the model as well.

Also, there was a bug where deleted forum items weren't shown even though the `search[id_deleted]=true` parameter was present, for both members and moderators. This was supposed to have been fixed in #4079, but that didn't change the logic in the index views so the problem still remained.


BrokenEagle:
Did a small fixup of the fixing `:inviter` to `:inviter_id` on the user model.

BrokenEagle:
Per our conversations on discord with @evazion, I changed the add API data commit to use the body element instead of adding a container to every show view. Also, to deconflict with models that had an actual `:user_id` data attribute, I prefixed all of the model API data with "show".

BrokenEagle:
Did a small fixup where I removed the now unused **html_data_attributes_for** function.

app/views/tag_aliases/_listing.html.erb
```diff
@@ -1,44 +1,34 @@
...
+  <% t.column "Status", {width: "5%"} do |tag_alias| %>
+    <span id="tag-alias-status-for-<%= tag_alias.id %>"><%= tag_alias.status %></span>
+  <% end %>
+  <% t.column "", {width: "15%"} do |tag_alias| %>
+    <%= link_to "Show", tag_alias_path(tag_alias) %>
 
-          <% if tag_alias.editable_by?(CurrentUser.user) %>
-            | <%= link_to "Edit", edit_tag_alias_path(tag_alias) %>
-          <% end %>
+    <% if tag_alias.is_pending? && tag_alias.editable_by?(CurrentUser.user) %>
```
evazion: Is adding the `tag_alias.is_pending?` check necessary here? Wasn't present before and it should already be handled by `editable_by?` (see `editable_by?` in `app/models/tag_relationship.rb`).

evazion: Oh, I see it was for symmetry with tag implications. In that case it's fine (both checks could be removed in a later commit though).

evazion:
> Also, to deconflict with models that had an actual :user_id data attribute, I prefixed all of the model API data with "show".

I think it would be better to change the prefix for current user attributes to `data-current-user-` and use `data-{model_name}-` for the show page API data. This would be a breaking change but it makes the attribute names cleaner.

app/models/forum_post.rb
```diff
@@ -73,7 +73,14 @@ def search(params)
     end
   end
 
+  module ApiMethods
+    def html_data_attributes
+      [:topic_id, :creator_id, :updater_id, :is_deleted?, [:topic, :is_deleted?]]
```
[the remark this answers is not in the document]
BrokenEagle: As discussed on Discord, this was already a part of the forum post index page.

PR #4247, opened 2020-01-07
Add pool version enhancements

### LINKED ISSUE #4224: Pool versions enhancements
1. Add a name changed column to the pool versions index (HTML)

    Currently there's no way to tell if a pool name has changed just from the index alone. One has to actually go to the index by pool ID, which can be time-consuming, especially for those that just want to browse the index for potential issues like abuse or mistranslations.

2. Add a search page for pool versions

    Currently all search terms have to be manually added, something most regular users cannot do. Besides that, add a search term for the first version of a pool, or every version but the first version. That way, searching for name changes or description changes becomes more productive since the vast majority of pool versions with those set to true are the first version.

3. Have a better diff page

    Visually comparing the old and new versions of the description is extremely hard, especially if the description is particularly long or the change is only minor. Instead it should be something like what the wiki page version diff does, where it superimposes the two versions and highlights the additions green and the removals red. I did a userscript mock-up of this with a Gist a while back.

    https://gist.github.com/BrokenEagle/a6e4fdd79aa8b783ec1a78796330654d

    One change that the above adds beyond the wiki page version diff is that it shows paragraph marks. That way the addition or removal of carriage returns becomes better highlighted.  Besides that though, maybe also highlight the name changes if there are any, since the post changes are already shown as well.

BrokenEagle:
Fixes #4224. Instead of adding another column, the pool versions index was made to be more like with wiki page versions index.  This also allows for multiple other status changes. Also, the logic for doing the diff for wiki page versions was moved out into a logic module so that it could be available for pool versions.

### DIFF
--- a/app/controllers/pool_versions_controller.rb
+++ b/app/controllers/pool_versions_controller.rb
@@ -12,6 +12,9 @@ class PoolVersionsController < ApplicationController
     respond_with(@pool_versions)
   end
 
+  def search
+  end
+
   def diff
     @pool_version = PoolArchive.find(params[:id])
 
--- a/app/helpers/pool_versions_helper.rb
+++ b/app/helpers/pool_versions_helper.rb
@@ -18,4 +18,9 @@ module PoolVersionsHelper
     status += ["Deactivated"] if !cur.is_active? && prev.is_active?
     status.join(" ")
   end
+
+  def pool_page_diff(pool_version, other_version)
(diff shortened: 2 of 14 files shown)


BrokenEagle:
Did a small fixup resolving a merge conflict in the pool versions listing view. Also reworded some of the commits for better clarity. There were some merge artifacts left over from that, so I had to fix that as well.

PR #4249, opened 2020-01-07
Have artist wiki links go to artist page

### LINKED ISSUE #4205: Redirect artist wiki links to their artist page instead of the wiki page
Most times the wiki page doesn't even exist, and one can edit the wiki from the artist page. This usually means that one has to go through 2 page loads just to get to the artist entry. Also, the post page already redirects to the artist page instead of the wiki page for the artist, so this would follow the same behavior.

However, have the wiki page still be accessible from the artist page or when using the wiki page interface. The only thing that needs to be changed are the artist DText links.

BrokenEagle:
Fixes #4205. It also checks for existing artist pages, similar to it does for wiki pages.

### DIFF
--- a/app/javascript/src/styles/common/dtext.scss
+++ b/app/javascript/src/styles/common/dtext.scss
@@ -109,7 +109,7 @@ div.prose {
     vertical-align: 1px;
   }
 
-  a.dtext-wiki-does-not-exist, a.dtext-tag-does-not-exist {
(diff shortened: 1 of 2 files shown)


evazion:
I'll do this for consistency with how artist tags in tag lists link to the artist page, but I'd rather have wiki pages for artist tags just show all the artist information themselves.

There's also a corner case here with non-artist tags that have artist entries.

Issue #3955, opened 2018-10-11
Reduce usage of cookies


BrokenEagle:
Discussed this issue on Discord with @evazion. This was prompted by the recent following error message.

![image](https://user-images.githubusercontent.com/21149935/46773085-5fe5b480-ccb0-11e8-9d6a-d5ba647aaeed.png)

I looked and there were a lot of Chrome created cookies.

![image](https://user-images.githubusercontent.com/21149935/46773101-78ee6580-ccb0-11e8-8805-8ac25fbf6447.png)

However, even without those, I checked and the cookie size was already > 2KB, half of the maximum 4KB size.  There's not much that can be helped about other programs setting cookies, however there are things Danbooru can use to minimize cookies, especially since many cookies don't need to be cookies.

Some of the current cookies could go to local storage, some of them could be ``<meta>``'s, and some need to stay as cookies.

### Local Storage
- dab
- hide_dmail_notice
- mode
- news-ticker
- current_tag_script_id
- tag-script-1-9

### Meta
- favorite_tags
- favorite_tags_with_categories
- recent_tags
- recent_tags_with_categories

### Cookie
- user_name
- password_hash
- _danbooru_session
- dmail_folder

evazion:
Things that control visibility of notices (hide_dmail_notice, news-ticker, etc) are better off as cookies, so that these things can be hidden in the html serverside. Otherwise if they were handled clientside we'd get the same problem as with blacklists, where page elements flicker because they start off as visible then get hidden later when javascript runs.

`user_name` and `password_hash` are set by the "Remember me" checkbox when logging in, so that you don't have to login again when your session expires. But that can be better achieved by just making the `_danbooru_session` cookie never expire, so these cookies can be eliminated.

`favorite_tags` and `recent_tags` are only used to pass down information to related_tags.js to generate the html for the Related Tags section. This could be avoided completely by building the related tags html on the server instead of in javascript. This would also simplify the related tags javascript overall.

BrokenEagle:
For ``hide_dmail_notice`` and ``news-ticker``, I was just going by the fact that they're currently being handled by Javascript.  However, I agree that if they can avoid being rendered at all server side, then that would be much better.

BrokenEagle:
I'll take a crack at handling the tag-scripts and mode.

evazion:
Most of these cookies have been eliminated.

PR #4251, opened 2020-01-08
Change tag script storage

BrokenEagle:
Addresses the script storage from #3955. Additionally, there was an issue where the numeric keypad was setting the script ID as "a-z`", which was fixed by rebasing the value off of 0.


evazion:
It should be mentioned that using local storage means settings won't be shared across domains, i.e. between safebooru.donmai.us and danbooru.donmai.us. This shouldn't be a problem, but it's worth noting.

BrokenEagle:
Well, since those were such heavy cookies, I believe in this case that the benefits outweigh the costs.

PR #4253, opened 2020-01-09
Fix missing post HTML attributes

BrokenEagle:
I noticed the other day that the post model only had `uploader_id` and `approver_id`. That's because the post model was missed when all of the other models were converted over. I did a directory search and didn't find any other ones that didn't get done.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -1481,10 +1481,6 @@ class Post < ApplicationRecord
       hash
     end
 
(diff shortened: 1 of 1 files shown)


evazion:
This was semi-intentional since there are a lot of attributes on posts and most of them either aren't very useful here, or they're internal implementation details that outside users really shouldn't rely upon (ex: the tag_count_* fields).

Issue #4047, opened 2019-01-19
Check for Twitter common hashtags in translated tags


BrokenEagle:
A very common hashtag for Twitter is the birthday hashtags, which I've seen used for characters and copyrights.

https://danbooru.donmai.us/artist_commentaries?search%5Btext_matches%5D=%2A%E7%94%9F%E8%AA%95%E7%A5%AD%2A

It takes the form of:

**name** + 生誕祭  + ***optional year***

This could be done similar to how the Pixiv "X number of bookmarks" tags.

https://github.com/danbooru/danbooru/blob/13dff046f7dc3d17ea29dee699ba5bf45af916e0/app/logical/sources/strategies/pixiv.rb#L220-L222

The Twitter version could be:

```ruby
      def normalize_tag(tag)
        tag.gsub(/生誕祭(?:\d+)?\z/, "")
      end
```

Besides the above though, there is also the one hour draw tags.  The following are the unique variances of that tag I was able to find on Danbooru.

https://danbooru.donmai.us/artist_commentaries?search%5Btext_matches%5D=%2Aお絵描き60分一本勝負%2A

> 版もうひとつの深夜の真剣お絵描き60分一本勝負
> 版深夜の真剣お絵描き60分一本勝負
> 深夜の真剣お絵描き60分一本勝負
> 版深夜のお絵描き60分一本勝負
> 版真剣お絵描き60分一本勝
> 版お絵描き60分一本勝負

So instead, we could loop through a regex array doing the replacement if possible each time.

```ruby
COMMON_TAG_REGEXES = [
    /生誕祭(?:\d+)?\z/,
    /版もうひとつの深夜の真剣お絵描き60分一本勝負(?:_\d+)?\z/,
    /版深夜の真剣お絵描き60分一本勝負(?:_\d+)?\z/,
    /深夜の真剣お絵描き60分一本勝負(?:_\d+)?\z/,
    /版深夜のお絵描き60分一本勝負(?:_\d+)?\z/,
    /版真剣お絵描き60分一本勝(?:_\d+)?\z/,
    /版お絵描き60分一本勝負(?:_\d+)?\z/
]

def normalize_tag(tag)
    COMMON_TAG_REGEXES.each { |rg| tag = tag.gsub(rg,"") }
    tag
end
```

I also noted that sometimes the hashtag is just the common tag format by itself, so either the **normalize_tags** function will need to reject empty strings, or the regexes need to be modified to add a negative lookbehind assertion.

[...]

PR #4255, opened 2020-01-10
Add additional Twitter hashtag

BrokenEagle:
It was mentioned on issue #4047 but was forgotten to be included


BrokenEagle:
On responding to evazion/translate-pixiv-tags#43, I realized that doing the regex operation only once would be more efficient.

PR #4256, opened 2020-01-13
Update HTML table column attributes

BrokenEagle:
This primarily adds a column parameter for use on columns which have no header name. With this, the **th** and **td** parameters don't have to be used to set the column class. This also prevents extremely long class names from being created, e.g. on the post versions view: (`label-input-type-checkbox-id-post-version-select-all-checkbox-class-post-version-select-checkbox-label-column`). It then adds missing column classes for all applicable tables.

Next , it standardizes the use of the width parameter for the column parameters to not use surrounding curly braces.

Finally, it fixes up the note versions index to remove the unneeded empty column and fix the search by post ID and note ID links so that they appear whenever not searching by that parameter, which includes the vanilla index endpoint, where I had need for them myself but they weren't there.

### DIFF
--- a/app/logical/table_builder.rb
+++ b/app/logical/table_builder.rb
@@ -2,8 +2,9 @@ class TableBuilder
   class Column
     attr_reader :attribute, :name, :block, :header_attributes, :body_attributes
 
-    def initialize(attribute = nil, th: {}, td: {}, width: nil, name: nil, &block)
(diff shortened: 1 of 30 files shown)


evazion:
Please use a blank line to separate the title from the body in commit messages. Otherwise in `git log` the body is treated as part of the title and the commit message becomes extremely long:

image

https://chris.beams.io/posts/git-commit/#separate

PR #4259, opened 2020-01-14
Add upload commentary enhancements

### LINKED ISSUE #2794: Add commentary_request to upload tagbox if commentary is included
Under the assumption that original commentary usually requires translation:

On the upload page, when the "Include Commentary" box checked, automatically add `commentary_request` to the tag box.  If it is later unchecked, remove `commentary` and `commentary_request` from the tag box.

BrokenEagle:
Fixes #2794. In addition to the request from that issue, several other enhancements were made to the upload interface for artist commentaries.

1. Can now translate commentary from the upload page

    Sometimes the translator is the one uploading. This facilitates adding the translation in one step instead of having the multiple step process of submitting the upload, waiting for the post to be created with a page reload, then adding the commentary with another page reload.

    The translated section starts out hidden by default, and the control to show the translated section is only shown if the commentary section itself is shown.

2. Can now add commentary tags with a checkbox

    The checkbox names are the same as those on the add commentary dialogue. The tags involved with just the original section of commentary are included in that section, namely **commentary** and **commentary_request**. The tags involved with the translated section of the commentary are included in that section, namely **commentary_check** and **partial_commentary**.

### Notes

Since the other commentary fields were being saved in the upload record, the new input fields were added to the upload record as well. Also, those upload fields were used instead of the **params** parameter in several places since the upload record already has those fields and it doesn't require converting to a string and then a boolean.

### DIFF
--- a/db/migrate/20200114204550_add_more_commentary_to_uploads.rb
+++ b/db/migrate/20200114204550_add_more_commentary_to_uploads.rb
@@ -1,0 +1,10 @@
+class AddMoreCommentaryToUploads < ActiveRecord::Migration[6.0]
+  def change
+    add_column :uploads, :translated_commentary_title, :text
+    add_column :uploads, :translated_commentary_desc, :text
+    add_column :uploads, :add_commentary_tag, :boolean
+    add_column :uploads, :add_commentary_request_tag, :boolean
+    add_column :uploads, :add_commentary_check_tag, :boolean
+    add_column :uploads, :add_partial_commentary_tag, :boolean
+  end
+end
--- a/db/migrate/20200114204550_add_more_commentary_to_uploads.rb
+++ b/db/migrate/20200114204550_add_more_commentary_to_uploads.rb
@@ -1,10 +1,6 @@
 class AddMoreCommentaryToUploads < ActiveRecord::Migration[6.0]
   def change
-    add_column :uploads, :translated_commentary_title, :text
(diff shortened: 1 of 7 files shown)


BrokenEagle:
Did a small fixup removing a debug statement that was left in.

db/migrate/20200114204550_add_more_commentary_to_uploads.rb
```diff
@@ -0,0 +1,10 @@
+class AddMoreCommentaryToUploads < ActiveRecord::Migration[6.0]
+  def change
+    add_column :uploads, :translated_commentary_title, :text
+    add_column :uploads, :translated_commentary_desc, :text
+    add_column :uploads, :add_commentary_tag, :boolean
+    add_column :uploads, :add_commentary_request_tag, :boolean
+    add_column :uploads, :add_commentary_check_tag, :boolean
+    add_column :uploads, :add_partial_commentary_tag, :boolean
```
evazion: These should all have `null: false` and defaults set (`default: false` for booleans and `default: ""` for strings`).

evazion:
I don't think the add commentary checkboxes are necessary here. They take up a lot of extra space, which I don't think uploaders will be happy about. I think related tags are better suited for this. Commentary tags should show up in your recent tags already if you use them a lot, otherwise you can put them in your favorite tags. This way you can include other tags like `english_commentary` that you may also need too.

PR #4267, opened 2020-01-18
Create the ability to send reports to moderators

### LINKED ISSUE #2837: Ability to "report" forum posts, comment threads and users
When something goes haywire in comment threads or forums and mod attention is needed, users currently have two choices - either create/find a forum topic to mention that something goes wrong, or dmail a few mods. First method is good for mod staff, but can make user a target for retaliation as a whistleblower, and second method causes quite a lot of annoyance from people who are dmailed if that happens more than a few times, not to mention they might not actually see it until much later.

What I propose is a way for users to "flag" or "report" inappropriate comment thread, forum post, forum thread, or user, which will simultaneously notify all of the mod staff without revealing the reporter to offenders. It should offer a possibility for user to leave a comment describing what's wrong.

I initially thought about a separate table and page where such reports would be recorded. An unresolved record would create a notice similar to "check the queue" for mod+ staff. The first mod to deal with the report would mark it as "resolved", so that others are not bothered with it if they login later. A report should also have an option to "hide" similar to queue if the mod can't solve it and don't want to be bothered. Visiting the report page should remove the notice for a certain period, unless new reports are added.

However, after thinking about it more, there's a much simpler method: create a mod+ topic, maybe make it sticky, record its ID in the config, and have "report" button create a forum post from reporting user in that topic. This way, every moderator who visits the forums would see it and react if necessary. On a plus side, it's much, much simpler to code, uses the existing facilities, and automatically gives mods means of discussing the reports. On the negative side, mods still should visit the forum to see it, and resolved case would still alert everyone.

BrokenEagle:
Fixes #2837. In addition to what was suggested there, it also creates records which facilitates having an index table plus being able to highlight the instances where they are for moderators.

One thing not added yet is highlighting per instance, as that needs to be done at the forum topic level for forum posts, and the post level for comments, to avoid sending too many SQL requests.

### DIFF
--- a/app/controllers/moderation_reports_controller.rb
+++ b/app/controllers/moderation_reports_controller.rb
@@ -17,7 +17,6 @@ class ModerationReportsController < ApplicationController
   def create
     check_privilege
     @moderation_report = ModerationReport.create(moderation_report_params)
-    @moderation_report.create_forum_post!
     respond_with(@moderation_report)
   end
 
--- a/app/controllers/moderation_reports_controller.rb
+++ b/app/controllers/moderation_reports_controller.rb
@@ -1,0 +1,49 @@
+class ModerationReportsController < ApplicationController
+  respond_to :html, :xml, :json, :js
+  before_action :builder_only, only: [:new, :create]
+  before_action :moderator_only, only: [:index]
+
+  def new
+    check_privilege
+    @moderation_report = ModerationReport.new(moderation_report_params)
+    respond_with(@moderation_report)
+  end
+
+  def index
+    @moderation_reports = ModerationReport.paginated_search(params).includes(:creator, :model)
+    respond_with(@moderation_reports)
+  end
+
+  def create
+    check_privilege
+    @moderation_report = ModerationReport.create(moderation_report_params)
+    @moderation_report.create_forum_post!
+    respond_with(@moderation_report)
+  end
+
+  private
+
+  def model_type
+    params.fetch(:moderation_report, {}).fetch(:model_type)
+  end
+
+  def model_id
+    params.fetch(:moderation_report, {}).fetch(:model_id)
+  end
+
+  def check_privilege
+    case model_type
+    when "User"
+      return if User.find(model_id).reportable_by?(CurrentUser.user)
+    when "Comment"
+      return if Comment.find(model_id).reportable_by?(CurrentUser.user)
+    when "ForumPost"
+      return if ForumPost.find(model_id).reportable_by?(CurrentUser.user)
+    end
+    raise User::PrivilegeError
+  end
+
+  def moderation_report_params
+    params.fetch(:moderation_report, {}).permit(%i[model_type model_id reason])
+  end
+end
--- a/app/javascript/src/styles/specific/comments.scss
+++ b/app/javascript/src/styles/specific/comments.scss
@@ -1,6 +1,12 @@
 @import "../base/000_vars.scss";
 
 div.comments-for-post {
+  div.moderation-comments-notice {
+    margin: 1em 0;
+    font-weight: bold;
+    color: red;
+  }
+
   div.hidden-comments-notice {
     margin: 1em 0;
   }
@@ -11,6 +17,10 @@ div.comments-for-post {
 
       &[data-is-sticky="true"] {
         background: var(--comment-sticky-background-color);
+      }
+
+      &[data-is-reported="true"] {
+        border: solid red;
       }
 
       &[data-is-voted="true"] {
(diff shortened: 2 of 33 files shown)


BrokenEagle:
Added highlighting per instance. For topics and posts, this includes loading all moderation reports in a single query and then passing those to the render functions.

The next thing to be added at this point is a pruning task which removes reports over a week old. This will just remove the records though, and will not remove the forum posts that are created as well.

BrokenEagle:
Added the pruning of the reports and the tests. A fixup was done and force pushed because one of the tests required an HTML new endpoint. This should be alright though since it probably needed one anyways, plus it better matches with post flags which the moderation reports were partly based on.

At this time everything should be completed, so I'm going to mark it ready for review.

BrokenEagle:
Did a small fixup after doing **rubocop** on the new test files.

BrokenEagle:
I pushed a couple of fix commits to avoid moving all of the other commits beneath my messages, but I can organize, rebase, and squash them if needed.

app/controllers/moderation_reports_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def index
+    @moderation_reports = ModerationReport.paginated_search(params).includes(:creator, :model)
+    respond_with(@moderation_reports)
+  end
+
+  def create
+    check_privilege
+    @moderation_report = ModerationReport.create(moderation_report_params)
+    @moderation_report.create_forum_post!
```
evazion: It's possible for `ModerationReport.create` to fail if the report fails validation. If this happens, `create_forum_post!` will create the forum post even though the moderation report didn't get created.

This can be fixed by calling `create_forum_post!` in an `after_create` hook on the model instead. This way if creating the report fails, then the after_create hook won't run.

app/javascript/src/styles/specific/comments.scss
```diff
@@ -1,6 +1,12 @@
 @import "../base/000_vars.scss";
 
 div.comments-for-post {
+  div.moderation-comments-notice {
+    margin: 1em 0;
+    font-weight: bold;
+    color: red;
```
evazion: CSS colors should use variables defined in `app/javascripts/src/styles/base/040_colors.css` instead of being hardcoded. This is so that light and dark mode can use the right colors.

app/models/moderation_report.rb
```diff
@@ -54,4 +54,8 @@ def self.search(params)
 
     q.apply_default_order(params)
   end
+
+  def self.prune!
+    where("created_at < ?", 1.week.ago).delete_all
+  end
```
evazion: Reports don't need to be pruned. I prefer to keep data unless there's a strong need to delete it.

app/views/comments/partials/index/_list.html.erb
```diff
@@ -3,10 +3,10 @@
     <%= render "comments/partials/index/header", :post => post %>
   <% end %>
 
-  <% if CurrentUser.is_moderator? && post.moderation_reports.any? %>
+  <% if CurrentUser.is_moderator? && post.moderation_reports.present? %>
     <div class="row moderation-comments-notice">
       <span class="info" id="moderation-comments-notice-for-<%= post.id %>">
-        This post has comments reported for moderation! (<%= post.moderation_reports.count %> <%= (post.moderation_reports.count == 1 ? "report" : "reports") %>)
+        This post has comments reported for moderation! (<%= post.moderation_reports.length %> <%= (post.moderation_reports.length == 1 ? "report" : "reports") %>)
```
evazion: Here you can use the Rails `pluralize` helper to pluralize the string: 

`pluralize(post.moderation_reports.length, "report")`.

app/models/user.rb
```diff
@@ -812,4 +812,8 @@ def initialize_attributes
   def presenter
     @presenter ||= UserPresenter.new(self)
   end
+
+  def viewable_moderation_reports
+    !is_moderator? && CurrentUser.is_moderator? ? moderation_reports : []
+  end
```
evazion: Is this meant to say that mods can't view reports generated against other mods? Because they'll still be visible in the forum thread. Reports against mod comments would also be visible.

Allowing reports against mods to be visible to other mods is fine for now. Otherwise we'd have to get rid of the forum thread to effectively hide these.

app/models/moderation_report.rb
```diff
@@ -0,0 +1,57 @@
+class ModerationReport < ApplicationRecord
+  belongs_to :model, polymorphic: true
+  belongs_to_creator
```
evazion: The model should have a `validates :reason, presence: true` validation to ensure that the reason can't be blank.

BrokenEagle:
I made all of the recommended changes in 47a3022.

BrokenEagle:
Did a quick fixup removing the `prune!` function for the moderation reports model (it was already removed from the daily tasks.

PR #4270, opened 2020-01-21
Add universal index redirect

### LINKED ISSUE #4204: Have single result searches go to the show pages for certain controllers
This would be to match the current behavior of the wiki pages controller, plus it would shorten what is currently a two step-process of search, then click the single result.

This should be done for:

1. Artists
    - Example: https://danbooru.donmai.us/artists?search%5Bany_name_matches%5D=maco_spl
2. Pools
    - Example: https://danbooru.donmai.us/pools/gallery?search%5Bname_matches%5D=Disgustingly_Adorable
3. Forum topics
    - Example: https://danbooru.donmai.us/forum_topics?search%5Btitle_ilike%5D=%5BBUR%5D+Girls+und+Panzer
4. Users
    - Example: https://danbooru.donmai.us/users?search%5Bname_matches%5D=albert

I'm not sure if this behavior on any of the other controllers would be useful.

Also, this could also be a potential search term addon for those that don't want that behavior, i.e. `search[display_single]=true` would bring up the show page for a single match whereas the false condition would not. I'm not sure which way should be the default though. However, to do that on the application controller would require overrides on all of the controllers without show pages, so it might be easier to this on the individual controllers instead if this actually gets implemented.

BrokenEagle:
Fixes #4204. This is a more comprehensive version of 6e5e070. 
- It allows users to redirect to the show page on all controllers that have index and show actions.
- It also fixes the quick search forms so that the redirect parameter is accurate to the intended purpose of the search
    - I.e. whether it will redirect or not if only one match is found.
- Only the search parameters for the quick search forms are being handled specially in the controller.
    - The use of any other search parameters will redirect as long as only one result is returned when the redirect parameter is true.
    - This should be fine as long as users are aware of that.
- All of the controllers with index actions had their render functions replaced even if they didn't have a show action
    - This is just in case a show action is ever added to that controller in the future.
- It also fixes a bug in the wiki page autocomplete caused by 3436827

I already have a use for this functionality myself, as I can add it to my IndexedAutocomplete script for the forum topic search, and perhaps add it to several other autocomplete fields. Besides that, I also have several other uses in mind.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -22,6 +22,28 @@ class ApplicationController < ActionController::Base
     end
   end
 
+  private
+
+  def respond_with(*options, &block)
+    if params[:action] == "index" && is_redirect?(options[0])
+      redirect_to_show(options[0])
+    else
+      super(*options, &block)
+    end
+  end
+
+  def redirect_to_show(items)
+    redirect_to send("#{controller_path.singularize}_path", items.first, format: request.format.symbol)
+  end
+
+  def is_redirect?(items)
+    action_methods.include?("show") && params[:redirect].to_s.truthy? && items.one? && item_matches_params(items.first)
+  end
+
+  def item_matches_params(*)
+    true
+  end
+
   protected
 
   def enable_cors
--- a/app/controllers/artists_controller.rb
+++ b/app/controllers/artists_controller.rb
@@ -35,11 +35,7 @@ class ArtistsController < ApplicationController
     @artists = @artists.includes(:tag) if request.format.html?
     @artists = @artists.includes(:urls) if !request.format.html?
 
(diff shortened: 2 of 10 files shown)


BrokenEagle:
Did a fixup removing the wiki page autocomplete fix as it was already handled by 0a3d038.

evazion:
Instead of having `render_index` and `default_render_index`, you should just override `respond_with`. This would avoid having to change `respond_with` to `rendex_index` everywhere. You can use `params[:action] == "index"` to check whether you're in an index action and `super` to call the real `respond_with`.

BrokenEagle:
Sorry for all of the force pushes, but there were several corrections that I didn't notice as I was updating this.

Anyways, it should be good to merge now.

PR #4273, opened 2020-01-26
Add parameters for searching attributes by array with case insensitivity

BrokenEagle:
It does the by performing a lowercase on the text, so the user of these params must also do a lowercase on their end before sending the query.

### Edit:

This has the immediate use case of searching batches of wiki page other names, as is a need with [Translate Pixiv Tags](https://github.com/evazion/translate-pixiv-tags).


BrokenEagle:
Added more parameters for searching text attributes by array with case insensitivity. Although many text fields are normalized to lowercase already, there are several that aren't, and this allows for using the array option instead of the singular option to find them.

Issue #3918, opened 2018-09-22
Various exceptions for some artist URLs from different sites


BrokenEagle:
Got the following when trying to edit an artist ([artist #72332](https://danbooru.donmai.us/artists/72332)):
```
OAuth2::Error exception raised
app/logical/pawoo_api_client.rb:108:in `get'
app/logical/sources/strategies/pawoo.rb:94:in `block in api_response'
app/logical/sources/strategies/pawoo.rb:93:in `each'
app/logical/sources/strategies/pawoo.rb:93:in `api_response'
app/logical/sources/strategies/pawoo.rb:53:in `profile_url'
app/logical/sources/strategies/pawoo.rb:77:in `normalize_for_artist_finder'
app/models/artist_url.rb:39:in `normalize'
app/models/artist_url.rb:118:in `normalize'
app/controllers/artists_controller.rb:64:in `update'
```

BrokenEagle:
Found another exception for [artist #71472](https://danbooru.donmai.us/artists/71472):

```
ArgumentError exception raised
absolute URL needed (not nil)
app/logical/sources/strategies/nijie.rb:114:in `page'
app/logical/sources/strategies/nijie.rb:56:in `profile_url'
app/logical/sources/strategies/base.rb:147:in `normalize_for_artist_finder'
app/logical/sources/strategies/base.rb:141:in `normalizable_for_artist_finder?'
app/models/artist_url.rb:38:in `normalize'
app/models/artist_url.rb:118:in `normalize'
app/models/artist.rb:186:in `block (2 levels) in save_urls'
app/models/artist.rb:185:in `map'
app/models/artist.rb:185:in `block in save_urls'
app/models/artist.rb:183:in `save_urls'
app/controllers/artists_controller.rb:64:in `update'
```

The exception occurred when trying to send the following list of urls:

> http://seiga.nicovideo.jp/manga/list?user_id=12609685
> http://nijie.info/members.php?id
> https://twitter.com/intent/user?user_id=165840275 
> http://www.nicovideo.jp/user/12609685
> http://seiga.nicovideo.jp/user/illust/12609685
> https://twitter.com/Cyorisukenosuke
> https://www.pixiv.net/stacc/tyotyotyori
> https://www.pixiv.net/member.php?id=910858
> -http://twitpic.com/photos/cyorisukenosuke

[...]

PR #4277, opened 2020-01-29
Fix Pawoo source

### LINKED ISSUE #4276: Pawoo source requests are broken
The following error occurs whenever uploading or editing a post with a Pawoo source.
```
Error
undefined method `to_h' for #<PawooApiClient::Status:0x000055f0e0f664a0> Did you mean? to_s

Details
NoMethodError exception raised
app/logical/sources/strategies/base.rb:286:in `to_h'
app/controllers/sources_controller.rb:7:in `show'
```
It occurs when trying to convert the API response to a hash.

https://github.com/danbooru/danbooru/blob/d29b88d43b87d69112dcf67582d5d391970bc110/app/logical/sources/strategies/base.rb#L286

This is because the Status object created does not have a **to_h** function.

BrokenEagle:
Fixes #4276.

### DIFF
--- a/app/logical/pawoo_api_client.rb
+++ b/app/logical/pawoo_api_client.rb
@@ -109,11 +109,26 @@ class PawooApiClient
 
   def get(url)
     if id = Status.is_match?(url)
-      return Status.new(JSON.parse(access_token.get("/api/v1/statuses/#{id}").body))
+      begin
+        data = JSON.parse(access_token.get("/api/v1/statuses/#{id}").body)
+      rescue
+        data = {
+          "account" => {},
+          "media_attachments" => [],
+          "tags" => [],
+          "content" => "",
+        }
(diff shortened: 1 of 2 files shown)


BrokenEagle:
I added in a few other fixes for Pawoo that were needed, such as what was mentioned in #3918. I also made it so that normalization of the account ID links will now work when the link is bad, as currently trying to save a bad account ID link in an artist's record will cause an exception. Up until now I've been having to mangle the web URL to "bad-pawoo.net" in order to get it to work.

https://danbooru.donmai.us/artist_urls?search%5Border%5D=id&search%5Burl_matches%5D=%2Abad-pawoo%2A

Speaking of #3918, I just checked, and this PR will fix the remaining issues from there, so it will be able to be closed as well.

PR #4281, opened 2020-02-02
Add embedded notes improvements

### LINKED ISSUE #4235: Embedded notes improvements
There have been several discussions on Discord about how to improve embedded notes. That way the translator could decide how they want each item to look on the post

1. No resize icon unless hovering

    This can often get in the way of the text.

2. No borders

    Borders interfere with having a seamless integration of the note on the post. This can be currently hacked with CSS, although it would be preferable if this was just the default for embedded notes.

3. Full opacity

    Same as above.

4. Settable background color

    Same as above.

5. Alternate shapes

    Such as circles, ovals, trapezoids, and  parallelograms.
    https://www.w3schools.com/howto/howto_css_shapes.asp

6. Custom shapes

   I.e. drawn by the user. This one might be the hardest to do, so it would be more of a stretch goal.

1 above should always be on for embedded notes. 2 and 3 could be a note setting instead of always on, 4-6 would have to be a note setting.

BrokenEagle:
Fixes #4235. In addition to the items there, it also adds the much needed key functions for nudge and resize. It also adds a few settable styles beyond background color on the note box itself, which facilitates having different shapes and orientations of the box. It does this by copying certain styles of the first element with the class of **note-box-attributes** that it finds in the note text if it exists.

Several style manipulations were converted to using classes instead of setting styles on the elements themselves, which allows for easier maintenance and setting of multiple styles at once. For the note preview function, the font-size of the parent inner box is set to the base font size (set at page load) so that the size calculations can be done correctly, and afterwards it clears that font size so that the note elements get scaled correctly according to the current font size of the note container. Finally, some of the dead embedded code was removed.

### DIFF
--- a/app/javascript/src/javascripts/notes.js
+++ b/app/javascript/src/javascripts/notes.js
@@ -95,15 +95,16 @@ let Note = {
           var $this = $(this);
           var $note_box_inner = $(e.currentTarget);
 
+          const note_id = $note_box_inner.data("id");
           if (e.type === "mouseover") {
-            Note.Body.show($note_box_inner.data("id"));
+            Note.Body.show(note_id);
             if (Note.editing) {
               $this.resizable("enable");
               $this.draggable("enable");
             }
             $note_box.addClass("hovering");
           } else if (e.type === "mouseout") {
-            Note.Body.hide($note_box_inner.data("id"));
+            Note.Body.hide(note_id);
             if (Note.editing) {
               $this.resizable("disable");
               $this.draggable("disable");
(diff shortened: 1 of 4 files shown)


evazion:
* Keyboard nudging should only be enabled when you're dragging the note, not when you're hovering over it. Hovering has too many problems:
** It's too easy to accidentally move notes when using arrow keys to scroll. If you're using up/down to scroll the page, and your mouse enters a note, then suddenly you're moving the note instead of scrolling the page.
** It only works as long as you keep the mouse inside the note. You have to constantly reposition the mouse to keep it inside the note, otherwise you stop moving.
** The popup for the note body is still visible while nudging the note. Normally it's hidden while dragging the note. This is a problem because the note body doesn't move with the note box, so you can't move down without the note body covering up the note box.

* Keyboard resizing has the same issues. You have to keep the mouse inside the resize handle at all times to keep resizing. Instead keyboard resizing should only be enabled after you've used the mouse to drag the handle and enter resize mode. Alternatively: use shift+arrow to resize when dragging the note.

* It's possible to use the keyboard to move or resize notes outside the image boundaries.

* Font resizing can produce notes that are too small to read on mobile. See [post #2393657](https://danbooru.donmai.us/posts/2393657) and [post #3438601](https://danbooru.donmai.us/posts/3438601) for examples. To a certain extent this is unavoidable, since the text really does have to be that small to fit in the note box, but still, it basically means that translators always have to increase the font size for embedded notes (and can never decrease the size), otherwise notes will be unreadable on mobile.

* It's possible to create invisible notes with `background-color: rgba(0, 0, 0, 1%)`. Colors that aren't 100% opaque shouldn't be allowed.

* It's possible to create notes that fill up the entire screen with `transform: scale(100)`. Only rotation should be allowed, otherwise you can break out of the image boundaries.

* Rotations break boundary detection. When dragging the note, the unrotated note is still used for boundary detection. This lets you create notes that break out of the image. It also means you can't place notes against the edges in some cases (trying making a tall skinny note, rotating it 45 degrees, then putting it against the top edge).

evazion:
The main issue here is the hover-to-nudge behavior. It's too easy to trigger accidentally when using the arrow keys normally. It should only be active when you're already dragging or resizing a note, or when you click the note to select it or something.

BrokenEagle:
> The main issue here is the hover-to-nudge behavior. It's too easy to trigger accidentally when using the arrow keys normally. It should only be active when you're already dragging or resizing a note, or when you click the note to select it or something.

It'll have to be clicking the note, as dragging it locks the note in position to where the mouse cursor as at.  You can nudge the note around, but as soon as you stop, the note shoots back into place. I'll have the click action select a note and indicate that with a different colored border. Clicking it again will unselect it, or selecting a different note will also unselect it.

evazion:
I see, that's fine. Click-to-select would also allow other keyboard shortcuts in the future. Think: being able to click a note box then press 'e' to edit it or 'del' to delete it. Or using tab or shift+tab to quickly move between notes. Or imagine being able to select multiple notes and delete them all at once, or move them together at the same time.

evazion:
For now I've merged 858fd01, e2874bc, f3bd304. You should be able to rebase to remove those commits and resolve the conflicts.

BrokenEagle:
I did a force push since I had to rebase everything.

Anyways, I made it so that the note box has to be clicked before the nudge or resize options become available. This status is indicated with a green border. Clicking the box again or another box will turn off that option for the box. For the resize option, it is now done with the shift key in addition to the arrow keys.

Several checks for invalid conditions were added in as well. Border detection was added in for both the nudge and resize functions. For backgrounds, I added a check for transparency. For transformation, only rotations are allowed.

For rotations especially, since they can break borders depending on the rotation, I added additional border detection. So now only values which keep the boxes inside the image and whose coordinates are also in the image are acceptable. This means that there are certain portions of an image where a rotated box cannot be moved to, even if it looks like it could, since it would require coordinates which are outside the image and are not allowed.  Also, if a user attempts to rotate a box that is so large that it will break through opposing sides of an image, then the rotation is rejected completely and an error message is sent to the user. There is also detection for when a user adds bad position values through the API or the browser dev console. For these, the box is repositioned upon page load and is given a red border with the `out-of-bounds` class.

BrokenEagle:
Posted some additional fixes.

One is directly related to notes, in that the dialogue text was wrong for new notes beyond the first note, as it was showing "Editing note #xx..." and so on instead of "Creating new note".

Another is partly related, involving the resize image to window link. For that it was using the original width and height for restoring the size to full, even if it was showing the 850px version of the image. This was causing excessive blur and artifacts in some cases.

Issue #2704, opened 2016-10-05
Various Controllers Update Vulnerability


BrokenEagle:
In trying to determine all of the variables usable as inputs to the Update function by all of the controllers, I learned of some variables which if modified in this backdoor fashion could cause problems.

Listed according to severity of the vulnerabity:

A. Tag aliases/Tag implications:
- "status" is settable to "active" through the update function

This does not update all of the tags as it normally would through the regular method, however any future tagging will be affected by that tag alias/impication

A crafty user could use the above vulnerability to create a tag alias/implication, quickly change the forum topic and post to something benign looking, then set the forum topic ID to NULL and the status to active for that particular alias.

I tried to search the existing active aliases/implications to see if this had been done in the past, but quickly learned that not all regular approved aliases/implications have an Admin ID as either the creator ID or approver ID attached to them, even as recently as [June this year](http://danbooru.donmai.us/tag_aliases/14099). The only way to verify would be to check the forum topics of all aliases/implications not created by an Admin to determine if an Admin responded in some way indicating that they had approved that particular alias/implication, and even then I'm not sure is a 100% sure-fire verification method. So there may be bogus aliases/implication with no way to programmatically detect...

B. Notes
- "post_id" is settable through the update function
  - This can leave orphaned note_versions
  - It can also leave semi-orphaned notes
    - A note_version does get created when post_id is changed
    - However that's only for the first such move... any other move within the hour-merge time-limit do not get recorded in the last_note_version.
  - Or completely orphaned notes
    - Related to the Posts vulnerability below
    - If the last_noted_at is set to null, that orphaned note completely disappears from view

C. Posts
- "last_noted_at" is settable
  - Can potentially affect notes as shown above

D. Comments
- "do_not_bump", "is_deleted", "post_id" are settable

post_id is the most serious, however it doesn't necessarily cause the same issues as the Notes vulnerability since comments are a singular instance with no versions.

is_deleted isn't that bad, especially since it might be desired to undelete a comment.

changing do_not_bump doesn't really seem to affect anything.  It's supposed to affect whether the "last_comment_bumped_at" gets [updated upon comment deletion](https://github.com/r888888888/danbooru/blob/master/app/models/comment.rb#L171).  However, neither that nor the "last_commented_at" got updated upon comment deletion, even when going through the Destroy function to set "is_deleted" instead of going through the Update function...?

evazion:
[...]

PR #4288, opened 2020-02-06
Add ability to use nested only parameter

BrokenEagle:
The only string works much the same as before with its comma separation, though nested attributes are now indicated with square brackets. The following is an example of how it can be used.

`http://danbooru.donmai.us/artists.json?only=name,other_names,tag[post_count],urls[normalized_url]&limit=1`
```json
[
  {
    "name": "0_05kmgrn",
    "other_names": [
      "壱"
    ],
    "tag": {
      "post_count": 0
    },
    "urls": [
      {
        "normalized_url": "http://twitter.com/intent/user?user_id=1137791083331514369/"
      },
      {
        "normalized_url": "http://twitter.com/0_05kmgrn/"
      }
    ]
  }
]
```

As seen, the two nested terms are the values immediately preceding the square brackets, and the only parameters for those are the comma separated terms inside the square brackets.

No limit was placed on the depth of nesting, except for you can only include a model twice if and only if the model was the endpoint, after which no more nested terms are accepted. The reason for the double inclusion is an example from the artist URLs endpoint.

`http://danbooru.donmai.us/artist_urls.json?only=normalized_url,artist[name,other_names,tag[post_count],urls[normalized_url]]&limit=1`
```json
[
  {
    "normalized_url": "http://twitter.com/0_05kmgrn/",
    "artist": {
      "name": "0_05kmgrn",
      "other_names": [
        "壱"
      ],
      "tag": {
        "post_count": 0
      },
      "urls": [
        {
          "normalized_url": "http://twitter.com/intent/user?user_id=1137791083331514369/"
        },
        {
          "normalized_url": "http://twitter.com/0_05kmgrn/"
        }
      ]
    }
  }
]
```

In this case, the double inclusion allows a user to search for an artist URL, and get back all of the related artist URLs, while also specifying that it only wants the normalized URL and nothing else. This use case is an actual usage from the **Translate Pixiv Tags** userscript (https://github.com/evazion/translate-pixiv-tags).

The following is an example of a nested inclusion which exceeds the double inclusion, because the artist inclusion did not start out on the endpoint.

`http://danbooru.donmai.us/artist_urls.json?only=artist[urls[artist[id]]]&limit=1`
```json
[
  {
    "artist": {
      "urls": [
        {
          
        },
        {
          
        }
      ]
    }
  }
]
```

For forbidden items, these are the inclusions that should not be seen either because of their sensitivity (e.g. moderation reports), or because the number of associated items has the potential to be very large (e.g. forum posts). Comments for posts weren't included in this, as the number of comments for any one post wasn't thought to be high.

As a matter of adding this capability, all of the index inclusions were moved to the model. This also allows them to be used elsewhere if needed, and it also prevents inclusions from being added where they aren't needed, e.g. on specific formats.

Finally, this also fixes the issue for artist URLs where the only parameter was not working at all, due to the options generated by the base application record not being passed to the format block in the controller.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -34,6 +34,25 @@ class ApplicationController < ActionController::Base
     end
   end
 
+  def model_includes(params, model = nil)
+    if params[:only] && ["json", "xml"].include?(params[:format])
+      includes_array = ParameterBuilder.includes_parameters(params[:only], model_name)
+    elsif params[:action] == "index"
+      includes_array = default_includes(params)
+    else
+      includes_array = []
+    end
+    includes_array
+  end
+
+  def default_includes(*)
+    []
+  end
+
+  def model_name
+    controller_name.classify
+  end
+
   def redirect_to_show(items)
     redirect_to send("#{controller_path.singularize}_path", items.first, format: request.format.symbol)
   end
--- a/app/controllers/artist_commentaries_controller.rb
+++ b/app/controllers/artist_commentaries_controller.rb
@@ -3,7 +3,7 @@ class ArtistCommentariesController < ApplicationController
   before_action :member_only, only: [:create_or_update, :revert]
 
   def index
-    @commentaries = ArtistCommentary.paginated_search(params)
+    @commentaries = ArtistCommentary.paginated_search(params).includes(model_includes(params))
     respond_with(@commentaries)
   end
 
@@ -35,6 +35,14 @@ class ArtistCommentariesController < ApplicationController
   end
 
   private
+
+  def default_includes(params)
+    if ["json", "xml"].include?(params[:format])
+      []
+    else
+      [{post: [:uploader]}]
+    end
+  end
 
   def commentary_params
     params.fetch(:artist_commentary, {}).except(:post_id).permit(%i[
--- a/app/controllers/artist_commentary_versions_controller.rb
+++ b/app/controllers/artist_commentary_versions_controller.rb
@@ -2,7 +2,7 @@ class ArtistCommentaryVersionsController < ApplicationController
   respond_to :html, :xml, :json
 
   def index
-    @commentary_versions = ArtistCommentaryVersion.paginated_search(params)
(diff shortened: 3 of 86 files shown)


BrokenEagle:
Did a small fixup of some leftovers that were fixed but left uncommitted.

BrokenEagle:
I went back and revised the double allowance for inclusion checking, since there should be no need to get twice of the same single instance. Now it's only allowed on inclusions that have the `:has_many` clause for that particular model.

evazion:
Two things:

* I don't want to specify default includes in the model. Models shouldn't need to know about how they're being used in views. This is a controller-level concern. 

* I'd really prefer a whitelist over a blacklist. It's very difficult for me to verify that everything is blacklisted correctly here. We had many security problems in the past when we used blacklists to hide sensitive attributes (see #2704).
  * One exploit I see right away is that you can create a forum post vote for a post in a mod-only forum topic (which currently isn't disallowed), then use `/forum_post_votes.json?only=forum_post` to get the mod-only post.

evazion:
#4291 pulled in a conflict for this.

BrokenEagle:
Did a force push when making the required changes since there were no code review elements yet.

> I don't want to specify default includes in the model. Models shouldn't need to know about how they're being used in views. This is a controller-level concern.

The default includes were moved to the controllers into the private section. Having them separate is needed to prevent unneeded includes when coming through the API for instance. Additionally, this provides extra visibility into what is being included based upon the input type and parameters. Right now the includes only accounts for the index action, as on the show action it is a singular event so nothing is gained by doing includes with it.

However, even though the includes aren't being preloaded on the show action, the only parameter with nesting is still available to be used.  This is because the **serializable_hash** function was altered, which is what is always called when requesting JSON or XML.

> I'd really prefer a whitelist over a blacklist. It's very difficult for me to verify that everything is blacklisted correctly here. We had many security problems in the past when we used blacklists to hide sensitive attributes (see #2704).

The available includes were changed from a blacklist into a whitelist. All associations were included except for those that were sensitive, those that were too many, or those that have exploits, such as the forum post vote one that was mentioned.

Besides adding this capability, several other improvements were noticed during development and were fixed accordingly.

PR #4291, opened 2020-02-08
Rework version views

BrokenEagle:
Standardizes most of the version views, reworks some of the views, and fixes several issues. For this, it adds/modifies several common helper functions and stylesheets.

1. Note versions
    - Uses the diff builder to highlight changes to the note body
    - Verbalizes all of the changes made in the status column
    - Splits up the position and size elements
        - Changes to the elements directly references the previous elements
        - Column headers are altered to inform on how the data is displayed
    - Date and user columns are combined into one to save space
2. Artist commentary versions
   - Uses the diff builder to highlight changes to the various text fields
   - Fields are shown in their unrendered form to better highlight changes
   - Title and description are now separated and labeled for ease in viewing
    - Verbalizes all of the changes made in the changes column
    - Date and user columns are combined into one to save space
    - Shows a post preview when searching by post ID much like the post versions index
3. Pool versions
    - Use the "diff-body" class instead of styles specific to pool versions
    - Change the status column and field names to be more intuitive
    - Only render the diff link if the name or description changes
    - Remove trailing whitespace on post links to fix line-through artifacts
4. Wiki page versions
    - Use the "diff-body" class instead of styles specific to wiki page versions
    - Change the status column and field names to be more intuitive
        - Also add more status names
5. Artist versions
    - Added a status/changes column listing all of the changes
        - This makes it easier to see what changes were made
        - It's also more in line with the other version views now
6. Artist versions
    - Added a status/changes column listing all of the changes
        - This makes it easier to see what changes were made
        - It's also more in line with the other version views now
    - Group name was moved to the "other names" column
        - Since it is typically wider than the "name" column

For the thumbnails on post/commentary revert listings, a top margin was added to make it more in line with the body of the table instead of the top of it. Also, a thumbnail wasn't added to note versions in the hope that eventually partial images will be able to be used on a per note basis using a viewbox or some other method.

For all of the links that drill down into the instance IDs which are the reversion pages, it uses the same anchor mechanism used on post versions on all other versions to highlight which version was clicked to navigate to the page.

Besides the changes to the version views, it also fixes a few other issues. The first is was caused when one form of a carriage return "\r\n" was replaced with another form "\n", or vice versa. It was causing the paragraph marks to be put on separate lines, adding extra line breaks where there were none before.

The next issue was the previous method on all of the versions, as any repeat calls to previous would call the same SQL query multiple times without any caching. Now it stores the data on the model as an array of one if the previous version exists, and an empty array if it doesn't. This makes it so that the previous function behaves the exact same way as before, where an existing previous version would return a single record, and no existing previous version would return nil.


evazion:
> The next issue was the previous method on all of the versions, as any repeat calls to previous would call the same SQL query multiple times without any caching. 

Technically Rails does cache repeated calls to the same SQL query, but they're still shown in the logs. RailsPanel incorrectly shows them as uncached calls, but in the actual console logs they're shown as cached calls.

PR #4295, opened 2020-02-16
Show the same post results through HTML and the API

### LINKED ISSUE #4195: Deleted results differ between API and HTML
Brought up by @Nikey646 on Discord. Basically, results returned through the API differ from those returned in HTML. In the API, it correctly returns deleted results when the deleted posts filter is off. With HTML, it incorrectly doesn't show results even when the filter is off.

Compare:
- https://danbooru.donmai.us/posts?tags=jav
- https://danbooru.donmai.us/posts.json?tags=jav

BrokenEagle:
Fixes #4195.

- It also makes the hide deleted posts setting work as it is described in the settings menu.
    - "Remove deleted posts from search results"
    - I.e. deleted posts should only be removed when the deleted posts filter is on.
- Even with the deleted posts filter on, it can still be overridden with the **show_deleted** parameter.

### DIFF
--- a/app/presenters/post_presenter.rb
+++ b/app/presenters/post_presenter.rb
@@ -7,7 +7,7 @@ class PostPresenter < Presenter
       return "<em>none</em>".html_safe
     end
 
-    if !options[:show_deleted] && post.is_deleted? && options[:tags] !~ /status:(?:all|any|deleted|banned)/ && !options[:raw]
+    if CurrentUser.hide_deleted_posts && !options[:show_deleted] && post.is_deleted? && options[:tags] !~ /status:(?:all|any|deleted|banned)/ && !options[:raw]
       return ""
     end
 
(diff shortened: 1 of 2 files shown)


evazion:
This would mean that deleted posts are no longer filtered out of tag searches unless you have the hide deleted posts option turned on. This option is off by default, so for most people it would mean they're suddenly seeing deleted posts when they weren't before. That's not the direction I want to go. 

The original purpose of this option was to change _how_ deleted posts are filtered out, not whether they're filtered out. When the option is on, it means deleted posts are filtered out at the database level instead of the ruby level. Filtering on the database level is slightly slower (or at least it was when the option was first added), but it avoids showing blank pages when all the posts on a page are deleted.

This option is confusing and the choice of whether to filter in the database or in ruby is something that end users really shouldn't be concerned about, so I plan to remove it eventually. Instead I plan to have a "show deleted posts" option that you can enable if you want to see all posts by default.

PR #4307, opened 2020-02-22
Show large image size by default on mobile devices

### LINKED ISSUE #3493: Always default to the resized image on mobile mode
Brought up [on Danbooru](http://danbooru.donmai.us/forum_topics/14866).  Related to issue #3074.

> 
> Insanity Demon said:
> 
> [Feature request] Separate image width settings for mobile and desktop
> Or even just default to 850px on mobile regardless of the user setting, since I think I can safely say that most smartphones today simply would not have enough free memory to load a bunch of absurdres images anyway.

I think this would be a good idea.  It doesn't affect me since I have the default set to 850 px, but it may affect those that don't negatively like Insanity Demon mentioned.

BrokenEagle:
Fixes #3493.


BrokenEagle:
Per the conversation on Discord (05/04/2020, [link](https://discordapp.com/channels/310432830138089472/310846683376517121/707041882898825256)), I changed this PR to avoid using a cookie.

I modeled this similarly to how post previews are rendered, i.e. using a picture element with 2 source elements.

One item of note, is that on my development instance, when in the desktop view, the image had a tendency to load twice, with the first being either a partial or a full image download. This was also seen when switching between desktop and mobile modes. Additionally, after I enabling cropped thumbnails, the double loading of images was seen in the index view as well. This behavior was only seen on Chrome, and searching the web, I see that that was an issue in the past.

https://stackoverflow.com/questions/33919840/why-does-the-picture-tag-download-an-image-twice

However, replicating the use of the same element structure on the Danbooru site, the same issue was not seen. I also tested it on Testbooru, and it wasn't seen there either. So it seems like that the production site is doing something that the development instance isn't? Or perhaps it could be something else, like the fact that the development instance is on a VM or using a private host IP?

So I'd recommend testing this on Testbooru first, just to make sure that the correct behavior is still seen, and after confirming that, rolling it out to the main site.

app/views/posts/partials/show/_image.html.erb
```diff
@@ -1,3 +1,7 @@
 <% if policy(post).visible? %>
-  <%= image_tag(post.file_url_for(CurrentUser.user), width: post.image_width_for(CurrentUser.user), height: post.image_height_for(CurrentUser.user), id: "image", class: "fit-width", "data-original-width": post.image_width, "data-original-height": post.image_height, "data-large-width": post.large_image_width, "data-large-height": post.large_image_height, "data-tags": post.tag_string, alt: post.presenter.humanized_essential_tag_string, "data-uploader": post.uploader.name, "data-rating": post.rating, "data-flags": post.status_flags, "data-parent-id": post.parent_id, "data-has-children": post.has_children?, "data-has-active-children": post.has_active_children?, "data-score": post.score, "data-fav-count": post.fav_count, "itemprop": "contentUrl") %>
+  <%= content_tag(:picture) do -%>
+    <%= tag.source media: "(min-width: 660px)", srcset: post.tagged_file_url -%>
```
lllusion3469, line 3: Wouldn't this cause the original url to be show for users with `user.default_image_size == "large"`?
Couldn't you just leave this one out entirely and let the browser fall back on the `` tag below?

app/views/posts/partials/show/_image.html.erb
```diff
@@ -1,3 +1,7 @@
...
+    <%= tag.source media: "(max-width: 660px)", srcset: post.tagged_large_file_url -%>
```
lllusion3469, line 4: Isn't this a problem if [`post.tagged_large_file_url` doesn't actually exist](https://github.com/danbooru/danbooru/blob/82c09858f2cafa0ea1636251ad0c9341ff4dcfc7/app/models/post.rb#L230-L234)? I guess the browser would get a 404 and fall back on the `` below but still, why even let it try?

app/views/posts/partials/show/_image.html.erb
```diff
@@ -1,3 +1,7 @@
...
+    <%= tag.img width: post.image_width_for(CurrentUser.user), height: post.image_height_for(CurrentUser.user), id: "image", class: "fit-width", "data-original-width": post.image_width, "data-original-height": post.image_height, "data-large-width": post.large_image_width, "data-large-height": post.large_image_height, "data-tags": post.tag_string, alt: post.presenter.humanized_essential_tag_string, "data-uploader": post.uploader.name, "data-rating": post.rating, "data-flags": post.status_flags, "data-parent-id": post.parent_id, "data-has-children": post.has_children?, "data-has-active-children": post.has_active_children?, "data-score": post.score, "data-fav-count": post.fav_count, itemprop: "contentUrl", src: post.file_url_for(CurrentUser.user) %>
```
lllusion3469, line 5: wouldn't it be more appropriate to put most of those attributes (including `id="img"`) on the `<picture>` tag? it would probably break some custom css but that's probably the case even as it is now.

app/views/posts/partials/show/_image.html.erb
BrokenEagle: The tagged URLs will return the non-tagged URLs if that option is not enabled.

app/views/posts/partials/show/_image.html.erb
BrokenEagle: All of the image and data attributes added are all on the `#image` element, and having the ID also on that element makes it slightly easier for the users to access. Also, anything that gets done to the `` element gets done to the image in general, so if the user does something like:

```javascript
$("#image").hide();
```

then the image will be hidden, even if if the original `src` set on the image isn't what is currently being shown. In fact, if there isn't an `` element at all in that grouping, then the whole thing doesn't work at all. I learned that to my own detriment after trying it out myself. In fact, as I understand it, the image element is what always has the image loaded onto it, and the `<source>` elements are only changing the current `src` that is set on the `` element.

app/views/posts/partials/show/_image.html.erb
BrokenEagle: > Wouldn't this cause the original url to be show

Yeah, I missed that.

> Couldn't you just leave this one out entirely and let the browser fall back on the  tag below?

To be honest, I was only copying what was being done on post previews for consistency. However, if it's fine to do away with that, I can go ahead and remove that.

app/views/posts/partials/show/_image.html.erb
lllusion3469: Yeah, sorry. I didn't actually check and assumend [this check](https://github.com/danbooru/danbooru/blob/6d78d22571fbaa637e1a97a1aa1a9f39a34e0e09/app/models/post.rb#L173) was there for a reason and that `StorageManager#file_url` would behave [like `#file_path`](https://github.com/danbooru/danbooru/blob/6d78d22571fbaa637e1a97a1aa1a9f39a34e0e09/app/logical/storage_manager.rb#L85), [which it doesn't](https://github.com/danbooru/danbooru/blob/6d78d22571fbaa637e1a97a1aa1a9f39a34e0e09/app/logical/storage_manager.rb#L62)

app/views/posts/partials/show/_image.html.erb
lllusion3469: Ok it seems you're right about that, `<picture>` is a bit weird.
[The WhatWG says this in the HTML spec](https://html.spec.whatwg.org/multipage/embedded-content.html#the-picture-element):

> Also, the picture element itself does not display anything; it merely provides a context for its contained img element that enables it to choose from multiple URLs.

Apparently I made some of my custom css for image previews a lot more complicated than necessary.

app/views/posts/partials/show/_image.html.erb
BrokenEagle: Yeah actually, doing that fixed the issue of double image loading that I noted in [my prior comment](#issuecomment-640105366), even for Chrome.

lllusion3469:
Apparently some people are really interested in being able to see the full resolution on mobile: https://danbooru.donmai.us/forum_topics/9127?page=307#forum_post_168268

But if they had "Default image width" set to "original" it would
* still show the large size
* but not show the "view original" button on the top of the page, I'm fairly sure

So if this `<picture>` approach is kept, it might be desirable to display this link / message regardless of what "Default image width" is set to on mobile, maybe via css media queries(?)

BrokenEagle:
If they want to see the full size, then they can just use the desktop mode instead.

evazion:
Not having the "view original" link will be an issue. This will probably still require some Javascript to detect whether we're showing the sample or the original.

[Client hints](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers#Client_hints) could potentially help here, but they're still experimental. Client hints let the browser tell the server what the screen width is with an HTTP header. This way we could tell server-side how big the screen is, and therefore which image to display.

We might also want to respect the [Save Data](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Save-Data) header here. This indicates that the client want to save bandwidth, and therefore probably would want sample images even in desktop mode.

PR #4312, opened 2020-02-25
Fix exceptions caused by the visible function

BrokenEagle:
>  VR-Man said ([forum #163433](https://danbooru.donmai.us/forum_posts/163433)):
> 
> I kept getting this error when I tried to write a feedback:

```
Error

wrong number of arguments (given 0, expected 1)
Details

ArgumentError exception raised
app/models/user_feedback.rb:24:in `visible'
app/controllers/user_feedbacks_controller.rb:17:in `show'
```

### DIFF
--- a/app/controllers/forum_post_votes_controller.rb
+++ b/app/controllers/forum_post_votes_controller.rb
@@ -10,7 +10,7 @@ class ForumPostVotesController < ApplicationController
   end
 
   def create
-    @forum_post = ForumPost.permitted.find(params[:forum_post_id])
+    @forum_post = ForumPost.visible(CurrentUser.user).find(params[:forum_post_id])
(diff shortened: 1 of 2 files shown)


BrokenEagle:
I redid things after reading 0ad42d2, so that the user is explicitly passed into the visible function.

BrokenEagle:
Added another commit from an error reported on Discord by @unbreaky.

> Unbreakable said:
> Trying to vote on a BUR gives me the 
```
undefined method `permitted' for #<Class:0x0000559c693f0b78>
backtrace: [
    "app/controllers/forum_post_votes_controller.rb:13:in `create'"
]
```

evazion:
Fixed before I saw this PR.

PR #4313, opened 2020-02-25
Fix exception on forum votes caused by using outdated permitted

BrokenEagle:
Used the commit from #4312.


evazion:
Sorry, fixed this in bf281df02 but forgot to push.

PR #4320, opened 2020-03-04
Convert notes to use percentages instead of exact coordinates

BrokenEagle:
Created this PR based on prior discussions on Discord with @evazion about the reversion of a commit with 4b9a297. Basically, the note boxes were converted from using exact coordinates into using percentages instead. This facilitates a much easier resizing of the note boxes, and allows for much of the Javascript resizing code to be removed. The only resizing required now is on the note container itself, needing its size updated to match the underlying image, as well as updating the font size based on the size ratio.

To accomplish this, several style elements had to be changed or removed. All style elements were removed from the inner box, both to unify all style elements on the note box itself, but also because any height or width styles set on the inner box heavily conflicted with getting the notes to work correctly. Now the inner note box just acts as a container for the note body of HTML code. The display of `table-cell` was also changed to `flex` for similar reasons as well, and so the centering styles were also changed to match that new display style. Since the note container now has a size, it had its z-index removed and a z-index added to the note boxes and images. This allows the note boxes to still be seen while still facilitating the user being able to right click or directly interact with the underlying image itself.

The image now automatically resizes to fill the width of the screen in responsive mode. Due to this, the "Always resize image" option and links were removed, since it's no longer needed for responsive mode, and the regular mode already has mechanisms in place to swap between the large and original images. The value was left in place though in the model so that the bit position can be reused in the future.

Previously, it was only possible to switch between the large and original image sizes using the keyboard hotkey. To fix this, a "View large" link was placed in the post **Options** section (so that it doesn't conflict with the image when in full size), and it performs the opposite function that the "View original" link does. The link is also capable of downloading the large size of image if needed.

Finally, this PR fixes an issue on Firefox with the `<br>` elements in the note body in which the **line-height** of the outer elements was made useless by the use of any `<br>` elements or carriage returns (which are converted to `<br>` elements). Currently, the only way to overcome this is to apply a **line-height** style to every single one of the `<br>` elements, which is very laborious plus it adds clutter to the note body.

### DIFF
--- a/app/javascript/src/javascripts/notes.js
+++ b/app/javascript/src/javascripts/notes.js
@@ -5,8 +5,6 @@ let Note = {
   HIDE_DELAY: 250,
   NORMALIZE_ATTRIBUTES: ['letter-spacing', 'line-height', 'margin-left', 'margin-right', 'margin-top', 'margin-bottom', 'padding-left', 'padding-right', 'padding-top', 'padding-bottom'],
   COPY_ATTRIBUTES: ['background-color', 'border-radius', 'transform'],
-  BOX_ONLY_ATTRIBUTES: ['transform'],
-  INNER_ONLY_ATTRIBUTES: ['background-color'],
   permitted_style_values: function(attribute, $attribute_child) {
     if ($attribute_child.length === 0) {
       return "";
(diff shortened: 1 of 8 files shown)


BrokenEagle:
Did a small fixup due to a mistakenly changed class on the last commit.

PR #4325, opened 2020-03-06
Related tag enhancements

### LINKED ISSUE #4321: Include overlap count in the related tag API response
I'd like to use that information in a userscript to display the percentage of overlap, and it's information that is already free and available from the **RelatedTagCalculator** as a result of the SQL queries used. It can just be included as a separate field from the information, kind of like the following.

```json
{
  "query": "nishizumi_miho",
  "category": "Character",
  "tags": [
    [
      "nishizumi_miho",
      4
    ],
    [
      "akiyama_yukari",
      4
    ]
  ],
  "tags_overlap": {
    "nishizumi_miho": 5756,
    "akiyama_yukari": 2356,
  },
  "wiki_page_tags": [ ],
  "other_wikis": {  }
}
```
https://danbooru.donmai.us/related_tag.json?search%5Bquery%5D=nishizumi_miho&search%5Bcategory%5D=Character

First change the **tags** function in **RelatedTagQuery** so that it caches the value (to avoid calculating twice), and then add the following function to get the needed hash for the **serializable_hash** function.

```ruby
def tags_overlap
  tags.map { |v| Hash[v.name, v.overlap_count] }.inject(&:merge)
end
```

BrokenEagle:
Fixes #4321. It also adds some other enhancements.

1. Included tag overlap
    
    This is a separate hash so that the values can be quickly looked up. There is no extra processing involved since the SQL queries already adds that information to the tag instances. There's no reason and no easy way to add those for pattern matching, so those types only return a blank hash.

2. Variable limit

    The default is still 25.

3. Settable type

   The user can now specify which type they want to use, regardless of the format of the query or whether a category is specified or not. It also fixes it where the category type wasn't settable for pattern matching.

4.  Included sample count

    This is the amount of samples that were actually used in the calculations. Since the sample size is not always the maximum, the query tag is not always included in the results, and the query term is not always a tag, it helps to know what the population size used was to help determine the percentage of the tag overlap.

5. Facilitate numerical category inputs

    So 0, 1 ,3, 4 and 5 can now be used as well as any of the string variants. Although currently only used by the related tag endpoint in such a way, it could potentially be used in a variety of other endpoints.

### DIFF
--- a/app/logical/related_tag_calculator.rb
+++ b/app/logical/related_tag_calculator.rb
@@ -6,13 +6,14 @@ module RelatedTagCalculator
     search_sample_size = [search_count, search_sample_size].min
     return [] if search_sample_size <= 0
 
-    tags = frequent_tags_for_search(tag_query, search_sample_size: search_sample_size, category: category).limit(tag_sample_size)
+    tags, sample_count = frequent_tags_for_search(tag_query, search_sample_size: search_sample_size, category: category)
+    tags = tags.limit(tag_sample_size)
     tags = tags.sort_by do |tag|
       # cosine distance(tag1, tag2) = 1 - {{tag1 tag2}} / sqrt({{tag1}} * {{tag2}})
       1 - tag.overlap_count / Math.sqrt(tag.post_count * search_count.to_f)
     end
 
-    tags
+    [tags, sample_count]
   end
 
   def self.frequent_tags_for_search(tag_query, search_sample_size: 1000, category: nil)
@@ -24,7 +25,7 @@ module RelatedTagCalculator
     tags = tags.where("tags.post_count > 0")
     tags = tags.where(category: category) if category.present?
     tags = tags.order("overlap_count DESC, tags.post_count DESC, tags.name")
-    tags
+    [tags, sample_posts.length]
   end
 
   def self.frequent_tags_for_posts(posts)
@@ -36,7 +37,7 @@ module RelatedTagCalculator
     Cache.get("similar_tags:#{tag_query}", cache_timeout, race_condition_ttl: 60.seconds) do
       ApplicationRecord.with_timeout(search_timeout, []) do
         CurrentUser.without_safe_mode do
-          RelatedTagCalculator.similar_tags_for_search(tag_query).take(max_tags).pluck(:name)
+          RelatedTagCalculator.similar_tags_for_search(tag_query)[0].take(max_tags).pluck(:name)
         end
       end
     end
--- a/app/controllers/related_tags_controller.rb
+++ b/app/controllers/related_tags_controller.rb
@@ -4,9 +4,10 @@ class RelatedTagsController < ApplicationController
   def show
     query = params[:query] || search_params[:query]
     category = params[:category] || search_params[:category]
+    type = params[:type] || search_params[:type]
     limit = params[:limit]
 
-    @query = RelatedTagQuery.new(query: query, category: category, user: CurrentUser.user, limit: limit)
+    @query = RelatedTagQuery.new(query: query, category: category, type: type, user: CurrentUser.user, limit: limit)
     respond_with(@query)
   end
 end
--- a/app/controllers/related_tags_controller.rb
+++ b/app/controllers/related_tags_controller.rb
@@ -4,8 +4,9 @@ class RelatedTagsController < ApplicationController
   def show
     query = params[:query] || search_params[:query]
     category = params[:category] || search_params[:category]
+    limit = params[:limit]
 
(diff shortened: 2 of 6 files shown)


BrokenEagle:
Did a small fixup on the first commit since there was a small error in breaking up the commits, so an earlier commit had something it shouldn't have yet. Then there was another one since that something that was removed from the earlier commit wasn't moved to the needed later commit.

evazion:
Overlap count shouldn't be relied on too strongly. It's an approximation that is known to be inaccurate in some cases, especially when the overlap between two tags is small. This should be treated as a peek at the internal implementation that doesn't come with any stability guarantees. I don't want to be locked into a particular implementation of related tags because downstream users have come to rely on them being calculated in a certain way.

BrokenEagle:
> Overlap count shouldn't be relied on too strongly. It's an approximation that is known to be inaccurate in some cases, especially when the overlap between two tags is small.

Yeah, that's what I was expecting. It's similar to the percentages that get displayed with my [DisplayPostInfo script](https://github.com/BrokenEagle/JavaScripts/blob/master/displaypostinfo.user.js), where the percentages there are only as accurate as the number of posts currently displayed on the page. Still, even though the data isn't definitive, it's still useful and informative.

app/logical/related_tag_calculator.rb
```diff
@@ -24,7 +25,7 @@ def self.frequent_tags_for_search(tag_query, search_sample_size: 1000, category:
     tags = tags.where("tags.post_count > 0")
     tags = tags.where(category: category) if category.present?
     tags = tags.order("overlap_count DESC, tags.post_count DESC, tags.name")
-    tags
+    [tags, sample_posts.length]
```
lllusion3469: did you mean `posts`?

in `self.frequent_tags_for_search` the variable is called `sample_posts` but it's passed to `self.frequent_tags_for_post_relation` as the value for the `posts` parameter

changing it to `posts` fixes it but - as there seems to be some kind of caching involved - changing it back didn't break it again *for the same search*, for a different one it would break again.

---

On my instance it causes errors such as this:
```
Error

undefined local variable or method `sample_posts' for RelatedTagCalculator:Module
Details

    ActionView::Template::Error exception raised
    app/logical/related_tag_calculator.rb:32:in `frequent_tags_for_post_relation'
    app/logical/related_tag_calculator.rb:21:in `frequent_tags_for_search'
    app/logical/related_tag_calculator.rb:9:in `similar_tags_for_search'
    app/logical/related_tag_calculator.rb:44:in `block (3 levels) in cached_similar_tags_for_search'
    app/logical/current_user.rb:93:in `without_safe_mode'
    app/logical/related_tag_calculator.rb:43:in `block (2 levels) in cached_similar_tags_for_search'
    app/models/application_record.rb:117:in `with_timeout'
    app/logical/related_tag_calculator.rb:42:in `block in cached_similar_tags_for_search'
    app/logical/cache.rb:18:in `get'
    app/logical/related_tag_calculator.rb:41:in `cached_similar_tags_for_search'
    app/presenters/post_set_presenters/post.rb:43:in `similar_tags'
    app/presenters/post_set_presenters/post.rb:28:in `related_tags'
    app/presenters/post_set_presenters/post.rb:13:in `tag_set_presenter'
    app/presenters/post_set_presenters/post.rb:59:in `tag_list_html'
    app/views/posts/index.html.erb:10
    app/views/posts/index.html.erb:1
    app/controllers/application_controller.rb:39:in `respond_with'
    app/controllers/posts_controller.rb:20:in `index'
```

BrokenEagle: Yeah, I don't know how this happened, but there was a merge conflict with 9ddf408ec5152cc011a634cc8cde7f17e321669f, even though Github never reported a merge conflict when I submitted this 4 days ago.

evazion: I'm removing the `sample_count` field. Calculating the sample count forces an extra query for all related tags calculations, even outside the API. The sample count should always be `min(search_count, 1000)`. For simple single tag searches, you can use the count of the largest tag in the `tags_overlap` field.

There was also a bug with the related tags button returning 250 tags instead of 25 during tag editing.

BrokenEagle: We should probably return the post count data as well then. That should be easy enough, since the data is also there as well for free.

BrokenEagle: Actually, nevermind, that may not be so useful. I'm thinking of an idea of including the post count for all of the displayed tags, but that would be better done as a separate query to the tags interface. Additionally, the post count of the tags wouldn't help for those times where the related tags query involves more than just simple tags, in which case a query to the post count endpoint would need to be done.

PR #4329, opened 2020-03-07
Update blacklist controls

### LINKED ISSUE #4323: The tag script function is playing havoc with custom CSS
Because of 869142e the entire thumbnail is being replaced, however it is not reprocessing the blacklist so the items remain without either the "blacklisted" or the "blacklisted-active" classes. This causes the thumbnails to completely disappear for those users which are using the no-flicker custom CSS from [forum #132142](https://danbooru.donmai.us/forum_posts/132142).

This could be solved simply by reprocessing the entire blacklist by calling `Danbooru.Blacklist.apply()`, but that's a whole lot of processing waste. This could instead be an opportunity to fix that so that individual posts can be processed, as well as fix the overly wasteful process of rendering a disabled blacklist. Currently when a blacklist is "disabled", it has to completely process every post with every blacklist line to create the blacklist controls, only to do the whole thing over again when it hides all of the posts.

BrokenEagle:
Fixes #4323. Besides that, it also adds several other improvements.

1. Fixed several issues with the post update JS render

2. Fixed multiple performance issues with the blacklist apply function being applied multiple times

3. Refactored regexp split used in several places into its own utility function

4. Adjust how classes get added and renamed the classes on the links to be more accurate

5. Used alternate iterators as `$.each` is a subpar performer

<hr>

One thing I realized doing this is that module functions/variables are referred to in a lot of places with **this** which is dangerous with the current standard format. Since most of the anonymous functions use the full function format instead of arrow functions, the **this** inside those refer to the anonymous function instead of the module. So for all of the modules, the anonymous functions should be replaced with arrow functions or all of the module usage of **this** should be converted to using the module name instead.

### DIFF
--- a/app/javascript/src/javascripts/blacklists.js
+++ b/app/javascript/src/javascripts/blacklists.js
@@ -118,23 +118,27 @@ Blacklist.apply = function() {
   var count = 0
 
   $.each(this.posts(), function(i, post) {
-    var post_count = 0;
-    $.each(Blacklist.entries, function(j, entry) {
-      if (Blacklist.post_match(post, entry)) {
-        entry.hits += 1;
-        count += 1;
-        post_count += 1;
-      }
-    });
-    if (post_count > 0) {
-      Blacklist.post_hide(post);
-    } else {
-      Blacklist.post_unhide(post);
-    }
+    count += Blacklist.apply_post(post);
   });
 
   return count;
 }
+
+Blacklist.apply_post = function(post) {
+  var post_count = 0;
+  $.each(Blacklist.entries, function(j, entry) {
+    if (Blacklist.post_match(post, entry)) {
+      entry.hits += 1;
+      post_count += 1;
+    }
+  });
+  if (post_count > 0) {
+    Blacklist.post_hide(post);
+  } else {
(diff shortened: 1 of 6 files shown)


app/javascript/src/javascripts/blacklists.js
```diff
@@ -41,16 +41,17 @@ Blacklist.parse_entries = function() {
...
     return entry.tags === tags;
   })[0];
   if (match) {
     match.disabled = !match.disabled;
     if (match.disabled) {
-      Blacklist.post_hide(e.target);
+      $link.addClass("blacklisted-inactive");
     } else {
-      Blacklist.post_unhide(e.target);
+      $link.removeClass("blacklisted-inactive");
```
evazion: `post_hide` was called intentionally here to fix a problem with `video_with_sound` posts playing even when blacklisted (#2787). I'm not sure if this is still a problem.

BrokenEagle: Ah, but it's not a post that's being passed in there, but the links in the blacklist menu. That wouldn't do anything to the video playing on the page regardless.

PR #4344, opened 2020-03-19
Fix renaming artists

### LINKED ISSUE #4303: Unable to rename artist entries with wiki pages
> Arcana55 said ([forum #163316](https://danbooru.donmai.us/forum_posts/163316)):
If an artist has a wiki, it can't be manually renamed at all, it raises an error.

I've confirmed this for myself. If the artist with a wiki is renamed first, it raises an error. However, if the wiki is renamed first and then the artist, then it doesn't raise an error.

BrokenEagle:
Fixes #4303. 

It also provides a method to rename the wiki page along with the artist. It currently starts out as unchecked since I wasn't sure that renaming the wiki should be the default.

Besides the above, it now redirects the artist wiki page to the artist when it exists. This is because there is now no reason to show the artist wiki pages separately in a different view, since everything is accessible from the artist page. This also goes along with how DText wiki links are rendered, in that they are changed to artist links when the tag is of type artist.

### DIFF
--- a/app/controllers/artists_controller.rb
+++ b/app/controllers/artists_controller.rb
@@ -49,6 +49,9 @@ class ArtistsController < ApplicationController
   end
 
   def update
+    if params[:artist][:wiki_page_attributes] && params[:artist][:wiki_page_attributes][:id] && params[:artist][:name] && params[:artist][:name] != @artist.name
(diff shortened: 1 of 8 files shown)


BrokenEagle:
Already fixed with ca00563.

PR #4345, opened 2020-03-20
Add ability to render expanded DText quotes

### LINKED ISSUE #2018: dmail: Turn nested quote block into expand block
Viewing/previewing dmail only:

Since the Respond link wraps the old message in a quote block, a Respond to that Respond wraps _that_ in a quote block, etc...; you end up with deeply-nested quote blocks taking up a lot of space.

Second level quote blocks should be displayed as expand blocks with the first line in the block as the label, e.g. `[expand=RaisingK said:]`.  The block remains `[quote]` in the raw text.

`QuoteA --> QuoteB --> QuoteC` displays as `QuoteA --> ExpandB --> QuoteC`.  When the user responds to that, the message displays as `QuoteX --> ExpandA --> QuoteB --> QuoteC`.

BrokenEagle:
Fixes #2018.

### DIFF
--- a/app/controllers/dtext_previews_controller.rb
+++ b/app/controllers/dtext_previews_controller.rb
@@ -2,7 +2,8 @@ class DtextPreviewsController < ApplicationController
   def create
     @inline = params[:inline].to_s.truthy?
     @disable_mentions = params[:disable_mentions].to_s.truthy?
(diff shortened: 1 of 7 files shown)


BrokenEagle:
Closing this per the discussion on Discord in #technical. I.e. quoted portions in Dmails will be removed when doing a reply.

PR #4350, opened 2020-03-21
Fix relationship preview alt text overflowing

### LINKED ISSUE #4348: Thumbnail boxes for posts with missing thumbnails have enormous width in the parent/child relationship preview
https://danbooru.donmai.us/posts/2770038 example. 
image

BrokenEagle:
Fixes #4348.


evazion:
I'm just going to remove the tags from the alt text entirely. They bloat the html and they're probably not beneficial for SEO (and possibly even detrimental if Google views them as keyword stuffing).

PR #4369, opened 2020-03-28
Various post version fixes

### LINKED ISSUE #4364: No way to see tag vandalism anymore
Before, you could see tag vandalism easily by looking at dark red/dark green changes in a user's post history. Now reverted tags are not shown differently anymore. This is a pretty serious loss of functionality.

Also ties to issue #4354 because this makes spotting differences much harder.

### LINKED ISSUE #4365: Post Versions: switching to advanced search should preserve current search parameters
Example:
https://danbooru.donmai.us/post_versions?search%5Bupdater_id%5D=508240&type=current
Click on "advanced", it redirects you to https://danbooru.donmai.us/post_versions/search and all search parameters are lost. It's expecially painful for large searches.

### LINKED ISSUE #4366: Post Versions search: Change "version" to dropdown and give it a clearer name
It's confusing what the current "version" field does in the post search 
image

It's not clear unless you're involved in the technicalities of the site that "version" means "version relative to each post". I didn't even know it did that until I asked in the discord.
A better way to have this parameter is turning it into a dropdown, maybe renamed to something like "Include uploaded version", and allow the parameters "Yes", "No", and "Only", with "No" setting the search to >1 and "Only" setting it to =1. This would help make the field clearer. 

"No" IMO should be the default. because the user case for that interface is that of searching for actual versions, not for uploaded versions, which end up taking a lot of space and are only useful to see the tags an user uploaded with (and that's already visible through the tag change report, so it's unlikely someone would be searching for that via the advanced search).

BrokenEagle:
Fixes #4364, fixes #4365, fixes #4366. 

1. Alternate post versions HTML format

    Converts the column **Tags** and places all of the current state of the post at that time (to included unchanged tags) into it. It also supports tag categories and was partially based off of the display for comments.

    Adds an additional column **Edits** and lists only the actual items that were added and removed for that particular post version. It also does away with the "+" and "-" to facilitate ease in copying. Finally, it brings back the "obsolete" category for the "subsequent" and "previous" views.

    #### Current
    image

    The "previous" mode does not have any additional comparisons to make against since those are the comparison values, so it has no "obsolete" category. No tag changes are shown on the "New" versions even though the tags are new, since that would just be a doubling up of the information found in the **tags** column.

    #### Previous
    image

    Since the "current" mode is most like the old presentation of the post versions, it is now the new default mode for that model. Besides those, it also fixes the post drill-down link in the **Versions** column to also include the type parameter.

2. Add additional Search operators

    The **is_new** parameter is a more user friendlier version of the **versions** parameters with a focus on the first versions.  The **tag_matches** parameter facilitates searching for a single tag, which it partially enforces by taking the first non-whitespace block of text as input. Searching for one or more tags would require doing a tag split on the "tags" parameter in the SQL, something which would undoubtedly cause it to fail for even single tag searches.

3. Advanced search parameters

    Clicking the "Advanced search" link will retain all of the current search parameters. This was left  off the secondary menu link because it may be desired to return to quickly return to a fresh search state.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -42,8 +42,8 @@ class ApplicationController < ActionController::Base
     super
   end
 
-  def set_version_comparison
-    params[:type] = %w[previous subsequent current].include?(params[:type]) ? params[:type] : "previous"
+  def set_version_comparison(default_type = "previous")
+    params[:type] = %w[previous subsequent current].include?(params[:type]) ? params[:type] : default_type
   end
 
   def model_name
--- a/app/controllers/post_versions_controller.rb
+++ b/app/controllers/post_versions_controller.rb
@@ -5,7 +5,7 @@ class PostVersionsController < ApplicationController
   respond_to :js, only: [:undo]
 
   def index
-    set_version_comparison
+    set_version_comparison("current")
     @post_versions = authorize PostVersion.paginated_search(params)
 
     if request.format.html?
--- a/app/helpers/post_versions_helper.rb
+++ b/app/helpers/post_versions_helper.rb
@@ -1,46 +1,51 @@
 module PostVersionsHelper
   def post_version_diff(post_version, type)
+    return "" if type == "previous" && post_version.version == 1
+
     other = post_version.send(type)
 
-    this_tags = post_version.tag_array
-    this_tags << "rating:#{post_version.rating}" if post_version.rating.present?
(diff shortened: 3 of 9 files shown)


evazion:
> Converts the column Tags and places all of the current state of the post at that time (to included unchanged tags) into it.

I'm skeptical of this. I think in most cases showing the full unchanged state will be distracting and will take up a lot of space.

PR #4394, opened 2020-04-07
Show or hide post vote buttons when (un)favoriting

lllusion3469:
Not sure if this is at all the proper way to do this, but it seems to work(?)

When favoriting/unfavoriting a post, it would also be upvoted/"un"voted internally but the upvote/downvote/undo vote buttons would stay shown/hidden as they were before unlike when pressing those buttons.

For example when you want change from favorite to merely upvoted you would have to reload the page in between.

Condition taken from here: https://github.com/danbooru/danbooru/blob/6d615001e4fcfdeb123f0f64cc47f3f30a044fed/app/views/posts/partials/show/_information.html.erb#L29

JS taken from: https://github.com/danbooru/danbooru/blob/6d615001e4fcfdeb123f0f64cc47f3f30a044fed/app/views/post_votes/create.js.erb#L4-L5
and https://github.com/danbooru/danbooru/blob/6d615001e4fcfdeb123f0f64cc47f3f30a044fed/app/views/post_votes/destroy.js.erb#L4-L5

### DIFF
--- a/app/views/favorites/_update.js.erb
+++ b/app/views/favorites/_update.js.erb
@@ -4,6 +4,14 @@ $("#score-for-post-<%= @post.id %>").text(<%= @post.score %>);
 $("#favcount-for-post-<%= @post.id %>").text(<%= @post.fav_count %>);
 $(".fav-buttons").toggleClass("fav-buttons-false").toggleClass("fav-buttons-true");
 
+<% if @post.can_be_voted_by?(CurrentUser.user) %>
+  $("#vote-links-for-post-<%= @post.id %>").show();
+  $("#unvote-link-for-post-<%= @post.id %>").hide();
+<% else %>
+  $("#vote-links-for-post-<%= @post.id %>").hide();
+  $("#unvote-link-for-post-<%= @post.id %>").show();
+<% end %>
+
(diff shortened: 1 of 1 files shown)


lllusion3469:
It might be more better to give `#post-info-score` a class and hide `#vote-links-for-post-#{post.id}`
`#unvote-links-for-post-#{post.id}` with CSS. In JS you would then just have to `.toggle` that class like it's already done with the favorite button.

If you want to, I could change it to that.
The only problem is the ids: `span[id^="vote-links-for-post-"]`works but is kind of ugly.

Also: two classes like with the favorite button or just one class and `:not(.foo)`?

evazion:
> It might be more better to give `#post-info-score` a class and hide `#vote-links-for-post-#{post.id}`, and `#unvote-links-for-post-#{post.id}` with CSS. In JS you would then just have to `.toggle` that class like it's already done with the favorite button.

Having `data-post-is-voted` and `data-post-is-favorited` attributes on the <body> tag would be even better. We do this with `data-post-current-image-size` to control visibility of the view original links. Putting state on the <body> tag lets you write CSS rules that target more of the page and it keeps more of the state in one place.

PR #4399, opened 2020-04-08
Post#normalized_source: Normalize more twitter urls

nonamethanks:
This normalizes the following urls:

 * https://twitter.com/ichizon7/status/1244475516180021248/photo/1 (caused by clicking on a picture) ([15k posts](https://danbooru.donmai.us/posts?tags=source%3A*twitter*%2Fphoto%2F*&ms=1 ))
 * https://twitter.com/ichizon7/status/1244475516180021248?s=20 (the number is variable, typically between 18-20 but I've seen other numbers too. Used to track the type of device it was shared from)  ([600 posts](https://danbooru.donmai.us/posts?tags=source%3A*twitter*%3Fs%3D*&ms=1)) 
 * https://mobile.twitter.com/ichizon7/status/1244475516180021248 (regularly fixed by multiple users, should be done server-side) 
 
It also normalizes urls with an ending slash (https://mobile.twitter.com/ichizon7/status/1244475516180021248/).

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -337,7 +337,7 @@ class Post < ApplicationRecord
       end
 
       case source
-      when %r{\Ahttps?://twitter.com/[^/]+/status/(\d+)\z}i
+      when %r{\Ahttps?://(?:mobile\.)?twitter.com/[^/]+/status/(\d+)(?:/(?:photo/\d)?|\?s=\d+)?\z}i
         "https://twitter.com/i/web/status/#{$1}"
 
       when %r{\Ahttps?://lohas\.nicoseiga\.jp/priv/(\d+)\?e=\d+&h=[a-f0-9]+}i,


lllusion3469:
Seems mostly fine
* can there be more than 10 images in one post? (you only have a single digit - `\d`)
* can there be both `/photo/\d+` and `?s=\d+`? in your regex it only matches one or the other
* is there any case when the stuff after `/{user}/status/\d+` actually matters? if not why not just get rid of the `\z`?

nonamethanks:
* Nope, twitter only supports a maximum of 4 images per post
* The ?s part is overwritten by the /photo part if you click on a picture, so I don't think that's possible. 
As for the line termination, I'm not sure but there's several cases in the url normalization where it's present, so I didn't want to touch anything done by convention. If it's not needed it should be removed from all cases.

evazion:
> is there any case when the stuff after /{user}/status/\d+ actually matters? if not why not just get rid of the \z?

I usually try to make regexes match tightly so that if the url format changes we're not suddenly matching things we didn't mean to. There have been bugs in the past caused by regexes matching too loosely.

Also it would be a good idea to either put the example urls in the code as comments above the regex or as tests in `test/unit/post_test.rb`. We have a lot of complex regexes where it's not clear what they're supposed to be matching.

PR #4401, opened 2020-04-08
tumblr: pick biggest image based on resolution

### LINKED ISSUE #4324: Add support for 2048 Tumblr image links
Tumblr now has a 2048 image format, so this new size needs to be checked for and used when it is available. The following link can be used to test any updates.

https://kome-no-id.tumblr.com/image/190996209038

lllusion3469:
`photo[:alt_sizes]` may contain a bigger image than
`photo[:original_size]`

fixes #4324


lllusion3469:
There may be some tests that need to be fixed/added but I would have to set up and environment where I could actually run them  first.

PR #4416, opened 2020-04-16
Artstation: artists can have underscore in their name

nonamethanks:
Fixes an issue with urls like https://hosi_na.artstation.com/projects/3oEk3B where the site gives up when getting the api response (so we miss the commentary, tags etc).

See: https://danbooru.donmai.us/source.json?url=https://hosi_na.artstation.com/projects/3oEk3B

### DIFF
--- a/app/logical/sources/strategies/art_station.rb
+++ b/app/logical/sources/strategies/art_station.rb
@@ -9,6 +9,7 @@
 # * https://www.artstation.com/artist/sa-dui
 # * https://www.artstation.com/sa-dui
 # * https://sa-dui.artstation.com/
+# * https://hosi_na.artstation.com
 #
 # Image URLs
 #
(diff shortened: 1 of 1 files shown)


lllusion3469:
What about these three lines: https://github.com/danbooru/danbooru/blob/3166c236c12cc85f21a6fce877fcf63b72721c1b/app/logical/sources/strategies/art_station.rb#L25-L27

e.g.
```
[1] pry(main)> Sources::Strategies::ArtStation.new("https://sa-dui.artstation.com/").artist_name
=> "sa-dui"
[2] pry(main)> Sources::Strategies::ArtStation.new("https://hosi_na.artstation.com/").artist_name
=> nil
```

nonamethanks:
Fixed, no idea how I missed them.

PR #4445, opened 2020-05-04
Add hentai-foundry support

### LINKED ISSUE #3437: Hentai-Foundry - Artist finder broken
Uploading images from Hentai-Foundry prevents user from finding artist entry.
Interestingly, after uploading, the artist finder + source redirect work as intended.

nonamethanks:
Also fixes #3437.

My first real contribution. Hopefully I didn't make too many mistakes.
I wasn't sure how much testing was required so I tried to cover all user cases I could think of. 

For the agent, it's basically useless to try and cache the cookie that allows browsing, because it expires extremely fast even on normal browsers.

### DIFF
--- a/app/logical/sources/strategies/hentai_foundry.rb
+++ b/app/logical/sources/strategies/hentai_foundry.rb
@@ -1,0 +1,112 @@
+# Image URLs
+#
+# * https://pictures.hentai-foundry.com/a/Afrobull/795025/Afrobull-795025-kuroeda.png
+#
+# Page URLs
+#
+# * https://www.hentai-foundry.com/pictures/user/Afrobull/795025/kuroeda
+# * https://www.hentai-foundry.com/pictures/user/Afrobull/795025o
+#
+# Preview URLs
+#
+# * https://thumbs.hentai-foundry.com/thumb.php?pid=795025&size=350
+#
+# Profile URLs
+#
+# * https://www.hentai-foundry.com/user/kajinman/profile
+# * https://www.hentai-foundry.com/pictures/user/kajinman
+# * https://www.hentai-foundry.com/pictures/user/kajinman/scraps
+# * https://www.hentai-foundry.com/user/J-likes-to-draw/profile
+
+module Sources
+  module Strategies
+    class HentaiFoundry < Base
+      BASE_URL =    %r!\Ahttps?://(?:www\.)?hentai-foundry\.com!i
+      PAGE_URL =    %r!#{BASE_URL}/pictures/user/(?<artist_name>[\w-]+)/(?<illust_id>\d+)(?:/[\w-]*)?(\?[\w=]*)?\z!i
+      PROFILE_URL = %r!#{BASE_URL}/(?:pictures/)?user/(?<artist_name>[\w-]+)(?:/[a-z]*)?\z!i
+      IMAGE_URL =   %r!\Ahttps?://pictures\.hentai-foundry\.com/\w/(?<artist_name>[\w-]+)/(?<illust_id>\d+)/[\w-]+\.\w+\z!i
+
+      def domains
+        ["hentai-foundry.com"]
+      end
+
+      def site_name
+        "Hentai Foundry"
+      end
+
+      def image_urls
+        image = page&.search("#picBox img")
+
+        return [] unless image
+
+        image.to_a.map { |img| URI.join(page_url, img["src"]).to_s }
+      end
+
+      def preview_urls
+        image_urls.map do
+          "https://thumbs.hentai-foundry.com/thumb.php?pid=#{illust_id}&size=250"
+        end
+      end
+
+      def page_url
+        return nil if illust_id.blank? || artist_name.blank?
+        "https://www.hentai-foundry.com/pictures/user/#{artist_name}/#{illust_id}"
+      end
+
+      def page
+        return nil if page_url.blank?
+
+        doc = Cache.get("hentai-foundry:#{page_url}", 1.minute) do
+          HTTParty.get("#{page_url}?enterAgree=1").body
+        end
+
+        Nokogiri::HTML(doc)
+      end
+
+      def tags
+        tags = page&.search(".boxbody [rel='tag']") || []
+
+        tags.map do |tag|
+          [tag.text, URI.join(page_url, tag.attr("href")).to_s]
+        end
+      end
+
+      def artist_name
+        urls.map { |url| url[PROFILE_URL, :artist_name] || url[PAGE_URL, :artist_name] || url[IMAGE_URL, :artist_name] }.compact.first
+      end
+
+      def canonical_url
+        image_url
+      end
+
+      def profile_url
+        return nil if artist_name.blank?
+        "https://www.hentai-foundry.com/user/#{artist_name}"
+      end
+
+      def artist_commentary_title
+        page&.search("#picBox .imageTitle")&.text
+      end
+
+      def artist_commentary_desc
+        page&.search("#descriptionBox .picDescript")&.to_html
+      end
+
+      def dtext_artist_commentary_desc
+        DText.from_html(artist_commentary_desc).gsub(/\A[[:space:]]+|[[:space:]]+\z/, "").gsub(/\n+/, "\n")
+      end
+
+      def normalizable_for_artist_finder?
+        artist_name.present?
+      end
+
+      def normalized_for_artist_finder?
+        url =~ PROFILE_URL
+      end
+
+      def illust_id
+        url[PAGE_URL, :illust_id] || url[IMAGE_URL, :illust_id]
+      end
+    end
+  end
+end
--- a/test/unit/sources/hentai_foundry_test.rb
+++ b/test/unit/sources/hentai_foundry_test.rb
@@ -1,0 +1,89 @@
+require 'test_helper'
+
+module Sources
+  class HentaiFoundryTest < ActiveSupport::TestCase
+    context "The source for a hentai foundry picture" do
+      setup do
+        @image_1 = Sources::Strategies.find("https://www.hentai-foundry.com/pictures/user/Afrobull/795025/kuroeda")
+        @image_2 = Sources::Strategies.find("https://pictures.hentai-foundry.com/a/Afrobull/795025/Afrobull-795025-kuroeda.png")
+      end
+
+      should "get the illustration id" do
+        assert_equal("795025", @image_1.illust_id)
+        assert_equal("795025", @image_2.illust_id)
+      end
+
+      should "get the artist name" do
+        assert_equal("Afrobull", @image_1.artist_name)
+        assert_equal("Afrobull", @image_2.artist_name)
+      end
+
+      should "get the artist commentary title" do
+        assert_equal("kuroeda", @image_1.artist_commentary_title)
+        assert_equal("kuroeda", @image_2.artist_commentary_title)
+      end
+
+      should "get profile url" do
+        assert_equal("https://www.hentai-foundry.com/user/Afrobull", @image_1.profile_url)
+        assert_equal("https://www.hentai-foundry.com/user/Afrobull", @image_2.profile_url)
+      end
+
+      should "get the image url" do
+        assert_equal("https://pictures.hentai-foundry.com/a/Afrobull/795025/Afrobull-795025-kuroeda.png", @image_1.image_url)
+        assert_equal("https://pictures.hentai-foundry.com/a/Afrobull/795025/Afrobull-795025-kuroeda.png", @image_2.image_url)
+      end
+
+      should "get the canonical url" do
+        assert_equal("https://pictures.hentai-foundry.com/a/Afrobull/795025/Afrobull-795025-kuroeda.png", @image_1.canonical_url)
+        assert_equal("https://pictures.hentai-foundry.com/a/Afrobull/795025/Afrobull-795025-kuroeda.png", @image_2.canonical_url)
+      end
+
+      should "download an image" do
+        assert_downloaded(1_349_887, @image_1.image_url)
+        assert_downloaded(1_349_887, @image_2.image_url)
+      end
+
+      should "get the tags" do
+        assert_equal([["elf", "https://www.hentai-foundry.com/search/index?query=elf&search_in=keywords"]], @image_1.tags)
+        assert_equal([["elf", "https://www.hentai-foundry.com/search/index?query=elf&search_in=keywords"]], @image_2.tags)
+      end
+
+      should "find the correct artist" do
+        @artist = FactoryBot.create(:artist, name: "Afrobull", url_string: @image_1.url)
+        assert_equal([@artist], @image_1.artists)
+        assert_equal([@artist], @image_2.artists)
+      end
+    end
+
+    context "An artist profile url" do
+      setup do
+        @site = Sources::Strategies.find("https://www.hentai-foundry.com/user/Afrobull/profile")
+      end
+
+      should "get the profile url" do
+        assert_equal("https://www.hentai-foundry.com/user/Afrobull", @site.profile_url)
+      end
+
+      should "get the artist name" do
+        assert_equal("Afrobull", @site.artist_name)
+      end
+
+      should "get the normalized url" do
+        assert_equal("https://www.hentai-foundry.com/user/Afrobull", @site.normalize_for_artist_finder)
+      end
+    end
+
+    context "A deleted picture" do
+      setup do
+        @image = Sources::Strategies.find("https://www.hentai-foundry.com/pictures/user/faustsketcher/279498")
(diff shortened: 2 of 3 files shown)


app/logical/sources/strategies/hentai_foundry.rb
```diff
@@ -0,0 +1,116 @@
...
+      end
+
+      def illust_id
+        url[PAGE_URL, :illust_id] || url[IMAGE_URL, :illust_id]
+      end
+
+      def agent
+        mech = Mechanize.new
+
+        mech.get("https://www.hentai-foundry.com/?enterAgree=1")
```
lllusion3469, line 109: For my personal danbooru instance I also wrote a hentai-foundry strategy (never bothered upstreaming it because it's fairly low-quality)

Here's how how I do the `page` method:
```rb
def page
  return nil unless page_url.present?
  doc = Cache.get("hf:#{page_url}", 1.minute) do 
    HTTParty.get("#{page_url}?enterAgree=1").body
  end
  Nokogiri::HTML(doc)
end
```

While it's also not optimal, here's a few points:
* it uses `?enterAgree=1` directly on the page we want instead of first visiting the root page
* it doesn't use mechanize, though it does still use httparty
  * my impression is that the `Danbooru::Http` wrapper for `httprb` seems to be the prefered http client now (?)
  * `httprb` doesn't seem to like cookies - when it follows a redirect it just seems to throw away the cookies it got
  * it uses manual caching, which seems to have mostly been removed in favour of `Danbooru::Http`, it does use caching, though, unlike yours

So I don't know what the preferred way to do it would be, but even if you keep using uncached Mechanize, you should still be able to use `?enterAgree=1` directly.

One thing I noticed, though, is
* when visiting [`https://www.hentai-foundry.com/pictures/user/Afrobull/795025?enterAgree=1`](https://www.hentai-foundry.com/pictures/user/Afrobull/795025?enterAgree=1&) (NSFW) it 301 redirects
* when visiting [`http://www.hentai-foundry.com/?enterAgree=1`](http://www.hentai-foundry.com/?enterAgree=1) it immediately responds with a 200 OK

test/unit/sources/hentai_foundry_test.rb
```diff
@@ -0,0 +1,89 @@
...
+        assert_equal("https://www.hentai-foundry.com/user/Afrobull", @image_2.profile_url)
+      end
+
+      should "get the image url" do
+        assert_equal("https://pictures.hentai-foundry.com/a/Afrobull/795025/Afrobull-795025-kuroeda.png", @image_1.image_url)
+        assert_equal("https://pictures.hentai-foundry.com/a/Afrobull/795025/Afrobull-795025-kuroeda.png", @image_2.image_url)
+      end
+
+      should "get the canonical url" do
+        assert_equal("https://www.hentai-foundry.com/pictures/user/Afrobull/795025", @image_1.canonical_url)
```
lllusion3469, line 37: should the `page_url` be the canonical source?
I thought the `image_url` was prefered if it was possible to reconstruct the `page_url` from it?

Fwiw: for [post #3858510](https://danbooru.donmai.us/posts/3858510) (NSFW), the `source` is what would be the `image_url`

app/logical/sources/strategies/hentai_foundry.rb
```diff
@@ -0,0 +1,116 @@
...
+# * https://www.hentai-foundry.com/pictures/user/kajinman/scraps
+# * https://www.hentai-foundry.com/user/J-likes-to-draw/profile
+
+module Sources
+  module Strategies
+    class HentaiFoundry < Base
+      BASE_URL =    %r!\Ahttps?://(?:www\.)?hentai-foundry\.com!i
+      PAGE_URL =    %r!#{BASE_URL}/pictures/user/(?<artist_name>[\w-]+)/(?<illust_id>\d+)(?:/[\w-]*)?\z!i
+      PROFILE_URL = %r!#{BASE_URL}/(?:pictures/)?user/(?<artist_name>[\w-]+)(?:/[a-z]*)?\z!i
+      IMAGE_URL =   %r!\Ahttps?://pictures\.hentai-foundry\.com/\w/(?<artist_name>[\w-]+)/(?<illust_id>\d+)/[\w-]+\.\w+\z!i
```
lllusion3469, line 27: [This code](https://github.com/danbooru/danbooru/blob/86e3bd8fbc6b11cee543e7cee3747b907456c9af/app/models/post.rb#L373-L374) could theoretically be deduplicated now, somewhat like [this](https://github.com/danbooru/danbooru/blob/86e3bd8fbc6b11cee543e7cee3747b907456c9af/app/models/post.rb#L362-L363)

app/logical/sources/strategies/hentai_foundry.rb
```diff
@@ -0,0 +1,116 @@
...
+        return nil if e.response_code.to_i == 404
+        raise
+      end
+
+      def tags
+        tags = page&.search(".boxbody [rel='tag']")&.to_a
+
+        return [] unless tags
+
+        tags.map(&:text)
```
lllusion3469, line 70: shouldn't it return an array of `[tag_name, tag_page]` for each tag, like the other strategies, e.g. [pixiv](https://github.com/danbooru/danbooru/blob/86e3bd8fbc6b11cee543e7cee3747b907456c9af/app/logical/sources/strategies/pixiv.rb#L222)

test/unit/sources/hentai_foundry_test.rb
nonamethanks: You're right, I'll change it.

app/logical/sources/strategies/hentai_foundry.rb
lllusion3469: Should never, under any circumstances, cause any web requests, though, if you do this.
That's why the method for DA is named <code>page_url_<b>from_image_url</b></code>

app/logical/sources/strategies/hentai_foundry.rb
nonamethanks: Yeah, the reason I didn't use ?enterAgree=1 directly on the post is because of the redirect. 
Also in some instances afterwards it redirected me to the main page instead of the post, not sure why. I can't seem to replicate it right now though. I'll try doing some mass tests and see if it was just an issue with my browser.

Using Danbooru::Http has the added benefit that fetching deleted sources correctly returns 404, unlike with mechanize, so a proper 404 error can be thrown. How would I go to allow redirections? The wrapper doesn't seem to make the [HTTP.follow](https://github.com/httprb/http/wiki/Redirects) method available.

app/logical/sources/strategies/hentai_foundry.rb
nonamethanks: Yeah, I didn't want to touch that part of the code because evazion said it all needs a revamp, and the normalization is already being done for this site there. Maybe in a future pull request.

app/logical/sources/strategies/hentai_foundry.rb
nonamethanks: Correct, my mistake.

app/logical/sources/strategies/hentai_foundry.rb
lllusion3469: From what I can tell, `.get(url, follow: {max_hops: 1})` seems to work - except that as I mentioned it doesn't actually work because httprb doesn't like cookies. As far as I can tell it doesn't actually keep the cookies it got from the first response (the redirect) and makes the second one without any cookies - which obviously doesn't work.

lllusion3469: See https://github.com/httprb/http/issues/264

lllusion3469: See /httprb/http / issues/264.
I deleted the first comment because they might not like a "Add hentai-foundry support" mention on their issues but it keeps them anyways... I don't want to add another one, though

nonamethanks: Yeah, github saves those even if they're deleted. No way to get rid of it now.  Same reason I added "fixes" in the pull request body rather than the commit.

Anyway, yeah, I'm not sure if there's a better way to do this than your method. HTTParty also returns 404 correctly on deleted urls, so it's preferable in any case to mechanize. I'll add that and add a check for deleted illustrations.

lllusion3469: So you'd have to do it manually, except:
```rb
[31] pry(main)> HTTP.get("https://www.hentai-foundry.com/").cookies
=> #<HTTP::CookieJar:0x0634f398
 @store=
  #<HTTP::CookieJar::HashStore:0x0634f2f0
   @gc_index=1,
   @gc_threshold=150,
   @jar=$CENSORED,
   @logger=nil,
   @mon_data=#<Monitor:0x0634f290>,
   @mon_data_owner_object_id=74448>>
[32] pry(main)> Danbooru::Http.cache(1.minute).get("https://www.hentai-foundry.com/").cookies
ArgumentError: bad argument (expected URI object or URI string)
from /usr/lib/ruby/2.7.0/uri/common.rb:739:in `URI'
```

nonamethanks:
I have fixed the mentioned issues. Thanks a lot for the input, I'm just starting out on ruby so most stuff is still a bit new to me. 
Does everything look fine now?

app/logical/sources/strategies/hentai_foundry.rb
```diff
@@ -0,0 +1,114 @@
...
+          HTTParty.get("#{page_url}?enterAgree=1").body
+        end
+
+        Nokogiri::HTML(doc)
+      end
+
+      def tags
+        tags = page&.search(".boxbody [rel='tag']")&.to_a
+
+        return [] unless tags
```
lllusion3469, line 69: nitpick: this seems a bit strange to me:
* you could get rid of all this safe navigation (`&.`) stuff if you did `return [] if page.blank?` at the start of the method
  * then you wouldn't need this `return [] unless tags` because it's just there in case `tags` is nil which should only happen if `page` is nil
* the `.to_a` seems unnecessary


There seems to be nothing wrong with using those but I didn't actually know Nokogiri had a `#search` method - I was always using the `#css` (seems to do the same) and `#at_css` (only returns one result) from [its tutorial](https://nokogiri.org/tutorials/searching_a_xml_html_document.html)

nonamethanks: > because it's just there in case tags is nil which should only happen if page is nil

Pictures can have no tags on hentai-foundry. Or am I misunderstanding you?

lllusion3469: Well, yes, but if `#search` can't find any tags, it just returns an empty Nodeset / array / whatever and `[].map { ... } == []` so there's no need to return early

lllusion3469:
While it's probably fine, I kind of don't like to assume a specific url format like [here](https://github.com/danbooru/danbooru/blob/2acaaee15e78fb8748e8780392c2b828184ef56b/app/logical/sources/strategies/hentai_foundry.rb#L42) (that the `https:` is left out) or [here](https://github.com/danbooru/danbooru/blob/2acaaee15e78fb8748e8780392c2b828184ef56b/app/logical/sources/strategies/hentai_foundry.rb#L72) (domain left out).

This could kind of be handled differently like this:
```rb
uri = Addressable::URI.heuristic_parse(href)
uri.site ||= BASE_URL
uri.scheme ||= "https"
uri.to_s
```
but it's also not really that pretty...

otherwise it seems fine to me, though I only looked at the code, haven't tested it.

I don't really have much experience with ruby either, only programmed it to modify danbooru, nothing else

app/logical/sources/strategies/hentai_foundry.rb
nonamethanks: Hm, not sure why i did it like this, but right now doing 
```ruby

      def tags
        tags = page&.search(".boxbody [rel='tag']") || []

        tags.map do |tag|
          [tag.text, "https://www.hentai-foundry.com" + tag.attr("href")]
        end
      end
```

Still works. That's fine too right?

I took the "or empty array" from https://github.com/danbooru/danbooru/blob/master/app/logical/sources/strategies/nijie.rb#L121

nonamethanks:
For the domain, I also took that from the Nijie example. IMO https should be fine since the site supports it, and i very much doubt any site going forward will ever go back to http. That can probably be reworked though when [this mess](https://github.com/danbooru/danbooru/blob/86e3bd8fbc6b11cee543e7cee3747b907456c9af/app/models/post.rb#L336) is rewritten.

lllusion3469:
What I meant was: what if they changed the format e.g. by adding the `https:` in front of the `//` instead of making it a relative url. I don't like to hardcode stuff like this when it's not necessary.

A better way to do it (than what I wrote above) would probably be this, though:
```rb
[12] pry(main)> page_url = "https://www.hentai-foundry.com/pictures/user/Afrobull/792459"
[13] pry(main)> img = "//pictures.hentai-foundry.com/a/Afrobull/792459/Afrobull-792459-fubuki_1.png"
[14] pry(main)> tag = "/search/index?query=opm&search_in=keywords"
[15] pry(main)> URI::join(page_url, img).to_s
=> "https://pictures.hentai-foundry.com/a/Afrobull/792459/Afrobull-792459-fubuki_1.png"
[16] pry(main)> URI::join(page_url, tag).to_s
=> "https://www.hentai-foundry.com/search/index?query=opm&search_in=keywords"
```
which is what the browser does

nonamethanks:
Ah, I see what you mean. I'll fix it tomorrow and rebase it.

nonamethanks:
Done. I also changed preview_urls to use illust_id rather than extracting from the regex again.

lllusion3469:
I guess it's reasonable to copy what the rest of the code base does but in case evazion looks at this:

Nokogiri has these search methods:
* [`#at`](https://nokogiri.org/rdoc/Nokogiri/XML/Searchable.html#search-instance_method): like `#search` but only one result
* [`#at_css`](https://nokogiri.org/rdoc/Nokogiri/XML/Searchable.html#at_css-instance_method): like `#css` but only one result
* [`#at_xpath`](https://nokogiri.org/rdoc/Nokogiri/XML/Searchable.html#at_xpath-instance_method): like `#xpath` but only one result
* [`#css`](https://nokogiri.org/rdoc/Nokogiri/XML/Searchable.html#css-instance_method) searches by CSS rule
* [`#xpath`](https://nokogiri.org/rdoc/Nokogiri/XML/Searchable.html#xpath-instance_method) searches by XPath query
* [`#search` / `#/`](https://nokogiri.org/rdoc/Nokogiri/XML/Searchable.html#search-instance_method) decides using [this nice regex](https://nokogiri.org/rdoc/Nokogiri/XML/Searchable.html#LOOKS_LIKE_XPATH-constant) whether the query looks like XPath or CSS and then behaves mostly like `#css` / `#xpath`

and `#search` definitely seems like the worst option to me, especially if you only want one result.

evazion:
> Nokogiri has these search methods:

I don't think I ever paid close attention to which search methods we used, but `#css` and `#at_css` sound like the best options.

app/logical/sources/strategies/hentai_foundry.rb
evazion: If you're wondering about why I use `.to_a` in certain places, this is a trick to convert `nil` to an empty array, which lets us avoid extra conditionals in some places.

app/logical/sources/strategies/hentai_foundry.rb
evazion: Leaving `Post#normalized_source` alone is fine for now, that can be cleaned up at some later point.

app/logical/sources/strategies/hentai_foundry.rb
evazion: I'm moving towards standardizing on using the `Danbooru::Http` wrapper for `httprb` for a few reasons:

* We directly or indirectly pull in too many dependencies on too many different http libraries. We have `httprb`, `HTTParty`, `Mechanize`, `Faraday` (pulled in by the `oauth2` gem), `Net::Http::Persistent` (pulled in by mechanize`), `Net::Http` (the ruby standard http library, which mechanize and httparty are just wrappers around).
* Having this many HTTP libraries makes it hard to configure logging or proxies.
* HTTParty doesn't support persistent connections. Meaning that if we have to make multiple HTTP calls for a strategy, we have to establish a new connection for each call. This adds noticeable latency.
* Mechanize has a bug we have to work around (see `config/initializers/mechanize_patch.rb`) and anyway, it's just a wrapper for `Net::HTTP` and `Nokogiri` with automatic cookie handling, but usually we have to manually handle cookies anyway for caching purposes. Mechanize is also designed for HTML scraping, not for JSON APIs, so we end up needing to use another HTTP library anyway for non-HTML calls.
* `httprb` supports persistent connections and has a cleaner API than the other libraries. It doesn't have automatic support for cookies or Nokogiri, but this isn't too much trouble to do ourselves (notwithstanding the cookie redirect bug).

Issue #4219, opened 2019-12-01
Deviantart broke download urls when not logged in


lllusion3469:
Forum: https://danbooru.donmai.us/forum_topics/9127?page=287#forum_post_161291
Related: mikf/gallery-dl#488

The download links that the `extended_fetch` returns stopped working (returning 404) when not logged in.

Unless there's a way to use the official API again (i.e. find the UUID) the simplest fix seems to be to export the cookies (`auth`, `auth_secure`, `userinfo`) from a browser once when logged in **and** keeping them updated when making requests from danbooru code (https://github.com/mikf/gallery-dl/issues/445#issuecomment-544151370).

Automatically logging in from danbooru seems to be virtually impossible with all the captchas.

I'm don't know where best to store the cookies - file system/redis/postgres.

Unlike `mechanize`, the `http` gem doesn't make it easy to keep track of cookies: `HTTP::Response#cookies` only contains the modified cookies and `HTTP::Client` only the original ones. With `mechanize` you could just do a `m.cookie_jar.load(path, format: :cookies_txt)` before the request and a `m.cookie_jar.save(path, format: :cookies_txt)` afterwards.

Edit: It's not that much code to `HTTP::CookieJar.new.load(...)` some cookies, to `http.cookie(jar)` (add) them, and after the request to `jar.add()` every cookie in `response.cookies.cookies` and then `jar.save(...)` it.

Edit: some kind of caching might also make sense: https://github.com/r888888888/danbooru/issues/4144#issuecomment-526820593

Edit: They also broke [this](part) (when not downloadable) though the fix should be fairly simple as they only changed the structure of the reply

Edit: [here](https://github.com/lllusion3469/danbooru/commits/da_napi_cache)'s some poc code

evazion:
My observations:

* The only cookies we need are `auth`, `authsecure`, and `userinfo`.
* `auth` appears to be the only cookie we have to refresh. I haven't seen the others change during the login session.
* You have to send the cookies with the first `extended_fetch` call. If you don't, you can still fetch the data, but the download url you get won't work.
* If your `auth` cookie is expired you might get a 404 with a new cookie, in which case you have to try again with the new cookie.
* It is possible to disable Eclipse and go back to the old design where UUIDs are still present. I don't want to go back to the old API though unless we have to, it's not a long term solution.

lllusion3469:
Just as a reminder: it's still not fixed
[...]

PR #4457, opened 2020-05-11
Fix Deviantart

### LINKED ISSUE #4431: Uploading from DeviantArt doesn't work
When fetching the link this error shows:
```
bad argument (expected URI object or URI string)
```

The artist, as well as the DA tags are fetched correctly.
Upon uploading a new message shows up:
```
Upload #2914215
Date: 2020-04-29 17:57:25 -0400
An error occurred: error: NoMethodError - undefined method `hostname' for nil:NilClass

    app/logical/downloads/file.rb:97:in `is_cloudflare?'
    app/logical/downloads/file.rb:66:in `uncached_url'
    app/logical/downloads/file.rb:31:in `download!'
    app/logical/upload_service/utils.rb:190:in `get_file_for_upload'
    app/logical/upload_service.rb:51:in `start!'
    app/controllers/uploads_controller.rb:55:in `create'
```

lllusion3469:
Since I never got any replies to any of my comments anywhere I just implemented it how I would find it acceptable since it's been broken for about half a year now (#4219).

I chose to get rid of the JS backend API client (`extended_fetch`) entirely as not to have two clients (for one). This meant I might as well go back to before it was introduced as a known good state. See https://github.com/danbooru/danbooru/issues/4431#issuecomment-621518209 for details.

This first reverts `app/logical/sources/strategies/{deviant_art,stash}.rb` to how they were in 7f482dc35bc7b77bce580ec06b4b708d49d3a29a with some very minor changes. I couldn't figure out how to to this other than `git checkout $commit -- $files`

Counting redirects this makes 3-5 non-image requests for what was 1-2 with `extended_fetch` but seeing as evazion never implemented caching (for `extended_fetch`) it doesn't seem he would care much(?).

Alternatives:
* use API only for downloading and do the rest with `extended_fetch`
* use API only for downloading and getting the image_url for non-downloadable posts when `extended_fetch` can't handle it very well (e.g. flash files) and otherwise still use `extended_fetch`

The (DA and stash) tests pass for me (got it to work)

This obviously requires client id and secret to be added back to the config if they were removed.

Doesn't make use of `.persist` (which `Danbooru::Http` doesn't have) connections (unless httprb somehow does it automatically) but I don't know how best to add that (make a HTTP::Client in the strategy and pass it to the api client methods?).

Should `preview_url` be added? If so, what are the criteria? Would any small image about the height of the batch upload previews do?

fixes #4431

also why no trailing commas?

Not necessarily finished but working

### DIFF
--- a/app/logical/deviant_art_api_client.rb
+++ b/app/logical/deviant_art_api_client.rb
@@ -15,22 +15,22 @@
 
 class DeviantArtApiClient
   class Error < StandardError; end
-  BASE_URL = "https://www.deviantart.com/api/v1/oauth2"
+  BASE_URL = "https://www.deviantart.com/api/v1/oauth2/"
 
-  attr_reader :client_id, :client_secret, :httparty_options
+  attr_reader :client_id, :client_secret
 
-  def initialize(client_id, client_secret, httparty_options = {})
-    @client_id, @client_secret, @httparty_options = client_id, client_secret, httparty_options
+  def initialize(client_id, client_secret)
+    @client_id, @client_secret = client_id, client_secret
   end
 
   # https://www.deviantart.com/developers/http/v1/20160316/deviation_single/bcc296bdf3b5e40636825a942a514816
   def deviation(uuid)
-    request("/deviation/#{uuid}")
+    request("deviation/#{uuid}")
   end
 
   # https://www.deviantart.com/developers/http/v1/20160316/deviation_download/bed6982b88949bdb08b52cd6763fcafd
   def download(uuid, mature_content: "1")
-    request("/deviation/download/#{uuid}", mature_content: mature_content)
+    request("deviation/download/#{uuid}", mature_content: mature_content)
   end
 
   # https://www.deviantart.com/developers/http/v1/20160316/deviation_metadata/7824fc14d6fba6acbacca1cf38c24158
@@ -43,19 +43,15 @@ class DeviantArtApiClient
       ext_stats: ext_stats,
     }
 
-    request("/deviation/metadata", **params)
+    request("deviation/metadata", **params)
   end
 
   def request(url, **params)
-    options = {
-      base_uri: BASE_URL,
-      params: { access_token: access_token.token, **params },
-      headers: { "Accept-Encoding" => "gzip" },
-      format: :plain,
-    }
+    params = { access_token: access_token.token, **params }
 
-    body, code = HTTParty.get(url, **options)
-     JSON.parse(Zlib.gunzip(body), symbolize_names: true)
+    url = URI.join(BASE_URL, url).to_s
+    response = Danbooru::Http.cache(1.minute).get(url, params: params)
+    response.parse.with_indifferent_access
   end
 
   def oauth
--- a/app/logical/deviant_art_api_client.rb
+++ b/app/logical/deviant_art_api_client.rb
@@ -1,59 +1,73 @@
-# https://github.com/danbooru/danbooru/issues/4144
+# Authentication is via OAuth2 with the client credentials grant. Register a
+# new app at https://www.deviantart.com/developers/ to obtain a client_id and
+# client_secret. The app doesn't need to be published.
 #
 # API requests must send a user agent and must use gzip compression, otherwise
 # 403 errors will be returned.
+#
+# API calls operate on UUIDs. The deviation ID in the URL is not the UUID. UUIDs
+# are obtained by scraping the HTML page for the <meta property="da:appurl"> element.
+#
+# * https://www.deviantart.com/developers/
+# * https://www.deviantart.com/developers/authentication
+# * https://www.deviantart.com/developers/errors
+# * https://www.deviantart.com/developers/http/v1/20160316
 
-DeviantArtApiClient = Struct.new(:deviation_id) do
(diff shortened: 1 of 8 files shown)


evazion:
Apologies for not getting back to you before. The last time I looked at DeviantArt the situation basically looked untenable:

* The new API requires manual authentication and it's impossible for us to stay authenticated for any length of time.
* The new Eclipse layout is React-based, which makes it very difficult to scrape without using a headless browser.
* The old API probably won't be around forever, considering that DeviantArt is redesigning everything and using a different API for their own frontend. Also it rests entirely on being able to scrape the UUID, which is only found in one place, so if we lose that we're done.

> Counting redirects this makes 3-5 non-image requests for what was 1-2 with extended_fetch but seeing as evazion never implemented caching (for extended_fetch) it doesn't seem he would care much(?).

It's less than ideal but it's fine.

> Should preview_url be added? If so, what are the criteria? Would any small image about the height of the batch upload previews do?

Preview images must be at least 150x150 and not be cropped. Some sites have cropped thumbnails, so we have to be careful to avoid that. The preview URL is also used for IQDB lookups, so it needs to match the full image.

It doesn't matter if the preview is bigger than 150x150, this is just an optimization to avoid downloading the full image if we can avoid it.

lllusion3469:
That was fast!

> Also it rests entirely on being able to scrape the UUID, which is only found in one place, so if we lose that we're done.

That's true but seeing as they added it back to the page and the JS backend and that the reason it's there is their app we might be fine for a bit longer, I hope.

> it's impossible for us to stay authenticated for any length of time.

Using my cookies.txt approach (see the issues) I've been authenticated for half a year now which is something, I would say. If you hadn't thrown away your cookies (https://github.com/danbooru/danbooru/commit/ea718122daeaedb949649808f4bcf0f5d020a462#r36326037) (if I'm reading that right) it might have worked for you, too. Is your redis instance persistent i.e. is the data even kept as long as would have been necessary?

evazion:
The last time I tried I kept having problems with getting logged out and not being able to recover:

* DeviantArt can invalidate all your login sessions remotely whenever they want. You can see this by logging in with multiple browsers and logging out in one browser. Your other browser will be logged out too.
* DeviantArt logged us out at unpredictable times whenever we did something they didn't like. IIRC this could happen if we made too many requests too quickly, tried to look up deleted posts, or made API calls in an order they didn't like.
* If we get logged out, all our login sessions are invalidated and we can only recover by logging in manually again.

The Redis cache is persistent, the problem is that our cookies would randomly become invalidated and after that we couldn't recover.

lllusion3469:
Since it's been working perfectly fine for me and since we may have to return to this method in the future I'd really be interested to know what you did to get those problems to occur because the only public thing I can go by is ea718122daeaedb949649808f4bcf0f5d020a462 which (unless I'm reading it completely wrong) has multiple problems that meant it could never work right.

In particular (just in case - to make sure we're on the same page):
* [one typo](https://github.com/danbooru/danbooru/commit/ea718122daeaedb949649808f4bcf0f5d020a462#r36326023)
* [one problem](https://github.com/danbooru/danbooru/commit/ea718122daeaedb949649808f4bcf0f5d020a462#r36326037) where the cookies it got weren't actually stored properly:
  * Cookies work through two HTTP headers: The `Cookie` request cookie and the `Set-Cookie` response cookie
  * We send (through `Cookie`) our cookies to authenticate ourselves
  * Sometimes deviantart will include some `Set-Cookie` headers in its response with new authentification information
  * `HTTP::Response#cookies` only contains those new cookies but the old ones that haven't been updated are still valid until they expire
  * [here](https://github.com/danbooru/danbooru/commit/ea718122daeaedb949649808f4bcf0f5d020a462#diff-5d4206c0dba287b91018efa0162db96eR31) you just override the old cookies with (basically a filtered) `response.cookies`
  * old but still valid and necessary cookies that haven't been overridden are thus lost
  * [this](https://github.com/lllusion3469/danbooru/commit/e0083217eda346b11e14c59df1bbfc5686b5785a#diff-5d4206c0dba287b91018efa0162db96eR40-R42) is how I did it (though I don't filter the cookies, I just save all of them, because why not)

lllusion3469:
Getting the size of `https://api-da.wixmp.com/_api/download/file?downloadToken=$TOKEN` by using HEAD requests doesn't work as it responds with a "405 Method Not Allowed".

This affects
* the [`Sources::Strategies::Base#size`](https://github.com/danbooru/danbooru/blob/0a12f35c47b6ea426cf50c59a7d848a05c5135d1/app/logical/sources/strategies/base.rb#L148-L150) method which could by overridden to use the API response which includes a file size (which should be correct but has had bugs in the past)
  * as far as I can tell this isn't used anywhere
* [`UploadService::ControllerHelper.prepare`](https://github.com/danbooru/danbooru/blob/0a12f35c47b6ea426cf50c59a7d848a05c5135d1/app/logical/upload_service/controller_helper.rb#L11-L12) 
  * used (only?) [here](https://github.com/danbooru/danbooru/blob/0a12f35c47b6ea426cf50c59a7d848a05c5135d1/app/views/uploads/_image.html.erb#L5) which isn't a big deal
  * don't know how to fix this as the strategy is long gone by this point

PR #4475, opened 2020-05-20
Refactor source normalization

nonamethanks:
tl;dr: move the source out of app/models/post.rb and into each strategy file.

In the process I fixed several normalizations, since a lot of url normalizations were actually not working. 

I still need to finish writing the tests for the null strategy but in the meanwhile I'd appreciate a review for the rest of the code. (I also noticed a typo in the commit description, that'll be fixed after I rebase it with the rest of the tests tomorrow).

All the tests can be launched at once with the following command:
` rails test test/unit/sources/* -n "/normalizing for source/"`

I explicitly avoided touching some parts of the code like pawoo save for adding a function and some regexes, because that was going out of the scope of this pull and is a bit too complex to lump it all in, but basically that logic should be removed from the api client. Maybe at a later date.

Main changes:
* Move the source normalization logic out of the post model
  and into individual sources' strategies.
* Rewrite normalization tests to be handled into each source's test,
  and expand them significantly. Previously we were only testing
  a very small subset of domains and variants.
* Fix up normalization for several sites.
* Normalize fav.me urls into normal deviantart urls.


app/logical/sources/strategies/deviant_art.rb
```diff
@@ -1,25 +1,36 @@
 # Asset URLs:
 #
-# * http://orig12.deviantart.net/9b69/f/2017/023/7/c/illustration___tokyo_encount_oei__by_melisaongmiqin-dawi58s.png
-# * http://pre15.deviantart.net/81de/th/pre/f/2015/063/5/f/inha_by_inhaestudios-d8kfzm5.jpg
-# * http://th00.deviantart.net/fs71/PRE/f/2014/065/3/b/goruto_by_xyelkiltrox-d797tit.png
+## NORMALIZED
```
lllusion3469, line 3: Sorry, but I find this a bit confusing so let me just check I understand it correctly: You divide these into "will be normalized", "will not be normalized but can be normalized" and "cannot be normalized" which you call "NORMALIZED", "NOT NORMALIZED", and "NOT NORMALIZABLE"

* "normalized" implies to me that it's already normalized
* "not normalized" implies to me that, well, it's not normalized as written, nothing more

Also, since you're already adding some cases why not handle those two under "not normalized" as well, should be fairly easy, shouldn't it? I would think even the `/deviation/{id}` would be better if extracting title and artist is too fiddly.

nonamethanks: I didn't want to make the deviantart regexes too confusing and rework them too much for this pull request, but honestly you're right, since I already added several cases I might as well fix those too and normalize all the normalizable. I'll add it tomorrow when I rebase everything.

nonamethanks: I think I'll add a separate variable for those though, otherwise it'll just become too hard to read and unmantainable.

lllusion3469:
I'm going to take a closer look today/tomorrow but it seems fine to me except for some bikeshedding maybe.

I was wondering how you were going to handle the sources that didn't have a strategy. Didn't think of just putting it in the null strategy.

Seems much better than before all in all.

[I thought I found a bug at first but it was my own strategy (not upstreamed) which broke it... no idea how it even worked til now]

app/logical/sources/strategies/deviant_art.rb
```diff
@@ -204,7 +223,7 @@ def dtext_artist_commentary_desc
       end
 
       def self.deviation_id_from_url(url)
-        if url =~ ASSET
+        if url =~ ASSET or url =~ FAVME
```
lllusion3469, line 226: There's not really a problem here but just so you know:
There is a difference in ruby between `||` and `or` (`&&` and `and`): the difference is precedence

Check these [SO](https://stackoverflow.com/a/2083118) [answers](https://stackoverflow.com/a/2083630) (apparently `and`/`or` are intended for control flow and not logic?)

Since rubocop complains, I guess it should be changed but it complains about so much other stuff prevalent in the code base - even `$~` - so I don't know...

nonamethanks:
Yeah, we already had null.rb so it was the better option over creating more stray files.

Tomorrow I'm also going to extract the top domains/hosts from a bigquery dump and add more normalization cases along with the tests. I wonder if there's a better way to structure such a large function than doing a very long case switch like we're already doing.

Another point: I did some stress tests with 16k sources (the max exportable from bigquery at once) for most domains (and all sources for deviantart's wixmp that we have available on danbooru, around 320) and I couldn't find any error. The average time was around 10 seconds for 16k sources, so half a millisecond per source, so this change shouldn't bring any performance degradation (which is the most important part when refactoring).

app/logical/sources/strategies/deviant_art.rb
nonamethanks: Ah, that slipped past rubocop because I had turned it off for a second to stop blinding me with the regex complaints. I'll fix it.

app/logical/sources/strategies/null.rb
```diff
@@ -24,6 +24,122 @@ def normalizable_for_artist_finder?
...
+        when %r{\Ahttp://www\.karabako\.net/images(?:ub)?/karabako_(\d+)(?:_\d+)?\.}i
+          "http://www.karabako.net/post/view/#{$1}"
+
+        # XXX http://twipple.jp is defunct
+        # http://p.twpl.jp/show/orig/myRVs
+        when %r{\Ahttp://p\.twpl\.jp/show/(?:large|orig)/([a-z0-9]+)}i
+          "http://p.twipple.jp/#{$1}"
+
+        when %r{\Ahttps?://blog(?:(?:-imgs-)?\d*(?:-origin)?)?\.fc2\.com/(?:(?:[^/]/){3}|(?:[^/]/))([^/]+)/(?:file/)?([^\.]+\.[^\?]+)}i
+          username = $1
```
lllusion3469, line 44: I guess it was already this way before but I'd personally prefer named groups instead of first only numbering them and then naming them each the next line, though it would make the regexes longer

nonamethanks: This depends on whether @evazion wants me to write multiline regexes. I didn't see them anywhere in the code so I avoided writing them, though it would look cleaner yeah.

nonamethanks:
Ok, adjusted the above and added some things:

* fixed more zerochan normalizations, as only half the urls were being normalized;
* rule34.xxx was getting the md5 from the post's properties. Now it gets it from the filename;
* removed the exhentai normalization by filename. As far as I know that method doesn't work anymore and hasn't been for a while.

lllusion3469:
Sorry, can't debug it right now but some cases from the deviantart tests are broken:
[Failed test 1](https://github.com/nonamethanks/danbooru/blob/a449c63543124562f496bee2e303fc2a9be32395/test/unit/sources/deviant_art_test.rb#L218)
```rb
url = "http://pre06.deviantart.net/8497/th/pre/f/2009/173/c/c/cc9686111dcffffffb5fcfaf0cf069fb.jpg"
[13] pry(main)> ref = "https://www.deviantart.com/edsfox/art/Silverhawks-Quicksilver-126872896"
[15] pry(main)> s = Sources::Strategies.find(url, ref)
[16] pry(main)> s.page_url_from_image_url
=> "https://www.deviantart.com/edsfox/art/Cc9686111dcffffffb5fcfaf0cf069fb-126872896"
[17] pry(main)> s.page_url
=> "https://www.deviantart.com/edsfox/art/Silverhawks-Quicksilver-126872896"
[18] pry(main)> s.canonical_url
=> "https://images-wixmp-ed30a86b8c4ca887773594c2.wixmp.com/intermediary/f/8b472d70-a0d6-41b5-9a66-c35687090acc/d23jbr4-8a06af02-70cb-46da-8a96-42a6ba73cdb4.jpg?token=$TOKEN"
```

[Failed test 2](https://github.com/nonamethanks/danbooru/blob/a449c63543124562f496bee2e303fc2a9be32395/test/unit/sources/deviant_art_test.rb#L240)
```rb
[23] pry(main)> url = "https://images-wixmp-ed30a86b8c4ca887773594c2.wixmp.com/intermediary/f/8b472d70-a0d6-41b5-9a66-c35687090acc/d23jbr4-8a06af02-70cb-46da-8a96-42a6ba73cdb4.jpg/v1/fill/w_786,h_1017,q_70,strp/silverhawks_quicksilver_by_edsfox_d23jbr4-pre.jpg"
[24] pry(main)> ref = "https://www.deviantart.com/edsfox/art/Silverhawks-Quicksilver-126872896"
[25] pry(main)> s = Sources::Strategies.find(url, ref)
[26] pry(main)> s.page_url_from_image_url
=> "https://www.deviantart.com/edsfox/art/Silverhawks-Quicksilver-126872896"
[27] pry(main)> s.page_url
=> "https://www.deviantart.com/edsfox/art/Silverhawks-Quicksilver-126872896"
[28] pry(main)> s.canonical_url
=> "https://images-wixmp-ed30a86b8c4ca887773594c2.wixmp.com/intermediary/f/8b472d70-a0d6-41b5-9a66-c35687090acc/d23jbr4-8a06af02-70cb-46da-8a96-42a6ba73cdb4.jpg?token=$TOKEN"
```

[Failed test 3](https://github.com/nonamethanks/danbooru/blob/a449c63543124562f496bee2e303fc2a9be32395/test/unit/sources/deviant_art_test.rb#L248)
For some reason I get a `HTTP::Redirector::TooManyRedirectsError` when running it in `rails test` but in `rails console` it seems to work fine... I don't get it. Same for [Failed test 4](https://github.com/nonamethanks/danbooru/blob/a449c63543124562f496bee2e303fc2a9be32395/test/unit/sources/deviant_art_test.rb#L45)

nonamethanks:
The issue is that I get 403 forbidden with Cloudflare in https://github.com/danbooru/danbooru/blob/master/app/logical/sources/strategies/deviant_art.rb#L251 

```ruby
[6] pry(main)> p = s.page_url_from_image_url
=> "https://www.deviantart.com/edsfox/art/Silverhawks-Quicksilver-126872896"
[7] pry(main)> Danbooru::Http.cache(1.minute).get(p)
=> #<HTTP::Response/1.1 403 Forbidden {"Server"=>"CloudFront", "Date"=>"Thu, 21 May 2020 15:05:34 GMT", "Content-Type"=>"text/html", "Content-Length"=>"919", "Connection"=>"close", "X-Cache"=>"Error from cloudfront", "Via"=>"1.1 91c31195544333f46c3622df2c6df353.cloudfront.net (CloudFront)", "X-Amz-Cf-Pop"=>"WAW50-C1", "X-Amz-Cf-Id"=>"DrisOQm7djZzPuPcdugwcVlQJMUCG8HP1nOCz9jABxArnY_6a7khoA=="}>
```

I really have no idea how to work around this. I added a regex to work around unnormalizable assets but I can't debug the rest of the issues due to this problem.

nonamethanks:
Ah, seems it's a recent problem. See mikf/ gallery-dl/issues/ 655. Deviantart fucking up stuff as usual.

Unfortunately for me it fails even on the first try every time - several people have encountered the same problem [judging by this google search.](https://www.google.com/search?q=deviantart+403+cloudfront&oq=deviantart+403+cloudfront&aqs=chrome..69i57.2899j0j1&sourceid=chrome&ie=UTF-8)

nonamethanks:
Ok, I managed to create a proxy on a secondary server of mine and now I can bypass Cloudfront. I'll try to figure out the rest of the tests. Fuck deviantart btw.

nonamethanks:
I did a dirty patch of danbooru's HTTP method to allow me to run the tests through proxy. I should probably create a patch to allow a configurable proxy in the future. But that's for another day.

For failed test 1 and 2 in https://github.com/nonamethanks/danbooru/blob/a449c63543124562f496bee2e303fc2a9be32395/test/unit/sources/deviant_art_test.rb#L218 and https://github.com/nonamethanks/danbooru/blob/a449c63543124562f496bee2e303fc2a9be32395/test/unit/sources/deviant_art_test.rb#L240 I think it's the tests that are wrong, am I correct @lllusion3469 ? We can extract the deviation ID from them, so they should be the canonical URLs according to the current logic. They're like that right now because intermediary URLs weren't being normalized before this pull.

For the too many redirects error, seems it's failing in https://github.com/danbooru/danbooru/blob/364343453c616a13ac215ff370fb37762ffe0f0d/app/logical/downloads/file.rb#L32 but I can't understand why it doesn't like intermediary URLs.

lllusion3469:
I think the problem with those last two tests with the redirect is basically the following:
* When downloading a file [it takes the url and runs it through the strategy again](https://github.com/danbooru/danbooru/blob/master/app/logical/downloads/file.rb#L78)
* In this case:
```rb
[39] pry(main)> s = Sources::Strategies.find("https://noizave.deviantart.com/art/test-no-download-697415967")
Sources::Strategies.find(s.image_url).page_url_from_image_url
[56] pry(main)> Sources::Strategies.find(s.image_url).page_url_from_image_url
=> "https://deviantart.com/deviation/697415967"
```
What changed is that this used to return `nil`.

This is a problem because
```
$ curl -I 'https://deviantart.com/deviation/697415967'
...
location: https://www.deviantart.com/deviation/697415967
...
```
this forwards twice and I set the max_hops to 1

nonamethanks:
Yep, setting the normalization to www fixes it. I'll fix the rest of those tests to add www, and add the regex for those urls.

What about my question for the intermediary canonical source?

lllusion3469:
I edit my comments too much:
> I would argue should be what the test expects because I find it prettier with title and artist - can be argued either way, though. I don't care too much.

lllusion3469:
Also when reviewing your code I sometimes think it might be helpful if you didn't always rebase and force-push because it's not immediately obvious what you changed - especially with such a huge commit. I say that even though I do the same myself. (Though evazion always just merged mine as-is)

I had to test whether you noticed this error in my comment above and fixed it (which you did apparently):
`https://www.deviantart.com/edsfox/art/Cc9686111dcffffffb5fcfaf0cf069fb-126872896`

Fwiw Github has this option if you worry about polluting the master branch
Screenshot_20200521_204837

nonamethanks:
Imo it's better to keep as canonical url the "best" version in the case of deviantart, as they've changed their URL schemas a lot in the past, so it's more convenient for us to save the page url since that's not likely to change too much in the future (so in the form of deviation/<id> or the title-artist schema, rather than the wixmp one).  It's not even guaranteed that the wixmp urls will work in the future.
Removing the canonical_url method from the deviantart strategy makes it fall back to that way of saving it, and since we already go through the api it always saves the title-artist method. 

Yeah, I was rebasing because I didn't want to fill the branch with "wip"s. Lemme try pushing partials.

lllusion3469:
Aren't we agreeing then? I was talking about [this](https://github.com/danbooru/danbooru/blob/2160aad26da10f97d8b8fe85faedb6f48e8085ff/test/unit/sources/deviant_art_test.rb#L58) test which now returns ugly a wixmp url which only contains the deviation_id. This means when normalizing it it would look like this: "/deviation/{deviation_id}" -  but I'd prefer it with the artist and title which are lost.

nonamethanks:
Ok, how does this look? 

Also I see the option to squash and merge, but as it'll be evazion who'll decide to merge this it'd probably be better for me to rebase everything once we're in agreement that it's complete. For now I'll keep pushing WIPs as required in order to make it better readable.

lllusion3469:
If you want to make sure it doesn't get merged while you have commits like "wip" you can set your PR as a "Draft".

Do you know how to use `git rebase -i`? It's fairly simple (although git is always bit intimidating imo)

nonamethanks:
Yeah, that's how I've been merging everything so far, with interactive. 

I didn't realize there was a draft option. It's too late now, as the work is mostly done and the only major thing left is a rebase, but for future ones I'll open them as draft.

lllusion3469:
> Yeah, that's how I've been merging everything so far, with interactive.

you do know about `git commit --amend`, right?

> Ok, how does this look?

pretty good, can't see anything wrong right now, though if you haven't noticed yet I care about DA the most.

> Ok, I managed to create a proxy on a secondary server of mine and now I can bypass Cloudfront. I'll try to figure out the rest of the tests. Fuck deviantart btw.

I forgot you were using some hosted server for your development. I'm just using my desktop PC with my home internet connection and have no problems.

> Fuck deviantart btw.

yeah if everything else didn't also have major problems

<details>
<summary>way off-topic comment</summary>

* pixiv doesn't even have an API we can use, we're still using their mobile app backend, aren't we? [this userscript](https://github.com/kepstin/Fix-pixiv-thumbnails) makes their frontend somewhat manageable at least
* twitter has lots of retweets of other artists, themselves, and lots of text, non-art in between the pics I care about (the "media" tab improves this a bit). And a horrible cut-off preview and no grid view... If I didn't still have an ancient account I would have needed to give them my phone number for API access iirc
* I don't event have to comment about tumblr, do I?
* DA itself also has a horrible frontend now that their redesign is mandatory... I'll have to do some major modifications via user{styles,scripts} the next days to make it bearable

</details>

---

If you're up for some bikeshedding
* because of [this thing](https://github.com/danbooru/danbooru/pull/4475#issuecomment-632265146) the DA strategy will unnecessarily run through all the API requests and the UUID thing again when actually downloading the image when the deviation_id can be extracted. Because of caching the API calls can be avoided but not necessarily the `#page` call. No idea how to avoid this, though.
* the naming: it can be argued (not sure myself) that `#normalize_for_source` shouldn't be named after it's purpose ("normalizing _for_ the source field") but after what it does ("normalize the url")

nonamethanks:
We already have "normalize for artist finder" so I just followed that example for naming.

For api calls, does Deviantart allow image replacement/edit? I'm not sure we want to cache those for too long- we already have a similar problem with Pixiv where uploaders are sometimes unable to upload the latest version because we cache the old url.

I feel like resolving that problem is out of scope of this pull though, as this was just supposed to be an extraction of source normalization from the post model.

lllusion3469:
> We already have "normalize for artist finder" so I just followed that example for naming.

fine by me

> For api calls, does Deviantart allow image replacement/edit?

I think so, yes.

> I'm not sure we want to cache those for too long

It's only cached for 60s

> I feel like resolving that problem is out of scope of this pull though

agreed

> I don't think the single commit can be helped

you could e.g. extract the changes to DA into a separate commit but I don't think it's worth it.

nonamethanks:
Alright, I'll merge that stray wip one final time later if there's nothing else.
I've ran the tests one more time and nothing went wrong save for an already existing fanbox failure.

app/logical/sources/strategies/deviant_art.rb
evazion: `||` and `&&` are preferred in Danbooru code.

lllusion3469:
>>  Also when reviewing your code I sometimes think it might be helpful if you didn't always rebase and force-push because it's not immediately obvious what you changed

I just figured out that Github links to a nice comparison view if you hover over "force-pushed"... I should probably have noticed that much sooner.

Issue #4212, opened 2019-11-16
Facilitate standard fonts for translating


BrokenEagle:
Discussed somewhat in [topic #15327](https://danbooru.donmai.us/forum_topics/15327).

Currently, picking a font while translating is pretty hit or miss, especially with smartphones. This can drastically affect the presentation of the note to the user, compared to how the translator intended it to be seen. Instead, site-wide fonts can be supported using the `@font-face` CSS rule.

https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face

This doesn't need to be an all-inclusive list of fonts, but at least several of the more common fonts like:

- Arial (sans-serif)
- Verdana (sans-serif)
    - standard font used on the site
- Helvetica (sans-serif)
    - standard font used on the site
- Times New Roman (serif)
- Comic Sans MS (cursive)

Perhaps also checking all of the current notes for instances of `font-family` to see what fonts the translators like to use.

evazion:
I added a fallback font for Comic Sans. The other basic fonts (Arial / Verdana / Times New Roman) are already supported on most systems, either by the actual font itself or by a builtin equivalent (i.e. Android uses Noto Serif as a substitute for Times New Roman and Roboto as a substitute for Verdana). I'm using http://fontfamily.io to get this info.

The big problem here is that most fonts used in notes are copyrighted. We can't distribute the font files ourselves without buying commercial licenses, which aren't that cheap. So for every font we want to support we have to find a free alternative that's close enough to the real thing.

BrokenEagle:
As discussed in [topic #16435](https://danbooru.donmai.us/forum_topics/16435), another suggestion for a font which meets the need for a Spooky font was suggested. I'm just adding the conversation here so that it gets documented.

> Zurreak said:
> 
> I have two potential candidates for the scary font.
> 
> The first is Anarchy. It's 100% free. It contains uppercase, lowercase, and basic punctuation, it does not contain a hyphen, accented letters, or special characters.
It has a good capital I, is easy to read, and is stylistically quite similar to the source's style on a lot of the posts I've used Chiller on in the past. I think it would make a good addition to the site.
> 
> The second is Frotten. It's 100% free. It contains uppercase, lowercase, basic punctuation, accented letters, and a variety of special characters.
[...]

PR #4482, opened 2020-05-24
Add monospace font

BrokenEagle:
Part of #4212.  Discussed on [topic #16435](https://danbooru.donmai.us/forum_topics/16435) and Discord's [#translations](https://discordapp.com/channels/310432830138089472/390639683119480863/713152504283529756).

I went with the IBM Plex Mono as suggested by Shinjidude as it looked about the right amount of robotic, plus the italic version looks pretty interesting.


evazion:
We don't need every single weight. For other fonts we only include regular (weight 400), bold (700), and bold italic.

BrokenEagle:
Alright, I updated it so that it only includes normal and bold weights. I chose the 200 weight for normal though instead, because 400 and 300 were doing some weird extra bolding on some of the letters on some of the font sizes.

PR #4487, opened 2020-05-27
Fix invalid artist URLs being allowed

### LINKED ISSUE #4192: Invalid URL links allowed on artist pages
Noticed this issue on the following artist:

https://danbooru.donmai.us/artists/188558

One of the URLs there is:

```
https://https://acrux808.tumblr.com/
```

I was able to perform a save on that artist without changing anything, which means that it still accepts that URL as valid

So URLs should probably be validated. For instance, in the hostname portion, nothing but word characters, periods or dashes should be allowed. I know that the port number can be specified with a colon, but none of the links we should be accepting should have them.

I'm not sure if there are any restrictions for the pathname portion, but if there are, then those should be checked as well.

BrokenEagle:
Fixes #4192. The problem was that the Addressable parser does not catch all invalid URL cases.

I used the following page in part of my search for a regex URL validator.

https://mathiasbynens.be/demo/url-regex

The one provided by **diegoperini** seemed to be the best.

https://gist.github.com/dperini/729294

One of the responders there mentioned creating a Ruby gem, which seems like the easiest to incorporate and maintain.

https://github.com/amogil/url_regex

### DIFF
--- a/app/models/artist_url.rb
+++ b/app/models/artist_url.rb
@@ -20,11 +20,9 @@ class ArtistUrl < ApplicationRecord
       nil
     else
       url = url.sub(%r!^https://!, "http://")
-      url = url.sub(%r!^http://([^/]+)!i) { |domain| domain.downcase }
       url = url.sub(%r!^http://blog\d+\.fc2!, "http://blog.fc2")
       url = url.sub(%r!^http://blog-imgs-\d+\.fc2!, "http://blog.fc2")
       url = url.sub(%r!^http://blog-imgs-\d+-\w+\.fc2!, "http://blog.fc2")
-      # url = url.sub(%r!^(http://seiga.nicovideo.jp/user/illust/\d+)\?.+!, '\1/')
       url = url.sub(%r!^http://pictures.hentai-foundry.com//!, "http://pictures.hentai-foundry.com/")
 
       # XXX should be handled by pixiv strategy.
@@ -105,7 +103,15 @@ class ArtistUrl < ApplicationRecord
   end
 
   def normalize
+    # Perform some normalization with Addressable on the URL itself
+    # - Converts scheme and hostname to downcase
+    # - Converts unicode hostname to Punycode
+    uri = Addressable::URI.parse(url)
+    uri.site = uri.normalized_site
+    self.url = uri.to_s
     self.normalized_url = self.class.normalize(url)
+  rescue Addressable::URI::InvalidURIError
+    # Don't bother normalizing the URL if there is errors
   end
 
   def initialize_normalized_url
--- a/app/models/artist_url.rb
+++ b/app/models/artist_url.rb
@@ -120,9 +120,18 @@ class ArtistUrl < ApplicationRecord
     end
   end
 
+  def validate_scheme(uri)
+    errors[:url] << "'#{uri}' must begin with http:// or https:// " unless uri.scheme.in?(%w[http https])
+  end
+
+  def validate_hostname(uri)
+    errors[:url] << "'#{uri}' has a hostname '#{uri.host}' that does not contain a dot" unless uri.host&.include?('.')
+  end
+
   def validate_url_format
     uri = Addressable::URI.parse(url)
-    errors[:url] << "'#{uri}' must begin with http:// or https:// " if !uri.scheme.in?(%w[http https])
+    validate_scheme(uri)
+    validate_hostname(uri)
   rescue Addressable::URI::InvalidURIError => error
     errors[:url] << "'#{uri}' is malformed: #{error}"
   end
--- a/test/unit/artist_url_test.rb
+++ b/test/unit/artist_url_test.rb
@@ -24,10 +24,18 @@ class ArtistUrlTest < ActiveSupport::TestCase
     end
 
     should "disallow invalid urls" do
-      url = FactoryBot.build(:artist_url, url: "www.example.com")
+      urls = [
+        FactoryBot.build(:artist_url, url: "www.example.com"),
+        FactoryBot.build(:artist_url, url: ":www.example.com"),
+        FactoryBot.build(:artist_url, url: "http://http://www.example.com"),
+      ]
 
-      assert_equal(false, url.valid?)
-      assert_match(/must begin with http/, url.errors.full_messages.join)
+      assert_equal(false, urls[0].valid?)
+      assert_match(/must begin with http/, urls[0].errors.full_messages.join)
+      assert_equal(false, urls[1].valid?)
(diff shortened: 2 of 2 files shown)


lllusion3469:
To be honest, I'm not a huge fan of this solution. The thing is that both of the urls mentioned _are_ valid:
* `https://https://acrux808.tumblr.com/`
  * domain is "https", port is "", path is "//acrux808.tumblr.com/", [all of which is allowed, even an empty string for the port](https://url.spec.whatwg.org/#url-port-string)
  * you could check that the host contains a "."
* `https://twitter.com/intent/use​r?user_id=2685107059`
  * it contains a non-breaking space
  * depending which mode you're in it might or might not be valid but most browsers accept non-urlescaped unicode nowadays so `Addressable::URI` does too
  * the `URI.parse` from the standard library complains as it requires the url to be ascii-only - if you want that you could switch to that

Using a random (0.0.3) gem that uses a regex of all things to "validate and extract URLs" doesn't seem like a great idea if you ask me.

It is, of course, reasonable not to accept every url even if it is valid but just adopting whichever behaviour this random gem has isn't the way to go imo.

If you anyone has the means to check that it might be interesting to know how many artist urls are using non-urlescaped unicode.

BrokenEagle:
If you have a better performing regex, by all means post your findings here. That regex is the best one I was able to find. Also, it's not "random behavior" or whatever. If you actually check the URLs I posted in the OP, you'll see exactly which URLs it accepts and which URLs it rejects.

> The thing is that both of the urls mentioned are valid:

That first URL should be rejected, even if modern browsers can handle it. The point isn't to have URLs that the browser can still handle, but to disallow URLs that aren't in the proper format. Having them be all screwy looks terrible and it makes finding duplicate URLs hard as well.

That second URL is accepted by the URL regex gem, which is fine by me.  Unless of course we want to have a list of characters like the zero-width space that we want to reject no matter what.

> the URI.parse from the standard library complains as it requires the url to be ascii-only - if you want that you could switch to that

I don't want a bunch of URLs which are an explosion of percent encoded strings. The non-percent encoded URLs have a lot more clarity to them, plus they're a lot more compact.

evazion:
> Using a random (0.0.3) gem that uses a regex of all things to "validate and extract URLs" doesn't seem like a great idea if you ask me.
>
> It is, of course, reasonable not to accept every url even if it is valid but just adopting whichever behaviour this random gem has isn't the way to go imo.

I agree that we shouldn't use this gem. It's unmaintained and it's not clear what it does without digging into the source code, at which point we may as well do it ourselves. It would be better to use Addressable::URI to parse the url then validate the url components ourselves.

> `https://https://acrux808.tumblr.com/`

Things like this are things we should reject even if they're technically legal. This is a common mistake in artist entries.

> If you anyone has the means to check that it might be interesting to know how many artist urls are using non-urlescaped unicode.

See here: https://danbooru.donmai.us/artist_urls?search%5Burl_matches%5D=%2F%5B%5E+-~%5D%2F

BrokenEagle:
I'm rewriting things, and the questions I have are:

1. Should Uppercase hostnames be allowed?
2. Should Unicode hostnames be allowed?
3. Should port numbers be allowed?
4. Should double forward slashes in the path be allowed?

Are there any other items besides the above that should be disallowed?

The following are the items that will be part of the check:

5. Scheme must be http or https (lowercase)
6. Must include at least one dot in the hostname
7. Must not be an IP address

Are there any issues with the above?

lllusion3469:
1. since it's the same I would change it lowercase
2. maybe normalize it to the [punycode form](https://en.wikipedia.org/wiki/Punycode) using something like `uri.host = uri.normalized_host`. This would also handle point 1
3. There are already urls using e.g. port 8080 but also ones explicitly using e.g. 80 so why not allow ports. You could get rid of 80 for http and 443 for https if you want to
4. `/` and `//` aren't guaranteed to be the same. maybe only the first one is accepted but in some rare cases the second one may be required for whatever reason
  * I found this url `http://www.pixiv.net//users/9731/` which isn't even valid (anymore?)
5. sure
6. ok
7. [I also found some ip addresses](https://danbooru.donmai.us/artist_urls?commit=Search&search%5Border%5D=id&search%5Burl_matches%5D=%2F%5Ehttps%3F%3A%2F%2F%5B0-9%5D%2B%5B.%5D%5B0-9%5D%2B%5B.%5D%5B0-9%5D%2B%5B.%5D%5B0-9%5D%2B%2F).

BrokenEagle:
Alright, I made the changes suggested. I also split up the validation additions, and then normalizations into two separate commits.

lllusion3469:
Seems good to me

Just three things:
* I just mentioned the existing urls with IPs so you could determine why they were there and if they should stay. I don't have an opinion on whether they should be allowed or not (though in doubt I would lean towards the permissive option).
* Same for `//`. I would be ok with forbidding it if you think it causes too much trouble even if there might be some legitimate use case, whatever that may be.
  * I already mentioned a pixiv url where it even responds with a 404 for me.
* Small nitpick: Couldn't you use `.include?(".")` / `uri.host&.include?(".")` (`nil` is falsey)

BrokenEagle:
I'm not overly familiar with Ruby. Does the `&.` essentially perform the same as something like the try function I use in the code, i.e. it only executes the follow on function if the preceding value is not nil?

But yeah, `include?` can be used as well. I just find myself using regexes more when doing string validation, so that's usually my go to method even for trivial cases.

As for IP addresses, I checked a few of them out and they're still valid, and they didn't have a valid routable hostname that I could find.

For the double forward slash, well, it is valid, and I can imagine that there may be some sight out there that uses such even though it shouldn't.

BrokenEagle:
> Small nitpick: Couldn't you use .include?(".") / uri.host&.include?(".") (nil is falsey)

I made this fix, but left everything else mostly the same. Some other minor changes were made after running rubocop.

lllusion3469:
> I'm not overly familiar with Ruby. Does the &. essentially perform the same as something like the try function I use in the code, i.e. it only executes the follow on function if the preceding value is not nil?

Sorry, was away yesterday. You probably already figured it out but yes, it's mostly the same as ActiveSupport's `#try` (with some exceptions).

See [this blog post](https://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/).

I only ever used ruby for danbooru as well, so I'm really not an expert. I only found this because I didn't know about `#try` and but did know about the feature [in JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) and the [proposal for python](https://www.python.org/dev/peps/pep-0505/).

PR #4496, opened 2020-06-02
Refactor nicoseiga

### LINKED ISSUE #4135: Bookmarklet fails on certain Nicoseiga posts
http://seiga.nicovideo.jp/seiga/im9208126

```
ActionView::Template::Error exception raised
image url not found for (http://seiga.nicovideo.jp/seiga/im9208126, http://seiga.nicovideo.jp/seiga/im9208126)

app/logical/sources/strategies/nico_seiga.rb:48:in `image_urls'
app/logical/sources/strategies/base.rb:75:in `image_url'
app/views/uploads/_image.html.erb:3:in `_app_views_uploads__image_html_erb___998738582218535715_47013196888600'
app/views/uploads/new.html.erb:14:in `_app_views_uploads_new_html_erb___1956998059314765135_47013197004820'
app/controllers/uploads_controller.rb:12:in `new'
```

Probably comes from this post being age-restricted (R-15).

### LINKED ISSUE #4340: Undefined Method error on posts with nicovideo.jp in source field
If you open post [3824122](https://danbooru.donmai.us/posts/3824122) and then press **Edit** to change the data, an error message is shown.

```undefined method `moniker' for nil:NilClass```

 
This seems to affect all posts that have as source nicovideo.jp with the notable exception of posts that have seiga.nicovideo.jp as source.

Search: [source:*nicovideo.jp* -source:*seiga.nicovideo.jp*](https://danbooru.donmai.us/posts?tags=source%3A*nicovideo.jp*+-source%3A*seiga.nicovideo.jp*)

### PR DESCRIPTION
**Pull request by nonamethanks**
Tested against 1000 nicoseiga sources currently on site, extracted at random. Couldn't find any error.

Changes:
####  Switch to Danbooru::Http. 
Get rid of mechanize completely.
I could not figure out a proper way to store cookies with http.rb so I just ended up generating the cookie header and storing that. There were also some bad things we were doing like saving the wrong adult header.

#### Switch to mobile api, improving speed
From my testing, this branch takes 3 seconds for the first call with login to do source.to_h, 2 seconds  on average for all subsequent fetches. Master in comparison takes 11 seconds with first login and 9(!) seconds on average for all subsequent fetches, so this branch is about 5 times faster.
This is because we're now using the mobile api, which is much, much faster than the xml api (which we only call to fetch manga artist names, since those aren't served through the normal json api)

#### Add full support for manga pages
Posts like https://seiga.nicovideo.jp/watch/mg316708 now work properly (on master the bookmarklet only fetches the first picture no matter what). Also use the right manga tag urls.

#### Merge main and manga clients
There wasn't much point in keeping them separate since they share many methods.

#### Add support for anonymous and r-15 images (fixes #4135)
https://seiga.nicovideo.jp/seiga/im520647 (anonymous) and https://seiga.nicovideo.jp/seiga/im9208126 (r-15) now work.

#### Store the right source
As mentioned in the inline comments, unfortunately we were saving the wrong source.    
There's no way to tell apart illust from manga from the direct image url alone. What's worse, nicoseiga itself doesn't know how to normalize back to manga, so if it's not an illust type then t's impossible to get the original manga page back from the image url alone. /source/ links on the other hand correctly redirect, hence we use them to normalize the currently existing saved direct sources on site.
Someone (probably me) will have to manually convert all current lohas.nicoseiga sources on live site to their seiga/im counterpart where possible.

#### Properly ignore links we cannot manipulate (fixes #4340)
We were naively assuming all nicoseiga urls had a corresponding api response. This is not true for videos. Also, allows users to upload oekaki links like  https://dic.nicovideo.jp/oekaki/52833.png, which currently fail on live site.
There's room for improvement, as we could theoretically grab the right artist profile and source metadata for a nicoseiga video, but I didn't want to spend even more time on this so i'll leave it to someone else or myself for the future.

#### Properly link "im" links in the artist commentary
nicoseiga does this automatically. I used our inline "seiga #number".

#### Implement the preview urls method for iqdb
And differentiate between manga and illust thumbs

### CODE REVIEW

**Comment by lllusion3469** [comment]
> I could not figure out a proper way to store cookies with http.rb so I just ended up generating the cookie header and storing that. There were also some bad things we were doing like saving the wrong adult header.

`HTTP::Response#cookies` gets you a `HTTP::CookieJar` which could be passed to `HTTP.cookies` (which isn't available through `Danbooru::Http`) but can't be used in a cache. But `HTTP::CookieJar#cookies` returns a array of `HTTP::Cookie`s which can.
I.e. Something like `HTTP.cookies(response.cookies.cookies)` should work and `response.cookies.cookies` can be cached. The only problem is that `Danbooru::Http` doesn't expose `#cookies`.

---

Also since you're already working on nicovideo I'm mentioning this bug here since I'm not sure what the cause is: [searching for `http://www.nicovideo.jp/user/33118578/`](https://danbooru.donmai.us/artists?commit=Search&search%5Border%5D=created_at&search%5Burl_matches%5D=http%3A%2F%2Fwww.nicovideo.jp%2Fuser%2F33118578%2F) seems to have a bug related to nico seiga:
```
Error

undefined method `user_id' for nil:NilClass
Details

    NoMethodError exception raised
    app/logical/sources/strategies/nico_seiga.rb:110:in `profile_url'
    app/logical/sources/strategies/base.rb:156:in `normalize_for_artist_finder'
    app/models/artist_url.rb:36:in `normalize'
    app/logical/artist_finder.rb:127:in `find_artists'
    app/models/artist.rb:224:in `url_matches'
    app/models/artist.rb:258:in `search'
    app/models/application_record.rb:19:in `paginated_search'
    app/controllers/artists_controller.rb:33:in `index'

```

nonamethanks:
Yes, that error is fixed with this branch.

As for the CookieJar, well, it's funny that you mention that, because it was my first attempt: I tried to hack something together as a preliminary experiment, but then I discovered that it's currently broken: httprb/ http/issues/ 511.

nonamethanks:
After more testing, we can't rely on manga thumbs completely, because it seems nicoseiga also has a secondary domain for this kind of manga (NSFW) https://seiga.nicovideo.jp/watch/mg313445. 
I added support for those too but we have to avoid using preview_urls with them.

nonamethanks:
It seems nicoseiga is now pushing out a new cdn domain for their images. I've added support for it.

PR #4497, opened 2020-06-03
Add ability to use parent post's artist commentary

### LINKED ISSUE #4450: Unify commentary of posts coming from the same source
In Pixiv they all share the same commentary, so changes should replicate across all of them (or alternatively, they should all share the same commentary).

BrokenEagle:
Fixes #4450.

I'm not sure that linking commentary outside of a parent/child relationship needs to be handled at this time, especially since doing so though would not be so simple. I'd say let's see how the users do with this change, and address any shortfalls later on if there are any.


BrokenEagle:
Closing this due to a lack of interest after 2 months.

PR #4504, opened 2020-06-11
Migrate post locks to instance

### LINKED ISSUE #4317: Implement temporary taglist lock for member-level users
This could be a way to stop sockpuppets and vandalizing users from making continuous changes even when reverted, as we've seen a lot lately.

Basically what could be done is to allow builders to temporarily lock (for example, for a week) a post's tags so that only gold+ users may edit them. I say a temporary lock, because there's really no need for it to be indefinite, and that way there's no risk of builders accidentally locking tags forever, given that tag lists are much more used than the other properties that can be already locked, like ratings. 

There's not even a need for the locking time to be settable by builders, it could just be hardcoded, and it'd work as an effective way to prevent freshly registered malicious users from constantly re-reverting their changes.

### LINKED ISSUE #4395: Record note/rating locking somewhere
Perhaps in the [moderation](https://github.com/danbooru/danbooru/issues/new) tab, but there should be a way to track who locked/unlocked something.

BrokenEagle:
Fixes #4317 and fixes #4395. 

Moves the current post locks to their own instance. This allows the history to be recorded, an expiration interval to be set, and more locks to be added or removed without changing the database structure. Additionally it adds accountability for adding/removing locks, as well as recording the bad actors for whom the locks have been added, which would assist with user moderation.

Default period is 7 days, though moderators can change that. Moderators can also set the minimum level to edit the post locks.

More notes on can be found in the commit message.

## Screenshots

Below are screenshots of various page items affected by the post locks along with explanatory notes.

### Index view

When a lock expires naturally, the expires field will show "expired" instead. When the previous lock was edited, it will show "edited". When all locks are removed, it will show "removed".

image

### Dialog view

It shows the current expiration if there is a current active lock, and what the expiration will be set at (if the duration field is not used). Controls are added for All locks and Post locks to allow for quick setting.

image

### Other  screenshots

I dropped many other screenshots on Discord channel that I don't want to clog the screen with.

https://cdn.discordapp.com/attachments/618878216815247372/720767182564491364/unknown.png
https://cdn.discordapp.com/attachments/310846683376517121/719731441872797817/unknown.png
https://cdn.discordapp.com/attachments/310846683376517121/719731170182692974/unknown.png
https://cdn.discordapp.com/attachments/310846683376517121/719733856604782622/unknown.png
https://cdn.discordapp.com/attachments/310846683376517121/719734240794640437/unknown.png
https://cdn.discordapp.com/attachments/618878216815247372/720766929106632764/unknown.png

### DIFF
--- a/app/controllers/artist_commentaries_controller.rb
+++ b/app/controllers/artist_commentaries_controller.rb
@@ -27,6 +27,9 @@ class ArtistCommentariesController < ApplicationController
     post_id = params[:artist_commentary].delete(:post_id)
     @artist_commentary = authorize ArtistCommentary.find_or_initialize_by(post_id: post_id)
     @artist_commentary.update(permitted_attributes(@artist_commentary))
+
+    @has_error = !@artist_commentary.valid?
+    flash[:error] = @artist_commentary.errors.full_messages.join("; ") if @has_error
     respond_with(@artist_commentary)
   end
 
--- a/app/controllers/pool_elements_controller.rb
+++ b/app/controllers/pool_elements_controller.rb
@@ -8,6 +8,9 @@ class PoolElementsController < ApplicationController
 
     @post = Post.find(params[:post_id])
     @pool.add!(@post)
+
+    @has_error = !@pool.valid?
+    flash[:error] = @pool.errors.full_messages.join("; ") if @has_error
     respond_with(@pool)
   end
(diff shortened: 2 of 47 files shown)


BrokenEagle:
Closing this due to lack of interest after 2 months.

PR #4516, opened 2020-06-17
Add NewGrounds support

nonamethanks:


lllusion3469:
If you want to add support for videos as well: It's probably not the prettiest way to do it, but this works ([tested here](https://www.newgrounds.com/portal/view/725554)):
```rb
script = html.at_xpath("//script[contains(text(), 'uploads.ungrounded.net\\/alternate')]")
JSON.parse(script.text[%r{"url":(?<url>".+?")}, :url])
```

In my example it is a fairly high quality video so I don't know what the meaning of `alternate` is.

flash isn't used much but it seems to work there as well (without the `alternate` part in the url) ([example](https://www.newgrounds.com/portal/view/461635))

---

I also don't quite see the point of [making an extra request for the artist name](https://github.com/danbooru/danbooru/pull/4516/files#diff-06d68d75c763dd427a29c52f108e784aR74). Isn't it already present on the art page (something like `.item-details-main > h4 > a`)? I couldn't find a case where the artist name is different from the user name and your tests don't seem to contain one either. Is there some case when we need to be able to get the artist name from something other than the art page or something?

Maybe a test should also be added for posts with multiple images (through this `#author_comments img[data-user-image='1']` thing; is it just a presentation thing or are you actually supposed to abuse comments for this?)

nonamethanks:
> I also don't quite see the point of making an extra request for the artist name. Isn't it already present on the art page (something like .item-details-main > h4 > a)? I couldn't find a case where the artist name is different from the user name and your tests don't seem to contain one either. Is there some case where we need to be able to get the artist name from something other than the art page or something?

Yeah, artist name and user name are different. The artist name can change, while the user name is static. I found a few like that while testing the strategy, but I don't have the cases at hand right now.
https://www.newgrounds.com/supporter You can see here that they let you change your displayed username (not the same thing as the registration name) every 30 days. URLs with the registration name will however still redirect to any new name you pick, AFAIK.

>Maybe a test should also be added for posts with multiple images (this #author_comments img[data-user-image='1'] thing; is it just a presentation thing or are you actually supposed to abuse comments for this?)

Newgrounds doesn't officially allow multi-image uploads, so artists upload them as "comments". The unfortunate result is that all images past the first are resized and are not available in full size. They said they were working on allowing more than one proper image per upload but who knows when they'll do it. 
I'll add a test for making sure we're getting all pics.


I didn't add video support for several reasons, mostly because like in your example there can be multiple artists associated with it, and I don't think we can support that currently, and also danbooru is not really a video site (we even have the limit of max 2 minutes for that reason), so it didn't feel like a priority.

lllusion3469:
> Yeah, artist name and user name are different. The artist name can change, while the user name is static. I found a few like that while testing the strategy, but I don't have the cases at hand right now.
https://www.newgrounds.com/supporter You can see here that they let you change your displayed username (not the same thing as the registration name) every 30 days. URLs with the registration name will however still redirect to any new name you pick, AFAIK.

Yeah, your [comment here](https://github.com/danbooru/danbooru/pull/4516/files#diff-06d68d75c763dd427a29c52f108e784aR85) says about as much. What I was wondering was in what case something like `page.at_css(".item-details-main > h4 > a").text.strip` wasn't enough. That does return the artist name (the one that can change), doesn't it? Also, [here](https://github.com/danbooru/danbooru/pull/4516/files#diff-06d68d75c763dd427a29c52f108e784aR22) (unlike the other regexes) you call it "artist_name", so which one is that?

> I didn't add video support for several reasons, mostly because like in your example there can be multiple artists associated with it, and I don't think we can support that currently, and also danbooru is not really a video site (we even have the limit of max 2 minutes for that reason), so it didn't feel like a priority.

Fair enough (I'll just have to add it downstream), there aren't that [many posts I could find](https://danbooru.donmai.us/posts?tags=source%3A*newgrounds.com%2Fportal*). I only mentioned it because I'd already written the code.

nonamethanks:
I have found a case for redirection: http://theolebrave.newgrounds.com/
For example https://www.newgrounds.com/art/view/theolebrave/sensitive-pochaco this still has the old username. I've added a test for that
I have also changed it to avoid making another request, please see if that's fine.


"alternate" seems to be to allow multiple resolutions: 
If https://uploads.ungrounded.net/alternate/1299000/1299281_alternate_72274.mp4 is the original then https://uploads.ungrounded.net/alternate/1299000/1299281_alternate_72274.360p.mp4 is the variation. They allow 360, 720 and 1080p variations as far as I can see.

lllusion3469:
> I have found a case for redirection: http://theolebrave.newgrounds.com/
> For example https://www.newgrounds.com/art/view/theolebrave/sensitive-pochaco this still has the old username. I've added a test for that
> I have also changed it to avoid making another request, please see if that's fine.

Interesting that they don't change urls and redirect for that...
While making fewer requests is definitely better I was mainly wondering why you did it the way you did it, so yeah.

> As for video support, how are you going to resolve it for multiple artists?

Just overriding [the `#artists` method](https://github.com/danbooru/danbooru/blob/f730951e7f454ee22720665d1d8ab5e493b30ad2/app/logical/sources/strategies/base.rb#L178) is good enough for me in cases like this, don't want to deviate too far from upstream otherwise rebasing is going to be too annoying.
```rb
def artists
  # something like this
  artist_urls.map { |url| ArtistFinder.find_artists(url) }.flatten
end
```
My branch is currently 120 commits ahead of upstream/master, if you're interested, [here's a summary of some of them](https://gist.github.com/lllusion3469/befe1982e2c54bf3136f963ed0f5b62d) but not the commits themselves, that would be too much effort right now. Feel free too comment there.

PR #4517, opened 2020-06-17
Add more tags to post replacements filtering and modqueue highlighting.

### LINKED ISSUE #4509: Remove by default when replacing posts: resolution_mismatch/source_*
Add `resolution_mismatch`, `source_larger`, and `source_smaller` to the default tags removed when replacing posts.

nonamethanks:
Added resolution_mismatch + children and the missing_image/sample/thumbnail tags, since replacements for old posts were fixed and now those posts can be replaced again.

Fixes #4509.


evazion:
This will have to be updated if we decide to rename source_larger / source_smaller. https://danbooru.donmai.us/forum_topics/16995

PR #4518, opened 2020-06-17
Fix pixiv profile url

### LINKED ISSUE #4464: Change the Pixiv artist profile URL to the new format for artist creations
See [forum #16887](https://danbooru.donmai.us/forum_topics/16887). Basically, we should start using the new format, with the preferred one being the default. Many other users that are fixing up artist entries are already doing this, but there are still many that are still being created and left with the old format.

This **won't** be a mass-update of all artist entries at once to the new format, it'll just be changing the URLs provided to the user when clicking the "Create artist" button on the post page.

BrokenEagle:
Fixes #4464.

Additionally during testing, several of the illustration IDs were failing. This was because they were explict/guro, whereas the Pixiv account used in testing had those blocked. Therefore, these were separated out into their own separate block to draw notice to this in order to help other users do the test without getting those errors.


lllusion3469:
I'm not sure if it's intentional on your part to leave it like this but if anyone else is wondering why the artist search (which relies on the normalized profile url) is still working even for old entries where the `normalized_url` field was set when the old `member.php?id=` format was used: it's because of this line:
https://github.com/danbooru/danbooru/blob/d5a7eef53d9d521b9dd75e6f927a658ca5c61430/app/models/artist_url.rb#L31

Though, correct me if I'm wrong, if you wanted to change the format used for `artist_url.normalized_url`, you'd need to add a migration.

BrokenEagle:
> I'm not sure if it's intentional on your part to leave it like this...

Yeah it was intentional. All of the Pixiv URLs need to be standardized to only one normalization so that users have a dependable method to lookup Pixiv URLs.

> Though, correct me if I'm wrong, if you wanted to change the format used for `artist_url.normalized_url`, you'd need to add a migration.

Yeah, the artist URLs would have to be updated, as well as all of the artist version records, which I'm not sure @evazion is willing to do at this point.

PR #4547, opened 2020-07-11
Uploads: Check if strategy is enabled before use

### LINKED ISSUE #4545: Show a "not implemented" message instead of showing an exception on source download methods
When trying to e.g. upload a picture using the "URL" field, if Danbooru is not configured with the proper informations to "log in" the provided address, an error message showing the exception and the stack trace is shown.
It would be better to add a "this feature is not implemented" page when the relevant settings are nil (or another "falsey" value) and one tries to use said feature.

It would probably be best to have this fine-grained for each website API, so that the message doesn't appear if the site using Danbooru has, for example, configured pixiv but not DeviantArt.

nonamethanks:
Avoid returning bare API traceback from pixiv et al when login details are not configured, and instead raise a generic error with the site's name.

There already was a skeleton implementation for this so I just added the missing functions for all relevant strategies.

Fixes #4545.

### DIFF
--- a/app/logical/sources/strategies/deviant_art.rb
+++ b/app/logical/sources/strategies/deviant_art.rb
@@ -77,6 +77,10 @@ module Sources
 
       FAVME = %r{\Ahttps?://(?:www\.)?fav\.me/d(?<base36_deviation_id>[a-z0-9]+)\z}i
 
+      def self.enabled?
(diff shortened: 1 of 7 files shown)


nonamethanks:
Added a check in the preprocessor that is needed to catch replacements too.

PR #4553, opened 2020-07-19
Add attribute includes chaining on searches

### LINKED ISSUE #4539: Add search chaining on model includes
Discussed on Discord with @evazion. For instance, a wiki search could chain on the tag search to find wikis where the post count is greater than 50. This already exists to a degree, especially with the user includes. This would just be extending it to the rest of the includes where it makes sense.

BrokenEagle:
Fixes #4539.

Besides just adding all of the available includes to be searched, it also adds several other complementary includes and search methods. Dozens of tests were added to the functional tests to validate both regular and includes searching. The controller helper had to be updated in order to facilitate several needed aspects. More development notes are contained in the commit messages.


BrokenEagle:
I did several fixup based upon the previous output of the various checks.

BrokenEagle:
Added additional search attributes to artist versions: **urls**, **other_names**

app/models/wiki_page.rb
```diff
@@ -189,12 +189,12 @@ def merge_version
...
   def merge_version?
     prev = versions.last
-    prev && prev.updater_id == CurrentUser.user.id && prev.updated_at > 1.hour.ago
+    prev && prev.updater_id == CurrentUser.id && prev.updated_at > 1.hour.ago
   end
 
   def create_new_version
     versions.create(
-      :updater_id => CurrentUser.user.id,
+      :updater_id => CurrentUser.id,
```
evazion, line 197: This seems like a bandaid fix. `CurrentUser.id` will return `nil` if `CurrentUser` isn't set, which is still wrong.

test/test_helpers/controller_helper.rb
```diff
@@ -9,28 +9,37 @@ module ControllerHelper
...
-      search_params = { search: params }
+      search_params = other_params.merge({ search: params })
       expected_items = @test_case.instance_eval(&@expected)
 
       @test_case.instance_eval do
         # calls e.g. "wiki_pages_path" if we're in WikiPagesControllerTest.
         index_url = send("#{subject.controller_path}_path")
-        get index_url, as: :json, params: search_params
+        # Allows for different authorization levels to be used, instead of just anonymous
+        if CurrentUser.user.present?
```
evazion, line 29: The user should be passed explicitly instead of implicitly relying on `CurrentUser`.

app/models/wiki_page.rb
BrokenEagle: Yeah, it was mostly just to get the tests to work.

test/test_helpers/controller_helper.rb
BrokenEagle: Yeah, I hadn't really thought of that. Though that would also necessitate passing in the the user for every single shoulda call that are done by levels other than anonymous, which are all or most in some cases.

Issue #4548, opened 2020-07-12
Voting error responses are not rendering to the user


BrokenEagle:
Neither the note popup nor the console message are appearing. Instead, the "Updating posts (X pending)..." just hangs there indefinitely. I checked and it's sending back the right Javascript, however that Javascript is not running. I'm wondering if it might be because the response is an HTTP 500 response.

Systems tested:
- Chrome 83.0.4103.116 (64-bit)
- Firefox 77.0.1 (64-bit)

BrokenEagle:
I just checked my instance which is still running commit fde6e39 and it is doing the error responses just fine, i.e. both the Javascript is working and the response is also an HTTP 500, so it doesn't have anything to do with the HTTP code.

I noticed this problem just recently within the last week, and there are only about a week of new commits since my instance which seems to confirm that.

Commits difference: https://github.com/danbooru/danbooru/compare/fde6e39...1cef3bf
Code comparison: https://github.com/danbooru/danbooru/compare/fde6e39..1cef3bf

BrokenEagle:
Alright, I rebased the commits back and forth, and I was able to narrow down that it's commit 1760d1f which is causing the problem. With commit 
efe5111 where jQuery is bundled with  Webpack it still works, but on 1760d1f where jQuery was upgraded to 3.5.1 it no longer works. I don't know why that jQuery version is causing issues though...???

BrokenEagle:
https://blog.jquery.com/2020/04/10/jquery-3-5-0-released/

> One bug worth highlighting is a bug we fixed in the Ajax script transport. jQuery used to evaluate any response to a request for a script as a script, which is not always the desired behavior. This is different than other data types where such a convention was fine (e.g. in the case of JSON). **jQuery 3.5.0 will now only evaluate successful HTTP responses.**

So it looks like this was done intentionally in jQuery 3.5.0. This means that we can no longer send Javascript responses for errors and expect them to evaluate. If we want this to still work as before, we'll have to change errors to use a 2xx HTTP status code for network calls on all Javascript endpoints.

BrokenEagle:
Closing because votes through the tag script function no longer error out.

PR #4577, opened 2020-08-08
Add (partial) fanbox support

### LINKED ISSUE #3912: Fanbox: Batch bookmarklet uploads wrong image
ref: https://danbooru.donmai.us/forum_topics/15585

Using the batch bookmarklet on a Fanbox gallery ([link](http://danbooru/uploads/batch?url=https%3A%2F%2Fwww.pixiv.net%2Ffanbox%2Fcreator%2F5792350%2Fpost%2F129637)) and choosing the last image results in the first image being uploaded instead. `image_url` isn't returning the right image from the gallery.

### LINKED ISSUE #4316: Unable to replace images using pixiv fanbox URLs (status 403)
Attempting to replace post 3803729 with https://fanbox.pixiv.net/images/post/686294/jlkY26dFjpZCw0LbWJnSpcxZ.jpeg yielded "error: Downloads::File::Error - HTTP error code: 403 Forbidden"

I had to replace the image from File instead.

nonamethanks:
Fixes  #3912. Fixes #4316.

Unfortunately age-restricted posts are not possible to support, because they require login and that's hidden behind ReCaptcha v3 nowadays, which is impossible for us to work around.
Other projects like PixivUtil ask the user to submit their cookie, but that's not a viable strategy for us, so if we figure out how to work around the login to fetch the metadata it can always be updated.

I've opted to raise an error for R18 posts so that the user understands what's going on if the upload fails.

There's no `match?` for the Fanbox strategy despite the fact that it gets pixiv.net and piximg.net pictures, because those domains are already captured by the Pixiv strategy, and the only thing that filters down is urls with the "/fanbox/" url in it.

Cover and profile images are normalized into the artist's homepage when possible.

### DIFF
--- a/app/logical/sources/strategies/fanbox.rb
+++ b/app/logical/sources/strategies/fanbox.rb
@@ -1,0 +1,188 @@
+# Image URLs #############################################################
+#
+# * OLD DOMAIN
+# ** https://fanbox.pixiv.net/images/post/39714/JvjJal8v1yLgc5DPyEI05YpT.png
+#
+# * NEW DOMAIN
+# ** https://downloads.fanbox.cc/images/post/39714/JvjJal8v1yLgc5DPyEI05YpT.png (full res)
+# ** https://downloads.fanbox.cc/images/post/39714/c/1200x630/JvjJal8v1yLgc5DPyEI05YpT.jpeg (sample)
+# ** https://downloads.fanbox.cc/images/post/39714/w/1200/JvjJal8v1yLgc5DPyEI05YpT.jpeg (sample)
+#
+# * POST COVERS
+# * https://pixiv.pximg.net/c/1200x630_90_a2_g5/fanbox/public/images/post/186919/cover/VCI1Mcs2rbmWPg0mmiTisovn.jpeg
+#
+# * PROFILE IMAGES
+# * https://pixiv.pximg.net/c/400x400_90_a2_g5/fanbox/public/images/creator/1566167/profile/Ix6bnJmTaOAFZhXHLbWyIY1e.jpeg
+# * https://pixiv.pximg.net/fanbox/public/images/creator/1566167/profile/Ix6bnJmTaOAFZhXHLbWyIY1e.jpeg (dead URL type)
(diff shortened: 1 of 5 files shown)


app/logical/sources/strategies/fanbox.rb
```diff
@@ -0,0 +1,188 @@
...
+      end
+
+      def api_response
+        return {} if illust_id.blank?
+        resp = client.get("https://api.fanbox.cc/post.info?postId=#{illust_id}")
+        json_response = JSON.parse(resp)["body"]
+        if json_response["restrictedFor"] == 2 && json_response["body"].blank?
+          # Pixiv Fanbox login is protected by Google Recaptcha, so it's not possible for us to extract anything from them (save for the title).
+          # Other projects like PixivUtils ask the user to periodically extract cookies from the browser, but this is not feasible for Danbooru.
+          raise Sources::Error, "Age-restricted posts from Pixiv Fanbox are not supported."
```
evazion, line 166: Sources should never raise exceptions. This causes error messages to appear when you open the edit tab on a post and the source data is fetched.

app/logical/sources/strategies/fanbox.rb
```diff
@@ -0,0 +1,188 @@
...
+      end
+
+      def artist_api_response
+        # Needed to fetch artist from cover pages
+        return {} if artist_id_from_url.blank?
+        resp = client.get("https://api.fanbox.cc/creator.get?userId=#{artist_id_from_url}")
+        JSON.parse(resp)["body"]
+      rescue JSON::ParserError
+        {}
+      end
```
evazion, line 181: In what situation can parsing the json fail? If it's when a non-200 response is returned, it would be better to check for that case explicitly.

app/logical/sources/strategies/fanbox.rb
nonamethanks: If that was the case then it should be removed, however in my instance the error is not appearing, which is why I didn't notice that was the case.

I'm wondering if it's because of issue #4548.

app/logical/sources/strategies/fanbox.rb
nonamethanks: During my mass tests there have been a few times where the result was garbled HTML or badly formatted japanese, which is why it wasn't reliable to parse the JSON without a rescue. Other than that, there's never any other errors as long as the origin header is passed.

PR #4581, opened 2020-08-12
Autotag source_request & warn for sourceless uploads

### LINKED ISSUE #4569: Add source_request automatically
Similarly to how we automatically add tagme, and display warnings for no copyright and artist, source_request should be added in case of uploads with empty sources. Exceptions have to be made for cases like self-uploads and commissioner uploads.

nonamethanks:
Fixes #4569. 

Adds a warning similar to the others when someone uploads with no source and doesn't use self_upload/commissioner_upload.


nonamethanks:
Forgot I had these open. Closing for lack of interest

PR #4583, opened 2020-08-13
Safebooru: update restricted tags

### LINKED ISSUE #4582: Safebooru should not block "censored" tag / Update documentation
Recently we ran into a user confused by a safe image being blocked on Safebooru and determined it must be due to the site blocking the "censored" tag, as the post contained an identity_censor. This seems counterproductive as any censor that is explicit should be blocked by other attributes such as rating or always-explicit tags while censors are a common comedic or satiric device.

During this adventure, the only related wiki article I could find that might explain this was help:censored_tags, which is misleading in that it bluntly states that "the big three" (will GitHub get mad at those words?) are the only blocked tags. At first I felt that wiki should be updated, but it seems more prudent to make a help:safebooru wiki detailing the specific workings of the subdomain and pointing out the difference between it and safebooru.org. I would make this wiki page myself, but I'm clearly not actually sure of all the details of Safebooru's operation given I didn't know "censored" was blocked until now.

nonamethanks:
* remove "censored": this should be already handled by rating, as censored genitals, sexual acts etc should always be q when they're explicit enough that they're being censored. There's also a lot of false positives in keeping this in, such as identity censor, censored foods/innocuous items for comedy's sake, etc.

* add "sex_toy" - this is a recently populated tag since all types of sex toys now implicate it, and adding this also makes sure we cover the "sexual items" part of censored content that might be rating:safe but not something we want on safebooru, and would filter through if "censored" is not completely restricted

Fixes #4582

### DIFF
--- a/config/danbooru_default_config.rb
+++ b/config/danbooru_default_config.rb
@@ -258,7 +258,7 @@ module Danbooru
 
     # Tags that are not visible in safe mode.
     def safe_mode_restricted_tags
-      restricted_tags + %w[censored condom nipples nude penis pussy sexually_suggestive]
+      restricted_tags + %w[sex_toy condom nipples nude penis pussy sexually_suggestive]
     end
 
     # Tags that are only visible to Gold+ users.


nonamethanks:
Forgot I had these open. Closing for lack of interest

PR #4636, opened 2020-12-14
Uploads preprocessing: get rid of same-source existence check

### LINKED ISSUE #4409: Replacing corrupted image returns "source already exists"
post #3859608 was only uploaded about halfway, not rendering in full size and having half a grey thumbnail. Leaving the replacement fields with their default contents as follows:

> Replacement URL: https://www.pixiv.net/artworks/80737765
> Final Source: https://i.pximg.net/img-original/img/2020/04/12/16/22/58/80737765_p0.jpg

fails with an error akin to "post with source [url] already exists." Given one's attempting to replace an existing image that seems like a method that shouldn't cause a failure. In the end I replaced it by downloading the source image and blanking the Replacement URL, but that's definitely not a desired solution.

nonamethanks:
Fixes #4409. 

I really tried to understand why these lines are in the code but for the life of me I couldn't figure it out. The github blame function only goes back to 3 years ago when the preprocessor.rb file was created.

These lines are not needed to stop duplicate uploads, because the md5 check does the same. At first I thought they stopped duplicate preprocessing, but after some tests I realized those were being created regardless of the lines being there. 
The only thing they do from what I could tell is stop replacements if the same source already exists, which basically breaks replacements in many cases unless you empty the source field before replacing and fix it afterwards (which is slowly making me insane with all the e-hentai replacements being cockblocked by this). And this check doesn't really make sense for us because the same source can be applied to many different posts, since most sites we support have multiple images per url.

I also tried running tests but I can't seem to configure post versioning properly on my instance, every test that didn't rely on the missing services didn't change in output from what I could see though.

### DIFF
--- a/app/logical/upload_service/preprocessor.rb
+++ b/app/logical/upload_service/preprocessor.rb
@@ -66,20 +66,6 @@ class UploadService
     def start!
       raise NotImplementedError, "No login credentials configured for #{strategy.site_name}." unless strategy.class.enabled?
 
-      if Utils.is_downloadable?(source)
-        if Post.system_tag_match("source:#{canonical_source}").where.not(id: original_post_id).exists?
-          raise ActiveRecord::RecordNotUnique, "A post with source #{canonical_source} already exists"
-        end
-
-        if Upload.where(source: source, status: "completed").exists?
-          raise ActiveRecord::RecordNotUnique, "A completed upload with source #{source} already exists"
-        end
-
-        if Upload.where(source: source).where("status like ?", "error%").exists?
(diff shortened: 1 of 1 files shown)


evazion:
I'm not entirely sure this is safe to remove. I believe the intent here was to prevent duplicate uploads from being reprocessed. Without this check, if you open a post with the bookmarklet, the image will be downloaded, resized, and distributed to the image servers, even if the image already exists on Danbooru. Of course, the way it tries to check for duplicates by searching by source is pretty broken.

PR #4672, opened 2021-01-17
Add wiki page redirects

### LINKED ISSUE #4664: Add redirect capability to wikis
https://en.wikipedia.org/wiki/Wikipedia:Redirect

This would allow a lot of the "See [[blah]].'' wikis to be fixed so that they redirect to the actual tag.  It could also be done as a part of tag renames/aliases when the consequent wiki already exists. For artist redirects, it should probably go to the artist page instead of the artist wiki page.

If the redirector does need to be edited for any reason, it could still be entered through the edit interface. As such, "edit" links would probably need to be added to the wiki page index, the same that is currently being done on the artist index.

BrokenEagle:
Fixes #4664.

- As mentioned on Wikipedia, "The capitalization of the word REDIRECT doesn't matter."
    - https://en.wikipedia.org/wiki/Wikipedia:Redirect#Editing_the_source_directly
    - Likewise, the capitalization and use of underscores/spaces does not matter since the title gets normalized.
- Only 1 redirect is allowed to prevent infinite loops.
- The `redirect=false` parameter can be used to go to the wiki directly. 
- Added edit links to the index to allow for easier editing.
    - Plus this now matches the controls already being done for artists.

### DIFF
--- a/app/controllers/wiki_pages_controller.rb
+++ b/app/controllers/wiki_pages_controller.rb
@@ -42,7 +42,7 @@ class WikiPagesController < ApplicationController
       if @redirect.blank? || params["redirect"] == "false"
         respond_with(@wiki_page)
       else
-        redirect_to wiki_page_path(@redirect, redirect: false)
+        redirect_to wiki_page_path(@redirect, redirect: false, from: @wiki_page.title)
       end
     end
   end
--- a/app/controllers/wiki_pages_controller.rb
+++ b/app/controllers/wiki_pages_controller.rb
@@ -38,7 +38,12 @@ class WikiPagesController < ApplicationController
     elsif @wiki_page.blank?
       raise ActiveRecord::RecordNotFound
     else
-      respond_with(@wiki_page)
+      @redirect = @wiki_page.wiki_redirect
+      if @redirect.blank? || params["redirect"] == "false"
+        respond_with(@wiki_page)
+      else
+        redirect_to wiki_page_path(@redirect, redirect: false)
+      end
     end
   end
 
--- a/app/helpers/wiki_pages_helper.rb
+++ b/app/helpers/wiki_pages_helper.rb
@@ -3,4 +3,12 @@ module WikiPagesHelper
     names_html = wiki_page.other_names.map {|name| link_to(name, "http://www.pixiv.net/search.php?s_mode=s_tag_full&word=#{u(name)}", :class => "wiki-other-name")}
     names_html.join(" ").html_safe
   end
+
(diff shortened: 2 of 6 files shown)


BrokenEagle:
Got some feedback in Discord, so a `(Redirected from blah)` message was added like is done on Wikipedia.

Also added a redirect column on the index page so that these wikis are better highlighted. It gives 2 statuses depending on whether it has the correct format and whether the target wiki exists. This shouldn't add a whole lot of extra SQL queries since it only performs a query if the body format is correct.

nonamethanks:
I have to ask, wouldn't this be very easily abusable by pretty much anyone? From the point of view of someone not familiar with our wikis, if they stumble onto a malicious redirect they'll be led to believe it's an alias or something similar.

Also, usually those "see x" wikis get most traffic from the index page, since they're usually created to match missed searches. If the redirect works from there too it's an even more abusable scenario.

BrokenEagle:
> I have to ask, wouldn't this be very easily abusable by pretty much anyone? From the point of view of someone not familiar with our wikis, if they stumble onto a malicious redirect they'll be led to believe it's an alias or something similar.

The same could be said true of any malicious edit, and it should be handled just the same. By reverts, dmails, records, and then bans.

If it would be useful, then the name of the redirect could be listed on the wiki page index instead of just "yes". Additionally, a search option could be added which would allow a user to list all of the wiki pages with redirect style bodies, which would really be just a short cut for `search[body_regex]=^#REDIRECT \[\[([^\]]+)\]\]$`  or something like that.

> Also, usually those "see x" wikis get most traffic from the post search page, since they're usually created to match missed searches. If the redirect works from there too it's an even more abusable scenario.

It only activates via the `wiki_page#show` action, which does not include the `post#index` action.

> I can't really think of a case where an alias wouldn't be better in cases where it's applicable, and when it's not we have the problem that we have to constantly update the redirects when the tag becomes ambiguous (think of all the first-name-only ambiguous chartags created 10+ years ago that we still haven't fixed)

The problem already exists where wiki links are old and need to be updated. This will not fix that, nor is it even fixable since wiki bodies are unstructured data. It just adds an additional tool in the wiki page gardeners toolbox.

Also, aliases don't work for wiki page searches. Only active wiki pages show up, as aliases/renames just delete or move the old wiki name. However, this makes finding wikis under alternative/old names impossible.

Additionally, fixing old redirects would be easy because they have now been given structure, which makes parsing and replacing them relatively straightforward.

BrokenEagle:
Closing after 2 months of no interest.

If this isn't going to be done, then #4664 should be closed as well.

PR #4750, opened 2021-03-04
Change "Inviter" to "Promoter"

smowtenshi:
It's kind of old and misleading

### DIFF
--- a/app/views/users/_statistics.html.erb
+++ b/app/views/users/_statistics.html.erb
@@ -63,15 +63,6 @@
       <% end %>
 
       <tr>
-        <th>Promoter</th>
-        <% if user.inviter %>
-          <td><%= link_to_user user.inviter %> <%= link_to "»", users_path(search: { inviter: { name: user.inviter.name }}) %></td>
-        <% else %>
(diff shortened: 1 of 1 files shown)


luni3359:
I agree that it's misleading, but in my opinion the whole field should just go. It made sense for it to exist before, but when was the last time you saw a site that told you who changed your member status?

nonamethanks:
The field is there for accountability mostly, to immediately check who promoted someone, but admittedly it's just easier to check the feedbacks, since it doesn't show whether it was a promotion or a demotion and which privileges if any were added or removed. I agree it's a better idea to just remove it.

evazion:
At this point it's a legacy field that only exists to track old users who were promoted before promotions were tracked in mod actions.

evazion:
I'm going to add this field back. Even after clearing the inviter for all Gold/Plat self-upgrades, and for all users who have a feedback or a modaction listing who promoted them, there are still about 600 users for whom the inviter field is the only record of who promoted them.

luni3359:
Could this at least become a moderator-only field? Do normal users need to see it?

Issue #4741, opened 2021-02-28
Note Control Enhancements


BrokenEagle:
### Hotkeys

It would be useful to have a few additional hotkeys when a note is in the selected state (blue border).

1. Open the edit menu

    Because it is very difficult sometimes to click the popup note for embedded notes to access the edit menu.

2. Save the note

    Because sometimes only the note is resized/repositioned, and the extra work to open the edit menu just to save it makes the work of doing that that much harder/unappealing, especially when they have the issue mentioned above for 1.

3. Deselect the note

   Because sometimes when clicking the note, it accidentally gets dragged a bit. In the case of rotated notes, this can cause the element to jump quite a bit.

### Embedded notes attributes

Besides the additional hotkeys, it would be useful to have the `z-index` attribute be available for embedded notes via the `.note-box-attributes`. Right now, there is no way to easily control which element gets stacked first to last (it's based upon the order that notes get inserted into the DOM), and sometimes it's unavoidable to have note boxes overlap.

BrokenEagle:
Not sure there is much of a desire for hotkeys, plus I've already created a userscript which adds those in for myself at least. If there's no further comment, I'll close this in a few days.

PR #4763, opened 2021-03-13
Add ability to determine stack order of embedded notes

BrokenEagle:
Partial fix of #4741.

This will allow for the stack order of embedded notes to be determined via the user. Currently, the order is determined by the order they get rendered, which makes it hard to precisely work with embedded notes, especially since overlap is sometimes impossible to avoid.

### DIFF
--- a/app/javascript/src/javascripts/notes.js
+++ b/app/javascript/src/javascripts/notes.js
@@ -10,6 +10,7 @@ class Note {
   static HIDE_DELAY = 250;
   static NORMALIZE_ATTRIBUTES = ['letter-spacing', 'line-height', 'margin-left', 'margin-right', 'margin-top', 'margin-bottom', 'padding-left', 'padding-right', 'padding-top', 'padding-bottom'];
   static COPY_ATTRIBUTES = ['background-color', 'border-radius', 'transform', 'justify-content', 'align-items'];
+  static BOX_LEVEL_CLASSES = Array.from({length: 5}, (_, i) => 'level-' + (i + 1));
   static RESIZE_HANDLES = "se, nw";
 
   // Notes must be at least 10x10 in size so they're big enough to drag and resize.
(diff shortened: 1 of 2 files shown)


app/javascript/src/javascripts/notes.js
```diff
@@ -134,7 +134,9 @@ class Note {
 
     copy_style_attributes() {
       let $note_box = this.$note_box;
-      let $attribute_child = $note_box.find('.note-box-attributes');
+      let attributes_body = this.note.preview_body || "";
+      let $attributes_note = $('<div>' + attributes_body + '</div>');
```
evazion, line 138: This contains an XSS bug. If you put `` in an embedded note then it will execute malicious Javascript for anyone viewing the post.

Not sure what `preview_body` is for but it should come from the sanitized HTML returned by the `/note_previews` call.

evazion:
Allowing full z-index control is too much freedom. Setting it to a negative value allows you to place notes beneath the image where they're impossible to see or edit. Setting it too high allows you to place notes above dialog boxes.

I think only two levels are needed, high and low. You should say something like `<div class="note-box-attributes high"></div>` to put a note on the high level or `class="... low` to put it on the low level, where `.high` and `.low` are CSS classes that map to some preset z-index level. This is enough to separate two overlapping notes; you shouldn't have more notes than that overlapping.

app/javascript/src/javascripts/notes.js
BrokenEagle: `/note_previews` currently strips out the z-index attribute. I think I can probably use a regex to safely harvest the z-index attribute though instead.

BrokenEagle:
Yeah, I was thinking that there would need to be some range limitation yesterday, but I don't think 2 values are going to cut it.

image

The above scenario is a bit contrived, but I can easily imagine there being 3 or more closely packed notes and the order needs to be set precisely.

It wouldn't be all that difficult to limit the range of numbers from something like 1-10, and then perhaps be adjusted in the future if there's the scenario where there's a post with more than 10 boxes need to be precisely overlapped.

BrokenEagle:
Oh wait, nevermind. If it's only going to be a small amount of values, then it'll be much easier to just create a few extra CSS rules that'll accommodate several levels. I'll start with a low number like 5.

BrokenEagle:
I went with a CSS/Javascript solution as the impression was that seemed safer, although if desired a Javascript-only solution can be done by setting the z-index directly on the note box itself.

Also, I cleared the way for a constant to be used, such as something defined in the Notes module, instead of a hardcoded 5. I wasn't sure if it was desired though, so I left it hardcoded in. Changing it would still be relatively easy, needing to only change it in two places.

PR #4772, opened 2021-03-19
Tag Aliases: automatically remove redundant implications

nonamethanks:
Ref: https://danbooru.donmai.us/forum_topics/18277, https://danbooru.donmai.us/forum_topics/18278

When aliasing a to b, remove an eventual implication from a to b as well.

Not sure if there's a simple way to get both implication directions in a single line, I couldn't find any way without having to use raw SQL.

### DIFF
--- a/app/models/tag_alias.rb
+++ b/app/models/tag_alias.rb
@@ -2,7 +2,7 @@ class TagAlias < TagRelationship
   validates_uniqueness_of :antecedent_name, scope: :status, conditions: -> { active }
   validate :absence_of_transitive_relation
 
-  before_create :delete_conflicting_alias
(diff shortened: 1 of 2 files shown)


evazion:
I didn't fully understand this at first. To explain it:

* If A implies B, but then we alias A to B, or B to A, then we need to remove the A -> B implication first.
* If A implies C, and we alias A to B, then the implication will be updated to B implies C elsewhere in [TagMover.move_implications!](https://github.com/danbooru/danbooru/blob/857776db9ee9a9adb6cb2d6f837baffb1ff8d7b4/app/logical/tag_mover.rb#L71-L79)

PR #4801, opened 2021-05-19
Add Lofter support

nonamethanks:
Cuz replacing posts manually when users upload samples is a pain in the ass, since Lofter blocks the right click menu.


nonamethanks:
Looking at the codecov test results it seems most of it is bogus. It's like it didn't find a single one of the tests I added.

PR #4820, opened 2021-06-19
Search: fix unmatched wildcards

### LINKED ISSUE #4435: Search: wildcards with no matches should return no results
https://danbooru.donmai.us/posts?tags=*qiojqoiwg* should return nothing instead of everything.

Thayol:
Fixes #4435 by using the wildcard itself in queries where it would not have matched anything.

I am not familiar with Ruby, so this may need some refactoring. If so, please lend me a helping hand.

## Behavior before

If wildcards did not match anything, they were completely ignored. If a search was a wildcard search and it did not match anything, it was like having an empty search box which then returned everything...

| User input | Real query | Result |
| :--- | :--- | :--- |
| \*non-existent\* |  | **Posts of the index page** |
| soranamae \*non-existent\* | soranamae | **Posts of soranamae** |
| serval_\* | ~serval_girl ~serval_print ~serval_tail | *Posts of serval_\** |
| takom serval_\* | takom ~serval_girl ~serval_print ~serval_tail | *Posts of serval_\* by takom* |

## Behavior after

Wildcards add their wildcard form to the search if they are not matched. Wildcard forms cannot be present in the system, which will inevitably lead to an empty search result.

| User input | Real query | Result |
| :--- | :--- | :--- |
| \*non-existent\* | \*non-existent\* | **Nobody here but us chickens!** |
| soranamae \*non-existent\* | soranamae \*non-existent\* | **Nobody here but us chickens!** |
| serval_\* | ~serval_girl ~serval_print ~serval_tail | *Posts of serval_\** |
| takom serval_\* | takom ~serval_girl ~serval_print ~serval_tail | *Posts of serval_\* by takom* |

*Note: Matched examples are cut down to have readable tables. Important behavior change marked by* ***bold*** *text.*

### DIFF
--- a/app/logical/post_query_builder.rb
+++ b/app/logical/post_query_builder.rb
@@ -82,11 +82,15 @@ class PostQueryBuilder
     negated_tags = negated_tags.map(&:name)
     optional_tags = optional_tags.map(&:name)
     required_tags = required_tags.map(&:name)
-
-    negated_tags += negated_wildcard_tags.flat_map { |tag| Tag.wildcard_matches(tag.name).limit(MAX_WILDCARD_TAGS).pluck(:name) }
-    optional_tags += optional_wildcard_tags.flat_map { |tag| Tag.wildcard_matches(tag.name).limit(MAX_WILDCARD_TAGS).pluck(:name) }
-    optional_tags += required_wildcard_tags.flat_map { |tag| Tag.wildcard_matches(tag.name).limit(MAX_WILDCARD_TAGS).pluck(:name) }
-
+    
+    matched_negated_wildcard_tags = negated_wildcard_tags.flat_map { |tag| Tag.wildcard_matches(tag.name).limit(MAX_WILDCARD_TAGS).pluck(:name) }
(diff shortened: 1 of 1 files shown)


Thayol:
In the [PostQueryBuilder tests](https://github.com/danbooru/danbooru/blob/7de862deb2594b16e7c2507a7a1ca8eb21edc69b/test/unit/post_query_builder_test.rb) I see `assert_tag_match` but no negative form. I believe this is the reason this bug existed, however I have no clue how to make one. If I had something like `assert_tag_no_match` I could go over each test case and add this to them or create completely new ones dedicated to testing this functionality.

nonamethanks:
You can just define a function that does the opposite, with assert_not_equal

evazion:
> In the [PostQueryBuilder tests](https://github.com/danbooru/danbooru/blob/7de862deb2594b16e7c2507a7a1ca8eb21edc69b/test/unit/post_query_builder_test.rb) I see `assert_tag_match` but no negative form. I believe this is the reason this bug existed, however I have no clue how to make one. If I had something like `assert_tag_no_match` I could go over each test case and add this to them or create completely new ones dedicated to testing this functionality.

You can use `assert_tag_match([], "*blah*")` to assert that `*blah*` returns the empty set.

PR #4824, opened 2021-06-21
Posts sidebar: add link to tag discussions in forums

nonamethanks:
Just like it's done in wikis. I often find myself in need of this link when browsing posts.
image


evazion:
"Discussions" could be comments or forum posts.

nonamethanks:
The wiki link also says "discussions":https://github.com/danbooru/danbooru/blob/f78d10a5911e632cf92e675e6af7fe42aca2232c/app/views/wiki_pages/_sidebar.html.erb#L13 so I simply opted to make it the same for consistency. "Forum discussions" was pretty long compared to the rest of the entries so imo it looked out of place.

luni3359:
Leaving it as "Comments" feels like it would convey the wrong meaning, I would expect comments from people casually talking about the tag itself rather than talks about what to do with it.

PR #4837, opened 2021-07-12
Fix skeb getting the wrong picture, again

nonamethanks:
Skeb changed something, again, so we have to exclude more samples.


nonamethanks:
I'm closing this for now because I need to think more about it. I just discovered that this bug actually allows us to get unwatermarked, slightly smaller pictures from skeb. See https://danbooru.donmai.us/posts/4642023 vs the source for example. 

In the meanwhile i'll keep manually replacing the ones that have a bigger unwatermarked version on skeb.

PR #4842, opened 2021-07-21
Post flags: allow users to search for their own flags on their own posts

nonamethanks:
I couldn't figure out how to write tests for this because the tests from here https://github.com/danbooru/danbooru/blob/master/test/functional/post_flags_controller_test.rb#L68 on forward seem to be about the flags page, not the post search page.


nonamethanks:
Closing for lack of interest

PR #4845, opened 2021-07-22
Uploads: prevent submission of unfinished uploads

### LINKED ISSUE #4608: Block uploads from disk until file is fully uploaded
If you submit a file before it's finished uploading, you get this:

The submit button should just return an error similar to the lack of rating rather than throw away the tags and whatever percentage was uploaded.

nonamethanks:
Fixes #4608 (though not the underlying issue, though there's already another issue for that).

Also fixes that you could submit an empty upload anyway because `$("#upload_file").val() ` is undefined, not an empty string.

I added a .+ to the source regex for consistency because if you submit with just "http://", which technically fullfilled the previous one, it still shows you the no source error in the error page.

### DIFF
--- a/app/components/source_data_component/source_data_component.html.erb
+++ b/app/components/source_data_component/source_data_component.html.erb
@@ -5,20 +5,7 @@
   <% if @source.present? %>
     <table class="source-data-content mt-2">
       <tbody>
-        <% if @source.artist_name.blank? %>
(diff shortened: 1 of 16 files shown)


evazion:
Merged in 8772bc78ecf3d61452019def4a70b6da2f32168c.

PR #4850, opened 2021-07-23
Exclude deleted pools in index by default

### LINKED ISSUE #4601: Hide deleted pools by default in pool search
And add a control to choose whether deleted pools should be visible (default should be off).

Example of what we really don't want to show to new users (nsfw): https://i.imgur.com/TRWNSWo.png

nonamethanks:
Fixes #4601.

### DIFF
--- a/app/models/pool.rb
+++ b/app/models/pool.rb
@@ -65,6 +65,10 @@ class Pool < ApplicationRecord
         q = q.apply_default_order(params)
       end
 
+      unless params[:is_deleted].present?
+        q = q.active
+      end
+
       q
     end
   end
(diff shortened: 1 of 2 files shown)


evazion:
This would be a breaking change to the API.

nonamethanks:
Hm I guess it is, but I can't really imagine the average api user would usually want to get deleted pools as well as active ones, given that they're always either personal pools or old pools moved to tags.

We don't display deleted forum topics by default either.

nonamethanks:
Made redundant by 2eb89a8354161bc47230495a539e163b45539053

evazion:
Fixed in 2eb89a8354161bc47230495a539e163b45539053. Putting default filters in the `search` in models is bad for a few reasons:

* There's no way to disable the filter if you don't want it. You can do `search[is_deleted]=true` and `search[is_deleted]=false`, but you can't get both deleted and undeleted pools at the same time if the default is to filter out deleted pools.
* Most other endpoints like `/posts.json` don't filter out deleted posts by default, so the expectation is that the /pools endpoint won't either.
* `search` methods in models are supposed to be generic search methods that just search, and don't apply any unasked-for filtering. Sometimes they're used in other contexts of a search performed by a user.

PR #4855, opened 2021-08-01
Log mod actions for changes to post locks

nottalulah:
Since pull #4504 was rejected, issue #4395 was left open. This is a simpler fix to track locking/unlocking of a post's rating, status, and notes, but it doesn't do anything for tracking embedded notes.


evazion:
I've been holding off on this because I plan to remove these locks eventually, and I hate to add this only to have to remove it later.

PR #4885, opened 2021-09-22
Bump Ruby to 3.0.2 in INSTALL.debian

GlassedSilver:


GlassedSilver:
Correct me if I'm wrong, but shouldn't `/bin/bundle`'s line 76 also refer to 3.0.2?

evazion:
Merged manually since there are a bunch of unrelated commits here.

> Correct me if I'm wrong, but shouldn't /bin/bundle's line 76 also refer to 3.0.2?

I believe this is the [RubyGems](https://github.com/rubygems/rubygems) version, which is different than the Ruby version.

GlassedSilver:
> 
> 
> Merged manually since there are a bunch of unrelated commits here.
> 
> > Correct me if I'm wrong, but shouldn't /bin/bundle's line 76 also refer to 3.0.2?
> 
> I believe this is the [RubyGems](https://github.com/rubygems/rubygems) version, which is different than the Ruby version.

Yeah sorry for all the messy commits there that are in between.

PR #4911, opened 2021-11-01
Add Plurk support

nonamethanks:
Adds support for publicly visible plurks. The adult ones require oauth authentication.


nonamethanks:
NVM on the adult thing, I actually found how to go around the adult restrictions without doing any kind of authentication.

Issue #4640, opened 2020-12-18
Tool to undo comment vote spamming


nonamethanks:
There should be a way to mass revert an user's comment votes. Maybe in the ban page, because look at this:
https://danbooru.donmai.us/comment_votes?commit=Search&search%5Buser_name%5D=NoBlueLives
https://danbooru.donmai.us/comment_votes?commit=Search&search%5Buser_name%5D=Cock%26Ball_Destroyer
https://danbooru.donmai.us/comment_votes?commit=Search&search%5Buser_name%5D=Admiral_Fagg0tPants

All same user using sockpuppets. There's more. Basically someone can keep creating new accounts and mass downvoting stuff into oblivion and there's no way to revert this, because all it takes is a member account to do so.

evazion:
This is how it works:

* Comment votes now have an `is_deleted` flag.
* Comment votes are soft deleted instead of hard deleted. This includes when a user removes their own vote.
* Admins can delete comment votes by other users. This is done by going to the comment vote list and clicking the Remove button on each one.
* Deleting a vote leaves a mod action.

I'm wary about giving this to moderators because I don't want mods deleting votes unless it's to undo mass spam by sockpuppet accounts. Imagine you were on Reddit or some other site and you knew moderators could silently manipulate comment scores by removing votes if people voted the "wrong" way. It would reduce trust in the site, even if it weren't being abused.

evazion:
Allowing mods to remove all of someone's votes when they permaban someone might be acceptable. That's harder to misuse than unrestricted vote removal. I'll leave this issue open for that.

nonamethanks:
Yeah, allowing it to be done only on account ban is definitely an acceptable restriction.

ghost:
Was this fixed with b5ba14bb04a919efa1143486221108e4110165dd or is something additional required?

PR #4923, opened 2021-11-18
Allow Members to vote

evazion:
- [x] Allow Members to vote.
- [x] Make upvotes public by default, the same way favorites are public.
- [x] Change the "Private favorites" account setting to "Private favorites and upvotes".
- [x] Make private favorites, upvotes, and favgroups Gold-only features. Existing Members with private favorites or favgroups can keep them, but if they disable privacy then they will need a Gold account to re-enable it.
- [x] Add score/upvote/downvote buttons beneath thumbnails (optional; off by default).
- [x] Make it so you can click or hover over a post's score to see the list of upvoters, and the upvote-to-downvote ratio. Only public upvotes are visible; downvotes and private upvotes are shown as "hidden".
- [x] Remove the ability for banned users and restricted users to favorite posts. This is so banned or restricted users can't upvote posts using favorites.
- [x] Make upvote/downvote buttons visible to logged-out users.
- [x] Make it so the list of favorites is shown in the same way as the list of upvotes.
- [ ] Make it so when a post is heavily downvoted, it's automatically hidden, similar to blacklisting.
- [ ] Make it so when you downvote a post, it's automatically hidden for you, similar to blacklisting.
- [ ] Make it so when you upvote a post, it's automatically visible for you, even if it gets deleted.
- [ ] Add a way to easily blacklist frequently-blacklisted tags.


evazion:
Screenshots:

image

image

nonamethanks:
Just a reminder, admins need to be able to delete post votes like with comment votes, to combat sockpuppets.

nottalulah:
Might be a good time to tackle issue #4640 as well.

BrokenEagle:
Just my opinion, but the thumbnail score/vote UI should be standard aligned to the bottom of the thumbnail box, not the bottom of the thumbnail. That way, all of the controls will line up in a row. If needed, include a border around the thumbnail box so that it's clear which controls go to which thumbnail.

What is the DB timeout for getting post votes on the highest voted post? (3K+, [post # 3410029](https://danbooru.donmai.us/posts/3410029?), NSFW) Also, consider a more extreme case when everyone can vote, like 10K+, 50K+, and perhaps even 100K+. If these are likely to timeout, could the tooltip code just fetch like 100 entries at a time, and do something like infinite scrolling instead?

> Make it so when you downvote a post, it's automatically hidden for you, similar to blacklisting.
> Make it so when you upvote a post, it's automatically visible for you, even if it gets deleted.

Will this be done with Javascript, like additional steps to the blacklist code? Because otherwise it'll introduce an additional time penalty for DB queries, which will make things more likely to timeout for users. If it's the latter, can this be made optional? I'd rather choose for myself whether I want to add the `upvote:BrokenEagle98` and `downvote:BrokenEagle98` metatags myself.

> Add a way to easily blacklist frequently-blacklisted tags.

Will this be intelligently done, or will it just append tags as new entries on their own lines at the end of the blacklist? Because right now, every line in the blacklist requires O(n) time to process. In my own blacklist, I've been able to combine over a dozen tags on some of the lines, which drastically reduces the processing time. Or perhaps, will a separate blacklist be maintained, and it pretty much will act like a single giant OR line to the old blacklist?

evazion:
Merging and deploying this to https://betabooru.donmai.us for now. Remaining issues will be addressed in future work.

>Just my opinion, but the thumbnail score/vote UI should be standard aligned to the bottom of the thumbnail box, not the bottom of the thumbnail. That way, all of the controls will line up in a row. If needed, include a border around the thumbnail box so that it's clear which controls go to which thumbnail.

This leads to big uneven gaps between the vote UI and the thumbnail, especially for wide images. This tends to look off in my opinion when there are irregular gaps between the UI and the thumbnail.

Most sites deal with this by cropping thumbnails to a fixed size, that way the UI always lines up. It's basically a choice between cropped thumbnails and perfect UI alignment, or uncropped thumbnails and imperfect UI alignment.

>What is the DB timeout for getting post votes on the highest voted post? (3K+, post # 3410029, NSFW) Also, consider a more extreme case when everyone can vote, like 10K+, 50K+, and perhaps even 100K+. If these are likely to timeout, could the tooltip code just fetch like 100 entries at a time, and do something like infinite scrolling instead?

Infinite scrolling is something we can do in the future. We already load the full list of favorites for the sidebar, even when the user doesn't look at it, which is even worse than loading the vote list in a tooltip. One of the changes here is to load favorites in a tooltip, that way we're at least not loading it in the initial pageload.

>Will this be done with Javascript, like additional steps to the blacklist code? Because otherwise it'll introduce an additional time penalty for DB queries, which will make things more likely to timeout for users. If it's the latter, can this be made optional? I'd rather choose for myself whether I want to add the upvote:BrokenEagle98 and downvote:BrokenEagle98 metatags myself.

Blacklists will probably have to go serverside, for various reasons, including this. Thumbnails would be given a `blacklisted`  class serverside, depending on whether they match one of your blacklist rules or not, then they can be hidden clientside in CSS.

We're already loading whether you've upvoted or downvoted the post, in order to render the vote buttons beneath the thumbnail correctly, so there's no extra penalty for showing or hiding posts based on your votes.

>Will this be intelligently done, or will it just append tags as new entries on their own lines at the end of the blacklist? Because right now, every line in the blacklist requires O(n) time to process. In my own blacklist, I've been able to combine over a dozen tags on some of the lines, which drastically reduces the processing time. Or perhaps, will a separate blacklist be maintained, and it pretty much will act like a single giant OR line to the old blacklist?

Handling blacklists serverside with a dedicated model for them will make this easier. Currently it's too hard to correctly add/remove tags to a user's blacklist without a bunch of messy Javascript.

BrokenEagle:
Sounds good. Thanks for the reply. I hope the extra work for blacklist processing doesn't put too much additional strain on the server though.

Would turning on/off individual lines of a blacklist still be supported, or should we get rid of that? One benefit of no longer supporting on/off for individual lines is that you can early bail on posts which already have a line match, so that the remaining lines don't have to be checked. Lines could also be optimized by ordering them by the preponderance of post matches on Danbooru, putting the lines most likely to have a match first. This ordering could be done when the blacklist gets updated, and perhaps once every day/week/month if no changes have been done within that period. Removing the on/off for individual lines would also simplify the UI, necessitating only a single control to turn the entire blacklist on or off. For myself at least, I cannot recall the last time I've actually used the individual line controls. Anyways, just some food for thought.

PR #4925, opened 2021-11-19
mod dashboard: minor cleanup

### LINKED ISSUE #4924: Moderator dashboard shouldn't show a "Promote" link that's unusable by non-moderators
At the very least it could be greyed out.

nottalulah:
* only show the [+] and [-] feedback buttons to gold users
* only show the [promote] button to moderators, and only when targetting users
  builder or below
* don't show either button if the target is the current user
* closes #4924

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -210,14 +210,16 @@ module ApplicationHelper
     html = ""
     html << link_to_user(user)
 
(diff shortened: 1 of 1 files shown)


evazion:
I'll just remove these completely. Mods can also access the "give feedback" or "promote user" options by hovering over a person's username and using the "..." menu in the tooltip.

PR #4950, opened 2022-01-03
Always use downscaled 720x720 thumbnails for gigantic post previews.

### LINKED ISSUE #4944: Use downscaled 720x720 thumbnails for gigantic post previews
Large, huge, and gigantic post previews all use 360x360 thumbnails by default, and 720x720 thumbnails when your device has a high pixel density ratio. Gigantic post previews should use 720x720 thumbnails by default, regardless of pixel density.

SuzakuMedli:
Use downscaled 720x720 thumbnails by default when viewing gigantic post previews, instead of only when the user has a high pixel density screen. Fixes #4944.

### DIFF
--- a/app/components/post_preview_component/post_preview_component.html.erb
+++ b/app/components/post_preview_component/post_preview_component.html.erb
@@ -19,8 +19,10 @@
           <% case size %>
           <% when "150", "180" %>
             <%= tag.source type: "image/jpeg", srcset: "#{media_asset.variant("180x180").file_url} 1x, #{media_asset.variant("360x360").file_url} 2x" %>
-          <% when "225", "225w", "270", "270w", "360" %>
+          <% when "225", "225w", "270", "270w" %>
             <%= tag.source type: "image/webp", srcset: "#{media_asset.variant("360x360").file_url} 1x, #{media_asset.variant("720x720").file_url} 2x" %>
+          <% when "360" %>
+            <%= tag.source type: "image/webp", srcset: "#{media_asset.variant("720x720").file_url} 1x" %>
           <% end %>
         <% end %>
 


BrokenEagle:
I imagine that if some users specify 360px, that they really mean 360px, and since we have a 360px, we should probably send a 360px. What I mean is, larger images require sending more information. This could cause issues with page loading latency, and some users could have limited data downloading availability. I don't know exactly know how 2x works, but I would hope that that option could be turned off as well for those that are data constrained. For myself at least, it'll irk me if I ever have a need to use the 360 size and see any kind of latency whatsoever, because I'll know that it didn't have to be that way.

SuzakuMedli:
Mobile users are already shown downscaled 720x720 thumbnails if they have selected large/huge post previews, since smartphones have high pixel density screens.
This change should only affect desktop users because, as far as I know, gigantic post previews aren't available to mobile users.

BrokenEagle:
Just use CSS rules then instead of forcing it like I mentioned above. I created some CSS rules which will allow one to do this.

```css
#c-posts #a-index .post-gallery-720 img {
    max-height: 360px;
}
#c-posts #a-index .post-gallery-720 .post-preview-container {
    height: 360px;
}
#c-posts #a-index .post-gallery-grid.post-gallery-720 .posts-container {
    grid-template-columns: repeat(auto-fill, minmax(360px, 1fr));
}
```
I also posted the above CSS onto the forums ([forum #203483](https://danbooru.donmai.us/forum_posts/203483)).

SuzakuMedli:
When you view small post previews on mobile it currently uses the same 360x360 thumbnails as when you view gigantic post previews on desktop. Something needs to be changed one way or another, because that just doesn't seem right.

The whole point of this is to make gigantic post previews look better for everyone, not just the people that use that chunk of custom CSS in conjunction with the `size=720` URL parameter.

BrokenEagle:
What you desire is something that only a portion of people would want, whereas I have no desire for that myself, and I cannot imagine that I'm the only one. Why should we be forced to get the 720px size when we really want 360px? Also, and this is my personal opinion, but the way that browsers resize images always looks like garbage compared to what actual image manipulation software can do.

Besides the point above,, there are a ton of things that the site won't do for users, and requires them to either use custom CSS or userscripts to fix the site to suit their own needs, and not changing things unnecessarily for every single user.

Also, for reference, one can also use different CSS for mobile and for desktop. Lookup `@media`.

SuzakuMedli:
I don't think that the people that select gigantic post previews "really want" 360x360, because they probably don't know exactly what size what they want, they just know that they want _gigantic_ post previews. I'm willing to bet that most desktop users, myself included, expect to see the highest quality thumbnails available when viewing the largest post preview size they can select.

Regarding resizing, large and huge post previews have been using 360x360 thumbnails downscaled to 225 and 270 respectively, and I'm pretty sure nobody has complained about them looking bad.

At the end of the day, the disparity in perceived quality between huge and gigantic post previews is the reason I ended up creating this pull request; huge looks fine, but gigantic feels like it should look better.

nonamethanks:
This has become obsolete over the months, since gigantic and absurd thumbnails use x720 webp files now.

PR #4952, opened 2022-01-04
Search: "Fix" negated ordered metatags

### LINKED ISSUE #4918: Search: -ordfav:user is broken
https://danbooru.donmai.us/posts?tags=-ordfav:evazion

```
ERROR:  missing FROM-clause entry for table "favorites" at character 36
STATEMENT:  SELECT COUNT(*) FROM "posts" WHERE "favorites"."user_id" != $1
```

thayol:
Fixes #4918 by guessing that the user wanted to search for the non "ord" version of the negated tags.

`-ordfav:USER` → `-fav:USER`
`-ordfavgroup:ID` → `-favgroup:ID`
`-ordpool:ID` → `-pool:ID`

**Note:** It doesn't fix `-order:rank` and other negated `order` searches.

### DIFF
--- a/app/logical/post_query_builder.rb
+++ b/app/logical/post_query_builder.rb
@@ -118,9 +118,13 @@ class PostQueryBuilder
 
   def metatags_match(metatags, relation)
     metatags.each do |metatag|
-      metatag.name = metatags_without_ord[metatag.name] if metatag.negated && metatags_without_ord.key?(metatag.name)
+      metatag_name = if metatag.negated && metatags_without_ord.key?(metatag.name)
+        metatags_without_ord[metatag.name]
+      else
+        metatag.name
+      end
 
-      clause = metatag_matches(metatag.name, metatag.value, quoted: metatag.quoted)
+      clause = metatag_matches(metatag_name, metatag.value, quoted: metatag.quoted)
       clause = clause.negate_relation if metatag.negated
       relation = relation.and_relation(clause)
     end
--- a/app/logical/post_query_builder.rb
+++ b/app/logical/post_query_builder.rb
@@ -117,24 +117,19 @@ class PostQueryBuilder
   end
 
   def metatags_match(metatags, relation)
-    fix_negated_ordered_metatags(metatags)
-    metatags.each do |metatag|
-      clause = metatag_matches(metatag.name, metatag.value, quoted: metatag.quoted)
-      clause = clause.negate_relation if metatag.negated
-      relation = relation.and_relation(clause)
-    end
-
-    relation
-  end
-  
-  def fix_negated_ordered_metatags(metatag)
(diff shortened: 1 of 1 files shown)


app/logical/post_query_builder.rb
```diff
@@ -118,6 +118,12 @@ def tags_match(tags, relation)
...
   def metatags_match(metatags, relation)
     metatags.each do |metatag|
+      if metatag.negated
+        metatag.name = "fav" if metatag.name == "ordfav"
+        metatag.name = "favgroup" if metatag.name == "ordfavgroup"
+        metatag.name = "pool" if metatag.name == "ordpool"
+      end
+
       clause = metatag_matches(metatag.name, metatag.value, quoted: metatag.quoted)
       clause = clause.negate_relation if metatag.negated
```
Thayol, line 128: Right now, `negate_relation` does `unscoped.where(all.where_clause.invert.ast)` (From [searchable.rb](https://github.com/danbooru/danbooru/blob/2fff56302705bc0e5ee723c72a851c51dbf150d4/app/logical/concerns/searchable.rb#L16))
I don't know the compatibility with other queries or how taxing it would be to change it but `unscoped.where.not(id: all.pluck(:id))` sounds good in my head.
Is this something to consider?

Thayol: Funny thing is that it works without explicitly `pluck`ing the IDs.

Thayol: See https://discord.com/channels/310432830138089472/310846683376517121/934147187175194674 and the messages after it. Summary: this creates a chunky query that cucks all other order statements so replacing the metatags is preferable.

thayol:
If this wasn't accepted because of styling issues, I've made it look nicer. Now it's a hash that's easy to extend and doesn't clutter up the method `metatags_match`.

evazion:
The issue is that modifying `metatag.name` in the middle of building the query is unsafe. Consider this:

```
$ bin/rails console
pry(main)> query = PostQueryBuilder.new("-ordfav:evazion")
pry(main)> query.has_metatag?("ordfav")
=> true
pry(main)> query.build; nil
=> nil
pry(main)> query.has_metatag?("ordfav")
=> false
pry(main)> query.metatags
=> [#<OpenStruct type=:metatag, name="fav", value="evazion", negated=true, quoted=false>]
pry(main)> query.to_s
=> "-fav:evazion"
```

The query has been modified so that the `ordfav` metatag has turned into a `fav` metatag. Building the query shouldn't modify the query. Instead it should be changed in a local variable like this:

```
metatag_name = metatags_without_ord[metatag.name] if metatag.negated && metatags_without_ord.key?(metatag.name)
clause = metatag_matches(metatag_name, metatag.value, quoted: metatag.quoted)
```

PR #5010, opened 2022-02-14
Moderation Reports: add link for moderators to navbar

nonamethanks:
It turns italic and is styleable when there's pending reports just like the forum link.

This is so that we can actually check at a glance if there's new reports without having to have a bookmark to the report page or using distill.


BrokenEagle:
Should site wide statuses like this be stored in the DB to facilitate lower DB load, since this query will be running on every page load.  As a separate but related topic, should user statuses like last forum updated be stored on the User table for the same reason?

PR #5023, opened 2022-02-28
favorite_groups: make search[order]=updated_at work

9suika:
The UI offers an option to sort by modification date but it doesn't do anything, might've been left out after a rewrite. This PR implements it.


evazion:
Strictly speaking, it should be `order(updated_at: :desc, id: :desc)` in case two favgroups have the same updated_at timestamp. This is very rare, but has been known to happen.

PR #5025, opened 2022-03-01
Limit the width of usernames in tables

NamelessContributor:
Fixes the layout issues caused very long usernames in the forum and elsewhere.

I'm not sure if this should go in `tables.scss` but it seemed the most logical to me.
I think a shorter username limit (suggested in the Discord) wouldn't be a replacement for this, as names using many wide characters could still cause issues on smaller screens.

### DIFF
--- a/app/javascript/src/styles/common/tables.scss
+++ b/app/javascript/src/styles/common/tables.scss
@@ -76,3 +76,11 @@ table.table-sm {
 table.table-md {
   th, td { padding: 0.5rem; }
 }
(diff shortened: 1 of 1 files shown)


evazion:
I think a simple `max-width: 180px` would be a bit better here. `15vw` is 288px at 1920x1080, which is pretty big.

For reference, the median username length is 8 characters. The 99% percentile is 18 characters. 18 characters is around 180px wide max.

PR #5039, opened 2022-03-08
Lofter: simplify tag extraction logic

nonamethanks:
Now that we have a separate parsing class we can just use it to properly parse tag urls as well.

### DIFF
--- a/app/logical/source/url/lofter.rb
+++ b/app/logical/source/url/lofter.rb
@@ -28,7 +28,7 @@
 class Source::URL::Lofter < Source::URL
   RESERVED_SUBDOMAINS = %w[www.lofter.com i.lofter.com]
 
-  attr_reader :username, :work_id
+  attr_reader :username, :work_id, :unescaped_tag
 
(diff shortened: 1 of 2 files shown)


app/logical/source/url/lofter.rb
```diff
@@ -48,6 +48,10 @@ def parse
       @username = $1
       @work_id = work_id
 
+    # https://gengar563.lofter.com/tag/%E5%BA%9F%E5%BC%83%E7%9B%90%E9%85%B8%E5%A4%84%E7%90%86%E5%8E%82
+    in /^([a-z0-9-]+)\.lofter\.com$/, "tag", tag
+      @unescaped_tag = CGI.unescape(tag)
```
evazion, line 53: Unescaping the path should probably happen globally in Danbooru::URL#initialize, but it would need testing to make sure it doesn't break other things.

https://github.com/danbooru/danbooru/blob/master/app/logical/danbooru/url.rb#L42

PR #5047, opened 2022-03-11
Add tag category: deprecated

### LINKED ISSUE #4562: Suggestion: Ambiguous/Do-not-use tag category
This specific suggestion, among the various categories, has been proposed many times before, but I just noticed Gelbooru has it too ([example](https://gelbooru.com/index.php?page=post&s=view&id=5414568&tags=booty)), which makes a strong case for adopting it since it seems to work fine for them, so I'm opening this for consideration.
image

The category would be basically for tags that should not be used. This doesn't sound very useful at first, but there are many implications to consider: first of all the ability to use a "ambtags:>0" (or whatever it would be named) search, and then the fact that a warning could be tied to such tags so that people using them immediately notice. Also, it'd be much easier to spot such tags in the sea of blue that normal taglists can be.

This kind of category would be especially useful for things that cannot be aliased [(keyboard](https://danbooru.donmai.us/posts?tags=keyboard) is an example).

I know we want to avoid having a rainbow on the sidebar, but as these tags should never be used, that shouldn't be a problem, as they would effectively be handled like tagme and regularly sorted through.

And we have many of these tags. Here's a very partial list: https://danbooru.donmai.us/wiki_pages/list_of_disambiguation_pages

Feel free to immediately close this of course if it's rejected.

nonamethanks:
* Only empty tags can be moved to deprecated manually
* No tags can be moved out of this category manually, only via BUR
* These restrictions don't apply to admins

Also adds a depretags:>0 search

Fixes #4562.

Ok, hear me out. This category is severely needed for proper gardening. There's a lot of massive tags like "Tied hair" (https://danbooru.donmai.us/forum_topics/20459), or the recent neckwear revert project, that would greatly benefit from the existence of this tag, or are simply impossible to tackle without it.

* it's impossible to currently be able to tell which tags are not supposed to be used. nobody reads wikis, and this is especially true for large tags with an "obvious" meaning
* having an entire category for deprecated tags encourages users to help in sorting them out, since it's immediately visible in the sidebar, not relegated to the wiki. If we want it to be more in-your-face it can be put above general in the sidebar, but for now it's under it but above meta to avoid making too many changes
* this change allows someone to use depretags:>0 to find posts that need to be handled, and lets us use a single search instead of browser bookmarks to keep track of people reusing the million empty ambiguous tags we have like "tree trunk" (https://danbooru.donmai.us/forum_topics/20488).
* this change allows us in the future to show warnings when someone uses a deprecated tag 
* the issue with people changing tag categories without telling anyone is handled by making only empty tags moveable to, but not from, the deprecated category. this allows anyone to move old ambiguous wikis, but prevents them from moving deprecated tags back to general without using a BUR. This means we don't need to implement a whole category versioning system just to keep track of them, because if someone empties a tag just to move it to deprecated we'll be able to see it in the post changes history
* realistically, any other complex rework of the tag system to mark wikis/tags is years away, while this category is needed yesterday

This is what it looks like:
image
image

It uses a similar color to gelbooru, but without the strikethrough because that makes it hard to actually read the tag name. It also uses the same color as Platinum, just like gold with meta, builder with copyright, etc, thus it's not a fully new color being introduced, but something that fits into our color scheme.

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -137,7 +137,7 @@ class Tag < ApplicationRecord
       end
     end
 
-    # define artist?, general?, character?, copyright?, meta?
+    # define artist?, general?, character?, copyright?, meta?, deprecated?
     TagCategory.categories.each do |category_name|
       define_method("#{category_name}?") do
         category == Tag.categories.send(category_name)
@@ -200,7 +200,7 @@ class Tag < ApplicationRecord
             # next few lines if the category is changed.
             tag.update_category_cache
 
-            if Pundit.policy!(creator, tag).can_change_category?
+            if Pundit.policy!(creator, tag).can_change_category? && !(category_id == TagCategory.mapping["deprecated"] && tag.post_count > 0 && !creator.is_admin?)
               tag.update(category: category_id)
             end
           end
--- a/app/policies/tag_policy.rb
+++ b/app/policies/tag_policy.rb
@@ -2,9 +2,11 @@
 
 class TagPolicy < ApplicationPolicy
   def can_change_category?
-    user.is_admin? ||
-      (user.is_builder? && record.post_count < 1_000) ||
-      (user.is_member? && record.post_count < 50)
+    return true if user.is_admin?
+    return false if record.category == TagCategory.mapping["deprecated"]
+    return true if user.is_builder? && record.post_count < 1_000
+    return true if user.is_member? && record.post_count < 50
+    false
   end
 
   def permitted_attributes
--- a/app/components/tag_list_component.rb
+++ b/app/components/tag_list_component.rb
@@ -40,6 +40,6 @@ class TagListComponent < ApplicationComponent
   end
 
   def is_underused_tag?(tag)
-    tag.post_count <= 1 && tag.general? && tag.name !~ /_\((cosplay|style)\)\z/
+    tag.post_count <= 1 && (tag.general? || tag.deprecated?) && tag.name !~ /_\((cosplay|style)\)\z/
   end
 end
--- a/app/components/tag_list_component/tag_list_component.html+categorized.erb
+++ b/app/components/tag_list_component/tag_list_component.html+categorized.erb
@@ -1,7 +1,7 @@
 <div class="tag-list categorized-tag-list">
   <% categorized_tags(TagCategory.split_header_list).each do |category_name, tags| %>
     <h3 class="<%= category_name %>-tag-list">
-      <%= category_name.capitalize.pluralize(tags) %>
+      <%= TagCategory.title_map[category_name] %>
     </h3>
 
     <ul class="<%= category_name %>-tag-list">
--- a/app/javascript/src/styles/base/040_colors.css
+++ b/app/javascript/src/styles/base/040_colors.css
@@ -295,11 +295,13 @@ html {
   --general-tag-hover-color: var(--link-hover-color);
   --meta-tag-color: var(--orange-3);
   --meta-tag-hover-color: var(--orange-2);
+  --deprecated-tag-color: var(--grey-5);
+  --deprecated-tag-hover-color: var(--grey-3);
 
   --user-admin-color: var(--red-5);
   --user-moderator-color: var(--character-tag-color);
   --user-builder-color: var(--copyright-tag-color);
-  --user-platinum-color: var(--grey-5);
+  --user-platinum-color: var(--deprecated-tag-color);
   --user-gold-color: var(--meta-tag-color);
   --user-member-color: var(--general-tag-color);
   --user-restricted-color: var(--general-tag-color);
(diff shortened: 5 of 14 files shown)


app/models/tag.rb
```diff
@@ -200,7 +200,7 @@ def find_or_create_by_name(name, creator: CurrentUser.user)
             # next few lines if the category is changed.
             tag.update_category_cache
 
-            if Pundit.policy!(creator, tag).can_change_category?
+            if Pundit.policy!(creator, tag).can_change_category? && !(category_id == 6 && tag.post_count > 0 && !creator.is_admin?)
```
BrokenEagle, line 203: This usage of the number 6 should be some kind of self-commenting variable instead of what is otherwise just a random number.

`TagCategory.mapping["deprecated"]` should work, but perhaps the `TagCategory` module could use a new enum so that hash representation doesn't have to be used. At one point in time the TagCategory module had such a thing, but it appears to have been removed at some point in the past.

app/policies/tag_policy.rb
```diff
@@ -2,9 +2,11 @@
 
 class TagPolicy < ApplicationPolicy
   def can_change_category?
-    user.is_admin? ||
-      (user.is_builder? && record.post_count < 1_000) ||
-      (user.is_member? && record.post_count < 50)
+    return true if user.is_admin?
+    return false if record.category == 6
```
BrokenEagle, line 6: Same issue as above. This should be some kind of self-commenting variable.

nonamethanks:
Fixed.

evazion:
My main qualm about this is that just going to add more noise to the tag list. Here's what I think is going to happen. We're going to add deprecated tags. We're going to mark a bunch of tags as deprecated. We're never going to clean them up. Because if we could, we would have done it already. Instead every post is going to have a bunch of deprecated tags sitting forever at the bottom of the tag list, just like all the other "fixme" tags (`bad_*`, `*_request`, etc) that never get fixed. Except worse, because now it's two separate sections of "fixme" tags at the bottom of every tag list instead of one.

As a regular user, this is just plain annoying. The fact that some tag is considered "do not use" or is in the process of being cleaned up is of no concern to me. I don't need the fact the uploader used the wrong tag called out to me; it's his problem, not mine.

I'd feel better about this if it were a flag rather than a category. For multiple reasons:

* Treating it as a category loses information about the actual category. Imagine we wanted to mark a copyright tag like `nintendo` as deprecated. Or say we wanted to mark all the ambiguous character or artist tags as ambiguous before we clean them up. But we can't because that would lose category information. A tag's type is conceptually separate from its ambiguous status.
* I hate having to add another column to the posts table just to support a `depretags:>0` search that I think will rarely get used.  If you want to garden deprecated tags, you're better off picking a single tag than trying to go through every post at once. Especially if we end up having 100k+ posts under dozens of different deprecated tags, like I think we're going to.
* The `tag_count_deprecated` column also makes marking a tag as deprecated very slow, because it requires updating every post when the category changes. It's just like an alias or mass update, it has to touch every post.
* I don't think we need to call out deprecated tags in a whole separate section in the tag list. Categories are supposed to be for calling out the most important tags. Instead we're using them to call out the least important tags. The hope is that people will see that e.g. `tied_hair` is deprecated and clean it up. I don't think it will play out that way. As a regular user, I don't care that some random tag is considered deprecated now; it's just noise to me.
* If we want people to stop using deprecated tags, we can just lock them. We don't need categories to inform people not to use a tag when we can just treat it as locked and prevent anyone from adding it to new posts.
* There are no good tag colors left after this. All we have left is shit brown and slightly different shades of existing colors. There have been various suggestions for new categories for things like outfit tags or meme tags, but we can't do that when all the best colors are used up on lower value things.

nonamethanks:
Then what do you propose? Having a way to mark tags as deprecated on top of their existing category? So a new column in the tags table?

evazion:
> Then what do you propose? Having a way to mark tags as deprecated on top of their existing category? So a new column in the tags table?

Yeah, just a new flag in the tags table. Something called `is_deprecated` or `is_locked` or `do_not_use` or something like that. It's all the same thing: a flag saying the tag shouldn't be used.

> How would that be reflected in the tag list? Only way I can think of that does not look ugly is a strikethrough.

I don't think the fact that a tag is deprecated necessarily needs to be shown in the tag list. If someone tries to add a deprecated tag, they can be warned then. Or there could be a notice in the wiki page, or when you search for the tag.

Look at it from the perspective of regular users. As a regular user, I don't care that some random tag that I never use anyway is considered deprecated. I'm not going to be the one to clean it up. What I'm afraid of happening is that we end up with a bunch of deprecated tags that never get cleaned up, so that every time you click on a post, you're nagged about some random bad tag you don't care about.

>That also would not allow us to easily search for all posts with deprecated tags.

I don't think searching for all posts with deprecated tags is a useful thing to do. Trying to clean up `black_sleeves` and `tied_hair` and N other deprecated tags at the same time isn't going to work; you're better off cleaning up one at a time.

The main use case for me would be finding new posts that have had deprecated tags added to them. But if we say that deprecated tags can't be added to new posts, then that can't happen.

luni3359:
What if there was a way to toggle on a "Contributor" mode, that allowed you to see them for ease of gardening? I think it's a good compromise that wouldn't affect normal users while offering options to contributors.

Normal users wouldn't have their results affected, and no information from tag history would be lost due to a tag being deprecated.

nonamethanks:
Well, I suppose making deprecated tags not usable and having a checkbox on /tags that allows to search for deprecated ones, with the already existing filter of post count > 0, is the alternative.

PR #5049, opened 2022-03-15
simple_form: Don't hide hints on mobile. Fix #4997

### LINKED ISSUE #4997: Show "forgot password" link on login page on mobile
At least once a week we get someone on discord who can't figure out how to reset their password, because the password reset button is only shown on desktop, not on mobile. It should be shown no matter the screen size.

NamelessContributor:
Hints are displayed below their fields on small screens.

Hints for `.text` and `.dtext` inputs weren't hidden even before this change.

### DIFF
--- a/app/javascript/src/styles/common/simple_form.scss
+++ b/app/javascript/src/styles/common/simple_form.scss
@@ -39,7 +39,8 @@ form.simple_form {
       padding-left: 1em;
 
       @media (max-width: 660px) {
-        display: none;
+        padding-left: 0;
+        display: block;
       }
     }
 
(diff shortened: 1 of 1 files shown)


evazion:
These were hidden because there are a lot of hints like "Use * for wildcard" that just get in the way on mobile. Or keyboard shortcut hints like "Ctrl+enter to submit" that are irrelevant on mobile.

NamelessContributor:
Ironically, the "Ctrl+enter to submit" are among those that were not hidden before this change: https://github.com/NamelessContributor/danbooru/blob/c3a5ce9019ae0690c38c97a211ffc4e48eca2cb4/app/javascript/src/styles/common/simple_form.scss#L60
Maybe we could add a class to explicitly hide some hints on mobile (or the reverse)?

PR #5053, opened 2022-03-16
posts: fix sidebar min-width on small screens (fix #5052)

### LINKED ISSUE #5052: Long blacklisted matches aren't correctly cut-off at a specific window width
This only happens at a specific window width, because this problem doesn't occur with a fully maximized window.

To reproduce have one long blacklist line and then slowly shrink your window horizontally before the layout changes to mobile until the line fully expands.

I don't know if it matters but I have thumbnails set to huge.

Blacklist bug

NamelessContributor:
I think the intent of c0610cb809f40d00bc0df02678986f9bf0103805 was to reduce the sidebar width on screens smaller than 1920px? Not sure why the 1080px query was added but it causes the sidebar to have no min-width on screens smaller than 1080px.

### DIFF
--- a/app/javascript/src/styles/common/main_layout.scss
+++ b/app/javascript/src/styles/common/main_layout.scss
@@ -32,6 +32,8 @@ footer#page-footer {
   #sidebar {
     flex: 0;
 
+    min-width: 230px;
+
     @media (min-width: 1080px) {
       min-width: 245px;
     }


evazion:
This would make it so that you can only fit one large-size thumbnail per row instead of two at 768x1024 (tablet in portrait mode)

Before:

image

After:

image

NamelessContributor:
Added a default min-width of 230px instead of removing the media query from the 245px.

PR #5060, opened 2022-03-21
twitter: support www.twitter.com URLs

NamelessContributor:
Fixes NoMethodError preventing posts from being viewed if the source is
www.twitter.com


evazion:
Better fix in 7394660ba901bf0423f81537612dcb6b52d13d29. Match by domain instead of hostname so we don't have enumerate every possible hostname.

PR #5074, opened 2022-03-29
change contributor -> unrestricted on users listing

CoreMack:
On https://danbooru.donmai.us/users the unrestricted privilege is just called Contributor.  No one calls unres users "contributors" anymore, to avoid confusion it should just say "Unrestricted?" instead of "Contributor?"

### DIFF
--- a/app/views/users/index.html.erb
+++ b/app/views/users/index.html.erb
@@ -5,7 +5,7 @@
     <%= search_form_for(users_path) do |f| %>
       <%= f.input :name_matches, label: "Name", input_html: { value: params[:search][:name_matches], data: { autocomplete: "user" } } %>
       <%= f.input :level, collection: User.level_hash.to_a, include_blank: true, selected: params[:search][:level] %>
-      <%= f.input :can_upload_free, label: "Contributor?", as: :select, include_blank: true, selected: params[:search][:can_upload_free] %>
+      <%= f.input :can_upload_free, label: "Unrestricted?", as: :select, include_blank: true, selected: params[:search][:can_upload_free] %>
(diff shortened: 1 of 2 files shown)


9suika:
Another instance of "Contributor", might be worth changing too for consistency:
https://github.com/danbooru/danbooru/blob/8b2798d487e75f2bf8ba9ca43c3420849d331af9/app/views/users/show.html%2Btooltip.erb#L17

CoreMack:
changed tooltip as well

nottalulah:
I don't really see the benefit of having a 'finer-grained' permission system when in practice we've fallen back into the previous hierarchy. Sure, you *can* have approvers who aren't builder/unres, and you *can* have mods who aren't approver/unres, but in practice we've just returned to Builder -> Contributor -> Janitor -> Moderator. I'm probably the minority opinion, but I preferred the old system.

CoreMack:
I don't think anyone active enough on the site to notice and care about the user tooltip would be confused about what "Unrestricted" means.  That said, it might be preferable to just change to "Builder" rather than "Unrestricted."  Either would be better than "Contributor", which is an outdated, confusing term.
If we want to discuss moving away from a permission system back to the older hierarchy, that should be its own issue.

evazion:
I'd rather switch back to calling them Contributors everywhere. It's easier to talk about "Contributors" and "Approvers" than "Users with unrestricted uploads permission" and "Users with approval permission".

"Unrestricted" is a bad term because it doesn't explain what it does (unrestricted from what?), it doesn't fit the same format as other roles (Member/Builder/Approver/Moderator/Administrator), and because there is already a "Restricted" role, which is definitely not the opposite of Unrestricted.

PR #5099, opened 2022-04-07
Add ability to mark tags as deprecated

### LINKED ISSUE #4562: Suggestion: Ambiguous/Do-not-use tag category

nonamethanks:
Let's try this again. Fixes #4562, following the discussion in #5047.

* Deprecated tags can't be added to posts, but existing deprecated tags in a post won't be removed
* Only empty tags can be marked as deprecated manually
* No tags can be manually undeprecated
* These limits don't apply to admins
* Deprecating or undeprecating a tag will create a new mod action to prevent people from going rogue
* Added deprecate/undeprecate commands for BURs
* Deprecating a tag via BUR removes all implications to and from it as well

Note: some of these tests fail for me, but other similar older tests fail too, and testing the api and various user levels it all seems to work fine, so I'm not sure if the tests don't work because my setup is wrong.

I noticed that we still have the is_locked attribute in the database for tags, but it was used for other things in the past so I thought it was a bad idea to repurpose it, given that it was originally meant to be for categories.

I did not add any visual indication to the tag list for now because as you said the average user won't really care. But we could mark the number as red like we do for low count posts, or mark them as strikethrough. I didn't want to add that in this BUR though because I didn't know what you preferred.

Lemme know if anything needs to be changed.


app/logical/bulk_update_request_processor.rb
```diff
@@ -182,6 +198,16 @@ def process!
           tag = Tag.find_or_create_by_name(args[0])
           tag.update!(category: Tag.categories.value_for(args[1]))
 
+        when :deprecate
+          tag = Tag.find_or_create_by_name(args[0])
+          tag.update!(is_deprecated: true)
+          TagImplication.active.where(consequent_name: tag.name).each { |ti| ti.reject!(User.system) }
+          TagImplication.active.where(antecedent_name: tag.name).each { |ti| ti.reject!(User.system) }
```
BrokenEagle, line 205: This could be done with a single line if you use an OR. It should be something like:
```ruby
TagImplication.active.where(consequent_name: tag.name).or(where(antecedent_name: tag.name)).each { |ti| ti.reject!(User.system) }
```

app/logical/bulk_update_request_processor.rb
```diff
@@ -182,6 +198,16 @@ def process!
...
+
+        when :undeprecate
+          tag = Tag.find_or_create_by_name(args[0])
+          tag.update!(is_deprecated: false)
+
```
BrokenEagle, line 210: Should the tag implications be reactivated? I'm thinking of the case where a mistake was made, or something else where the original status needs to be restored.

app/logical/bulk_update_request_processor.rb
```diff
@@ -240,6 +266,8 @@ def to_dtext
         else
           "nuke {{#{args[0]}}}"
         end
+      when :deprecate, :undeprecate
+        "#{command.to_s.tr("_", " ")} [[#{args[0]}]]"
```
BrokenEagle, line 270: The `.tr("_", " ")` shouldn't be necessary, since neither command has an underscore.

BrokenEagle:
Should users other than Admins be able to undeprecate tags when the post count is 0? I'm thinking that mistakes could be made, and we only have so many active Admins.

Also, I didn't see any sort of CSS additions to highlight a tag's deprecated status. At the very least, the tag HTML entities should get a class so that users can add custom CSS styles themselves. Perhaps something like `tag-deprecated` ?

app/logical/bulk_update_request_processor.rb
nonamethanks: I don't think mistakes in BUR are likely to happen, so if they did it they can just be readded manually.

app/logical/bulk_update_request_processor.rb
nonamethanks: This doesn't work for me, it fails with "undefined method `where'". I think it has to be `.or(TagImplication.active.where`, which becomes a really long line which is hard to read.

nonamethanks:
It's this way because I want to avoid having someone removing deprecation without realizing what they're doing. 

The admin load right now is pretty manageable, and I don't think we'll be deprecating hundreds of tags a month, so I don't think it'll be an issue, but if @evazion would prefer for everyone to be able to remove deprecation from empty tags I'll remove it.

Keep in mind ideally every deprecated tag would have a count of 0, so that check would in theory apply to every deprecated tag.

BrokenEagle:
Yeah, one way to perhaps mitigate users from undeprecating tags that they shouldn't, would be to check if there was a BUR or an Admin which deprecated the tag, although admittedly the logic for that would be pretty complicated.

app/logical/bulk_update_request_processor.rb
BrokenEagle: Yeah, I know I had done it before myself, and I was just speaking off the cuff, since I don't currently have a up-to-date working instance. I was specifically looking at the following when coming up with that suggestion.

https://github.com/danbooru/danbooru/blob/036341d8bab735a94c7a8122b00e8adb489c84ec/app/models/tag_relationship.rb#L69-L71

Is it possible to do something like the following though?

```ruby
q = TagImplication.active
q.where(consequent_name: tag.name).or(q.where(antecedent_name: tag.name)).each { |ti| ti.reject!(User.system) }
```

evazion:
> Should users other than Admins be able to undeprecate tags when the post count is 0? I'm thinking that mistakes could be made, and we only have so many active Admins.

Regular users shouldn't be able to undeprecate tags. Part of the use case here is to prevent people from repopulating tags that shouldn't be populated.

I'm a little worried about letting regular users deprecate empty tags. I'm afraid people are going to go nuts marking every possible typo tag as deprecated to prevent them from being added to posts.

> I did not add any visual indication to the tag list for now because as you said the average user won't really care. But we could mark the number as red like we do for low count posts, or mark them as strikethrough. I didn't want to add that in this BUR though because I didn't know what you preferred.

This can be added later if I change my mind or we don't end up with an overwhelming number of deprecated tags. It's easier to add it later than to take it away after users have gotten used to it.

PR #5100, opened 2022-04-08
Update _secondary_links.html.erb: help:upload -> howto:upload

iDolmatov:
help:upload has beed deleted from the wiki. 
Fix the button "Help", re-assigned to "howto:upload".


nonamethanks:
I deleted help:upload because it was a wall of text of outdated nonsense, but I didn't realize it was linked in the site. I'll move howto:upload to help:upload instead, since the rest of the hardcoded pages follow that naming scheme too.

PR #5104, opened 2022-04-09
Tags: don't allow deprecation of tags without wiki

nonamethanks:
This was raised as a concern in the discord, and I agree that a requirement for deprecating tags would be to offer alternatives in their wiki.

Also fixed and moved the tests for the BUR because I realized I had put them in the wrong place.

### DIFF
--- a/app/logical/bulk_update_request_processor.rb
+++ b/app/logical/bulk_update_request_processor.rb
@@ -131,13 +131,19 @@ class BulkUpdateRequestProcessor
 
         when :deprecate
           tag = Tag.find_by_name(args[0])
-          if tag.is_deprecated?
+          if tag.nil?
+            errors.add(:base, "Can't deprecate #{args[0]} (tag doesn't exist)")
+          elsif tag.is_deprecated?
             errors.add(:base, "Can't deprecate #{args[0]} (tag is already deprecated)")
+          elsif tag.wiki_page.blank?
+            errors.add(:base, "Can't deprecate #{args[0]} (tag must have a wiki page)")
           end
 
         when :undeprecate
           tag = Tag.find_by_name(args[0])
-          if !tag.is_deprecated?
+          if tag.nil?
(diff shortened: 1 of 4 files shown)


app/logical/bulk_update_request_processor.rb
```diff
@@ -131,13 +131,15 @@ def validate_script
...
+          elsif tag&.wiki_page.blank?
+            errors.add(:base, "Can't deprecate #{args[0]} (tag must have a wiki page)")
           end
 
         when :undeprecate
           tag = Tag.find_by_name(args[0])
-          if !tag.is_deprecated?
+          if !tag&.is_deprecated?
             errors.add(:base, "Can't undeprecate #{args[0]} (tag is not deprecated)")
           end
```
BrokenEagle, line 144: If I'm reading this correctly, then currently, there will be no errors if the tag doesn't exist. Which means that it won't require a wiki page to deprecate. I understand the intention of doing that, i.e. you want to be able to deprecate tags which don't currently exist, however, if you DO want a wiki to be created, even if the tag doesn't exist, then you should change the following

```ruby
elsif tag&.wiki_page.blank?
```

...to...

```ruby
elsif WikiPage.find_by_name(args[0]).blank?
```
(*not entirely sure about the syntax*)

Also, if the tag does not exist, at what point does it get created? Does the BUR create it? Or will the BUR error out because the tag does not exist when it gets executed?

nonamethanks: No, right now if the tag does not exist it will throw out an error about a missing wiki page. `tag&.wiki_page.blank?` Will always be false if the tag does not exist, because it equals to nil.blank?, which is falsey.

I'm not sure if we should allow people to deprecate tags that have never been used before. That opens a rabbit hole of trying to deprecate every ambiguous term under the sun.

nonamethanks: There was still the issue of the wrong error being shown for tags that don't exist I guess. I'll fix it.

PR #5120, opened 2022-04-13
favgroups: allow favgroup:any/none searches

nottalulah:
Only matches on the current user's favgroups.


evazion:
`any` and `none` are valid favgroup names and there is a user with a favgroup literally named `any`.

PR #5156, opened 2022-05-05
post tooltips: add support for non-web sources

NamelessContributor:
Show "non-web source" in the tooltip instead of "no source", link to the
`non-web_source` tag and show the source on mouseover.


evazion:
It would be better to use `!@post.web_source?` here.

PR #5220, opened 2022-06-29
Uploads: allow admins to bypass duration limits again

### LINKED ISSUE #5081: Admin ability to upload posts longer than 140 seconds is broken
It just returns "Upload failed: Duration must be less than 140 seconds."
You can try with https://files.catbox.moe/3mdnv8.mp4 (not that I wanted to upload it, just one example from the request thread)

### LINKED ISSUE #5217: Twitter video .02 seconds over the duration limit
https://twitter.com/mutturineko/status/1130149811720560641

140.02 seconds long, so its blocked from uploading. 
I would appreciate a longer duration limit in general, since there's a decent amount of touhou animations going beyond 2:20, though I get the aversity to those 14 minute overwatch SFM porn videos.

nonamethanks:
Fixes #5081, also fixes #5217.


evazion:
Validations shouldn't use `CurrentUser.user`. That's an anti-pattern that makes testing harder. Media assets in general shouldn't rely on the current user because they're meant to be completely independent of the uploader.

Issue #5206, opened 2022-06-14
Add `api:` to the list of meta wikis


BrokenEagle:
I.e. the following piece of code:

https://github.com/danbooru/danbooru/blob/15a0fd604bc58c1da1f3d028160bf4bac28734bf/app/models/wiki_page.rb#L6

Just change it to:

```ruby
  META_WIKIS = ["list_of_", "tag_group:", "pool_group:", "howto:", "about:", "help:", "template:", "api:"]
```

nonamethanks:
Fixed in ab900beffd0fc108395b5923575c0938ecb64752.

PR #5250, opened 2022-09-11
Added `api:` to the list of META_WIKIS

jwood7423:
As of right now, api wiki pages are categorized as "general" when they should be categorized as "meta". Please see issue #5206


evazion:
I'd rather consolidate all the help pages into the `help:` namespace than have N different namespaces.

BrokenEagle:
> I'd rather consolidate all the help pages into the help: namespace than have N different namespaces.

I'd be fine with this.

Issue #5275, opened 2022-10-03
Allow artist tags to implicate meta tags


Mokou:
When discussing the issue of Danbooru images being used for AI art training [I suggested](https://danbooru.donmai.us/forum_posts/223089) that a meta tag such as `do_not_train` be created for implication from artist tags for those that don’t want to be used for training AI models. However it seems that cross-category implications have been disabled in danbooru@4a4c198.

I think it could be useful to allow for this type of implication, specifically in this case, but potentially for other cases as well.

luni3359:
I wonder why it was changed. I feel like it's a good feature to have in the grand scheme of things.

nottalulah:
I believe the issue was people kept requesting dumb implications like some_pokemon -> gen_x_pokemon or character -> elf. So the ability to request them at all was removed.

Mokou:
Yeah, that makes sense. I wouldn't think those sorts of requests would a problem if only artist -> meta implications were re-enabled.

PR #5277, opened 2022-10-04
BURs: allow implying tags from artist to meta categories

Mokou:
Addresses #5275
My editor complains that the methods `artist?` and `meta?` don't exist, but also complains about similar methods (i.e. `general?`) not existing as well.


Mokou:
Forgot to update the comments.

nonamethanks:
See issue #5275.

PR #5459, opened 2023-03-09
Add Mistagged and Untagged Posts options from Tags to Wiki Sidebar

pokeball99:


evazion:
I'm worried about mistagging if the average user trusts this too much and starts blindly adding tags without verifying them. Right now AI tags are sort of hidden away, so that only more advanced users even know about them.

pokeball99:
> I'm worried about mistagging if the average user trusts this too much and starts blindly adding tags without verifying them. Right now AI tags are sort of hidden away, so that only more advanced users even know about them.

if it does becomes too much of a problem you could always restrict it above member

Issue #5518, opened 2023-05-06
Add page selection menu on mobile


HeeroWingZero:
Mobile QOL request related to #5505 - as the screenshot below shows, at the bottom of forum threads, search pages, etc there is simply a page number at the bottom of the page in mobile browsers.

![344195037_2206182389769952_4237541885749566635_n](https://user-images.githubusercontent.com/117999306/236593250-19fe58e1-bfd8-4f96-a546-df74afe74811.jpg)

In addition to the previously suggested button for forums, being able to tap on the page number and bring up a pop-up menu to select a page to jump to would make browsing the site on mobile much more streamlined, as currently you'd have to go and change the URL for the same effect. 

On desktop this feature could be implemented by adding a drop-down menu that shows up by clicking the ... shown below

![pagecount](https://user-images.githubusercontent.com/117999306/236593434-e5b60c50-3831-4888-a100-758af5ae8929.JPG)

nonamethanks:
Made the ellipsis "..." clickable with a prompt to insert an arbitrary page.

On mobile, I made the first and last page always show, that way it's not as intrusive as the desktop paginator, which can get quite large:

<img width="529" height="67" alt="Image" src="https://github.com/user-attachments/assets/71afda0e-69ea-49f2-8425-bfbe97015dff" />

PR #5523, opened 2023-05-09
Add page selection menu

ongsalt:
related #5518 
On mobile 

I also add Utility.isMobile to determine user platform based on screen size similar to css .mobile-only


BrokenEagle:
Yo man... you're going to want to squash this down to a single commit to have any hope of this ever being merged.

ongsalt:
> Yo man... you're going to want to squash this down to a single commit to have any hope of this ever being merged.

Done. I'm kinda new to this may I ask why

BrokenEagle:
Yeah, don't include b41563b. If any merging needs to be done, evazion can do it on his side. However, if there are any merge conflicts, then you'll have to resolve them yourself and update the commit.

As for why this is done, it's cleaner and easier to follow. Basically, each commit should encapsulate one entire change to the system, whether it's a minor fix or a feature add. If a pull request has multiple new features and/or changes, then it's fine to break them up each of them up into their own separate commits.

ongsalt:
Is this okay?

BrokenEagle:
For the `Utility.isMobile` function, use the `window.screen` variable instead, so as not to falsely identify browers on desktop computers which have had their widths shrunk beneath the 660px cutoff.

```javascript
Utility.isMobile = function () {
  window.screen.width <= 660 || window.screen.height <= 660;
}
```

I went and added a check for the height as well, since many of your newer devices have a screen height greater than 660, in which case it would falsely detect it as being a desktop when the user has the phone rotated horizontally.

It's outside the scope of this PR, but the mobile CSS should probably be updated to reflect that as well.

```css
/**** Mobile-only CSS ****/
@media (max-width: 660px) and (max-height: 660px) { } 
```

ongsalt:
As for screen height. There are a lot of `@media (max-width: 660px)` in this project I don't know if changing any of this will break something else.

PR #5529, opened 2023-05-27
artists: allow API clients to request sorted_urls

hdk5:
Also includes some changes to make in properly work with the `only` query param and do record prefetching via `includes`.

I want [translate-pixiv-tags userscript](https://github.com/evazion/translate-pixiv-tags) artist tooltip to show artist urls in the same convenient order as danbooru itself does.

### DIFF
--- a/app/logical/parameter_builder.rb
+++ b/app/logical/parameter_builder.rb
@@ -16,19 +16,24 @@ class ParameterBuilder
     # Attributes and/or methods may be included in the final pass, but not includes
     seen_objects << object.class.name
     only_array.each do |item|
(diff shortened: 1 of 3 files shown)


hdk5:
Closing this since what I needed it for was re-implemented client-side (https://github.com/evazion/translate-pixiv-tags/pull/72, https://github.com/evazion/translate-pixiv-tags/pull/73).

Still kinda hope for a solution that would not require changes in userscript whenever there are any changes in Source::URLs, but considering how rarely that happens and how quickly PRs to userscript repo get merged, that is not really necessary.

PR #5538, opened 2023-06-21
Uploads: add source search field

NamelessContributor:


evazion:
This is case-sensitive, so it won't work if you're searching for something like a Twitter username and the case is wrong. It's also based on the raw unnormalized source as entered by the uploader, so if you upload say a [page URL](https://www.pixiv.net/en/artworks/117207551) and an [image URL](https://i.pximg.net/img-original/img/2024/03/24/15/44/18/117207551_p0.png), then you'll have two different uploads with different sources for the same post. Lastly, this column doesn't have an index for wildcard search, so it's going to be slow.

evazion:
Made it case-insensitive. After testing it out, a trigram index on the source column to support wildcard searches probably isn't worth it. It would cost 2GB+ and it wouldn't help most of the time, since most people have a relatively small number of uploads so Postgres finds to more effective to filter by the `uploader_id` index than by the source index.

PR #5539, opened 2023-06-21
Fix bilibili uploads

hdk5:
I noticed that bilibili checks User-Agent header, forcing browsers other than the newest Firefox or Chrome to pass captcha, but for some reason empty User-Agent also works.
Additionally, use more realistic url - of a work page - as a Referer.


hdk5:
Reopening with an alternative proposal that's been working on localbooru for quite a while.

The idea here is to construct the latest firefox user-agent string using its convenient 4-week release schedule.

Issue #3130, opened 2017-06-08
Allow intelligent tag merging when uploading duplicates


Type-kun:
Related to #2027, discussed in later pages of http://danbooru.donmai.us/forum_topics/14067. 

Currently, when a duplicate is uploaded, the tags uploader specified are discarded completely. It's discouraging the users and wasting their efforts, so the tags should get merged. However, merging them automatically is problematic as mentioned in http://danbooru.donmai.us/forum_topics/9999, since there's a possibility of wrong source or md5 collision.

I propose the following: 
1) Create upload merge controller, which would take upload ID as the parameter. If the upload status is not  `duplicate: \d+`, error out.
2) if the status IS `duplicate: \d+`, then grab the post ID, show user the following:
* Post ID with link to said post
* Image 850px sample (or maybe the full image based on user preferences?)
* List of tags the image already has, plus non-editable source and rating fields for info
* Editable textbox containing the tag diff - upload tags minus tags that are already on the post.
* "Merge" button
* Text prompt explaining that the upload was a duplicate and urging the user to double-check the tags
3) If user merges the upload, treat it the same way as if user added the tags manually.
4) if tag diff is empty - that is, all upload tags are already in the post, redirect user to the post and display a flash notice saying that upload was a duplicate instead.

Things to ponder about:
* Rating and source can be specified via tags. Not sure how to handle them, probably better to filter out from the diff
* Pools can also be specified through tags and might be already present on the post. Ideally, diff should contain only the pools that are not already present, but that might prove difficult to implement.
* The controller should probably throw an exception if the upload doesn't belong to the user who opens it

evazion:
I suggest that, instead of creating a separate controller, have the post page take a `merge_tags` param that opens the edit tab with the given tags appended to the end of the tag edit box. This would avoid the need to duplicate large sections of the show page in another place, and it would allow people to use the same editing interface they're accustomed to.

nonamethanks:
@r888888888 this is still a problem unfortunately. Typically if a user submits an upload that is a duplicate they'll be forced to copy their tag list, open the old post and paste it manually. It'd be nice if there was some way to automate or make easier the merging of tags, while accordingly excluding selected meta tags such as the various *_request and commentary/translated tags.

crs00:
I agree with NNT. This issue should be reopened.

Unbreaky:
[...]

PR #5541, opened 2023-06-25
uploads: approve original post when a duplicate is uploaded by approver

hdk5:
Somewhat related to #3130.
To my experience, most approvers would manually approve the post they got sniped on anyway, so I can't think of any reason why this should not be done automatically.


evazion:
What if you upload something and it turns out it was AI-generated? There are times when you might not want to automatically undelete something because it was previously deleted for a good reason you weren't aware of. An approver can always manually undelete it after seeing the deletion reason.

PR #5592, opened 2023-11-09
pixiv: pull original request text for commissioned works

### LINKED ISSUE #5412: Add support for pulling commentary from Pixiv requests
Right now I'm having to copy the commentary manually, and filling in the `[tn]Source:[/tn]` blocks as well.

#### Example:

-  [post #5994576](https://danbooru.donmai.us/posts/5994576)
- https://www.pixiv.net/en/artworks/104214277
- https://www.pixiv.net/requests/122459

Pixiv doesn't use a JSON endpoint for this, but the data is available as JSON encoded into the page itself.

- **HTML ID:** `__NEXT_DATA__`
- **JSON path:** `$.props.pageProps.serverSerializedPreloadedState`
    -  This is an escaped string which is also JSON, i.e. the `"` are `\"` and the `\` are `\\`
- **Sub-JSON path:** `$.pageCommission.requests.122459.requestProposal.requestOriginalProposal`

It's also available on the `<meta>`  **twitter:description** and **og:description** in the page HTML , although I don't know if those would potentially strip/crop information. Using the JSON I described above seems like it would be the best way to get the original data.

hdk5:
Similar to how skeb extractor pulls commissioner request and thanks messages, and that some even [care enough](https://danbooru.donmai.us/posts?tags=commentary%3A"*[tn]Source%3A+https%3A%2F%2Fskeb.jp%2F*") to manually add it to works from other sources, I believe that keeping original commission text from works made with Pixiv Request would make sense as well.

Depends on #5591, fixes #5412


hdk5:
Rebased onto latest master.
The idea of including commissioner text was kinda approved at [forum #266590](https://danbooru.donmai.us/forum_posts/266590)

PR #5598, opened 2023-11-19
Adjusts on text when the Upgrade to Gold is unavailable

### LINKED ISSUE #5265: Handle Gold upgrades not being available
Gold upgrades currently aren’t available and there seems to be no ETA for their return. We’ve been getting an increasing about of complaints from member-tier users about that, not just because they want to throw money at the site, but also because of the upgrade banner and missing Gold features.

Please consider these temporary changes until Gold upgrades become available again:

* Prevent the “Upgrade to Gold” banner from being shown to member-tier users. They can’t upgrade at the moment, so it only annoys them or gives them false hope.
* Increase the tag search limit to 4 or 6 for member-tier users. The 2-tag limit is quite bothersome if there’s no way to increase it. It might also incite more users to upgrade to Gold once that’s available again, because then they know what they’re missing out on. (Not sure how to deal with Saved Searches that assume a higher tag limit after everything goes back to normal, though.)

TheKoakuma:
Based on the comment from unitooni on [forum #21157](https://danbooru.donmai.us/forum_topics/21157?page=24#forum_post_259546) ,those commits are very straightforward, making some changes on the view when upgrade (a.k.a Gold) is unavailable.

Also (partially) Fixes  #5265

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -201,7 +201,7 @@ class User < ApplicationRecord
   concerning :ValidationMethods do
     def validate_enable_private_favorites
       if enable_private_favorites_was == false && enable_private_favorites == true && !Pundit.policy!(self, self).can_enable_private_favorites?
-        errors.add(:base, "Can't enable privacy mode without a Gold account")
+        errors.add(:base, "Can't enable privacy mode without a gold account or higher")
       end
     end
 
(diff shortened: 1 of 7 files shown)


pokeball99:
Personally i feel like it would better to implement a way of toggling upgrade membership purchase rather than hardcoding it like this one way or another

TheKoakuma:
> Personally i feel like it would better to implement a way of toggling upgrade membership purchase rather than hardcoding it like this one way or another 

As far as I could study the code, there is a toggle, but only the upgrade page were using the flag that states this, I just made the changes on Users and Posts pages to react `UserUpgrade.enabled` too.

TheKoakuma:
@nonamethanks @evazion is there any necessary changes to my PR?

ghost:
Well between lack of interest, grammatical errors, programming errors that make this PR not even run, and merge conflicts, it would probably just be easier to rewrite this from scratch if such a change was actually desired.

PR #5600, opened 2023-11-21
uploads: add Ci-En support

hdk5:
```
Post count for ~source:"*ci-en.dlsite.com*" ~source:"*ci-en.net*": 389
Post count for ~source:"*ci-en.dlsite.com*" ~source:"*ci-en.net*" user:hdk5: 55
```

Copying real source and artist commentary is pain, and getting the full size image url which is hidden in html data attribute may be unintuitive for others, so I _really_ would like to have this one merged.

Note that I couldn't make `dtext_artist_commentary_desc` work properly for all supported formatting ([example](https://ci-en.dlsite.com/creator/22367/article/1007342)), but for the most cases it should be fine enough I think.


evazion:
Removed the thing that scrapes the image from the `<meta property="og:image">` tag. Unless I'm missing something, I don't think this is usually desirable since it's usually just a cropped sample of another image in the set.

hdk5:
On the linked example (https://ci-en.net/creator/13341/article/638200, https://danbooru.donmai.us/posts/6783365), that one image seems to appear only as the article preview, but not in the article itself.

evazion:
As far as I can tell, the `og:image` can be either `private/attachment` (which I think is an auto-generated sample of the first image in the article), `public/article_cover` (a custom image set by the artist, which can potentially be unique), or `public/cover` (the artist's profile banner image, which may be used by either their self-introduction article or by articles that don't have any other images). Therefore I think the `og:image` is safe to filter out when it's `private/attachment`.

hdk5:
ok, that sounds reasonable

also, some artist/post urls need to be normalized, since a lot of them link to the adult website domain

PR #5623, opened 2024-01-10
uploads: add Inkbunny support

hdk5:


app/logical/source/extractor/inkbunny.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  def submission
+    api_response[:submissions].to_a.first || {}
+  end
+
+  memoize def api_response
+    return {} unless submission_id.present?
+
+    params = {
+      sid: Danbooru.config.inkbunny_session,
```
hdk5, line 81: I'm slightly unsure about this part tbh.
[Docs](https://wiki.inkbunny.net/wiki/API#Login) say that `sid` should remain valid for several days after its last use, which probably would be good enough.
But, the correct way to do that is to login with username/password and cache the sid until it expires - which I don't really know how to implement here.

evazion: Nijie is an example of how to do this. But it's better to avoid logging in unless we have to. Sites often have captchas or rate limits on their login page to prevent automation. The Nijie tests break half the time they try to log in over and over again and that triggers rate limiting.

evazion: I changed it do username/password login. Now that I actually look at their API, I guess this is what they want us to do and it should be fine if we have to login repeatedly.

PR #5631, opened 2024-01-22
lofter: extract artist display name and post title

hdk5:
https://danbooru.donmai.us/artist_commentary_versions?search[post_tags_match]=source%3A*lofter*&search[updater_id]=499533

### DIFF
--- a/app/logical/source/extractor/lofter.rb
+++ b/app/logical/source/extractor/lofter.rb
@@ -12,9 +12,19 @@ module Source
         if parsed_url.image_url?
           [parsed_url.full_image_url]
         else
-          images = page&.search(".imgclasstag img, .ct .txtcont img")
-          images.to_a.pluck("src").map { |url| Source::URL.parse(url).full_image_url }
+          images = [*images_from_photo_post, *images_from_text_post]
+          images.map { |url| Source::URL.parse(url).full_image_url }
         end
+      end
+
+      def images_from_photo_post
+        page_json.dig("postData", "data", "postData", "postView", "photoPostView", "photoLinks").to_a.pluck("orign")
+      end
+
+      def images_from_text_post
+        content = page_json.dig("postData", "data", "postData", "postView", "textPostView", "content").to_s
+        html = Nokogiri::HTML5.fragment(content)
+        html.css("img").pluck("src")
       end
 
       def profile_url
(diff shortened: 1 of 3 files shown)


9suika:
> <https://danbooru.donmai.us/artist_commentary_versions?search[post_tags_match]=source%3A*lofter*&search[updater_id]=499533>

These have been extracted from the mobile version of the site which can be accessed by requesting the page URL with a mobile user agent.
Details:
URL: same as desktop site
User-agent: `Mozilla/5.0 (Linux; Android 14; LM-Q720) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.6099.43 Mobil
e Safari/537.36`
Post data is stored in the `window.__initialize_data__` JSON object.

None of the current test cases include actual titles by the way, they're just shortened versions of the description. The title is present in the `window.__initialize_data__.postData.data.title` field, empty for all current test URLs. This is the main reason I didn't scrape titles from the desktop site, it was too unreliable to determine which titles are actual titles and not just redundant placeholders. I've been meaning to send a patch to make the extractor use the mobile version but have been putting it off.

hdk5:
> Post data is stored in the `window.__initialize_data__` JSON object.

Oh, I see! That would be even better, since it looks uniform at glance and doesn't have all the retardedness of different desktop themes. I'll look into it. Thank you!

hdk5:
Should be better now

app/logical/source/extractor/lofter.rb
```diff
@@ -12,11 +12,21 @@ def image_urls
         if parsed_url.image_url?
           [parsed_url.full_image_url]
         else
-          images = page&.search(".imgclasstag img, .ct .txtcont img")
-          images.to_a.pluck("src").map { |url| Source::URL.parse(url).full_image_url }
+          images = [*images_from_photo_post, *images_from_text_post]
```
9suika, line 15: Potential regression: <https://jiuhaotaiyangdeshexian.lofter.com/post/73f37cdf_2b86a4ae7> ([post #6128400](https://danbooru.donmai.us/posts/6128400)) doesn't have `photoPostView` or `textPostView` but does have `answerPostView` with an `images` field. This is the only post on Danbooru with this field. Another unhandled field is `videoPostView` ([examples](https://danbooru.donmai.us/posts?tags=source%253A*lofter.com*+video+status%253Aany+-bad_id)) but the current version of the extractor doesn't seem to support Lofter videos anyway.

hdk5: Fixed both answer and video posts

hdk5:
Also, https://sifeizui.lofter.com/post/20401f6e_2b9c30add doesn't work, but that just seems to be an unsupported video format

PR #5644, opened 2024-02-20
uploads: fix skeb

hdk5:
On first request it responds with HTTP 429 and provides `request_key` cookie (with no javascript checks!) that should be sent with subsequent requests

### DIFF
--- a/app/logical/source/extractor/skeb.rb
+++ b/app/logical/source/extractor/skeb.rb
@@ -54,7 +54,31 @@ module Source
       end
 
       memoize def api_response
-        http.cache(1.minute).headers(Referer: profile_url, Authorization: "Bearer null").parsed_get(api_url) || {}
+        response = http.cache(1.minute).get(api_url)
+         
+        if response.status == 429
+          self.cached_request_key = response.cookies.cookies.find do
+            |cookie| cookie.name == "request_key"
+          end&.value
+          response = http.cache(1.minute).get(api_url)
+        end
+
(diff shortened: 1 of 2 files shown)


evazion:
The problem here is that the `request_key` is tied to your user agent and your IP, and in production there are multiple machines with multiple IPs. This may lead to a lot of cache invalidation where machine A gets a cached `request_key` meant for machine B, then invalidates and replaces it, then machine B gets a key meant for machine A, then invalidates and replaces it, and so on...

hdk5:
> The problem here is that the `request_key` is tied to your user agent and your IP, and in production there are multiple machines with multiple IPs. This may lead to a lot of cache invalidation where machine A gets a cached `request_key` meant for machine B, then invalidates and replaces it, then machine B gets a key meant for machine A, then invalidates and replaces it, and so on...

Right. Could this `request_key` be shared not through global `Cache`, but instead, for example, just stored as a `Source::Extractor::Skeb` class variable on each running job executor? Or just not stored at all (making it make an extra request on every extractor initialization)?

evazion:
In production all requests go through an HTTP proxy, and there are multiple proxies (for redundancy purposes). So it can't be cached inside the job worker because each HTTP connection may go through a different proxy and therefore have a different external IP.

>Or just not stored at all (making it make an extra request on every extractor initialization)?

Not caching would certainly be simpler. I guess it doesn't hurt, but it probably doesn't help much in production either.

PR #5646, opened 2024-02-24
The DeviantArt original image exploit doesn't work anymore so just get the highest quality sample possible

viliml:


NamelessContributor:
It should still get the original for deviations with download enabled.

Issue #5624, opened 2024-01-11
Include file type and file size on post tooltip


hdk5:
i.e. format it like on the media asset page like `1.49 MB .jpg, 1000x1484` instead of just `1000x1484`

TPT [has](https://github.com/evazion/translate-pixiv-tags/blob/d783e3c9a842658876ad944328147a26d70f35a9/translate-pixiv-tags.user.js#L2440) this in its artist tooltip, I find it extremely useful, and there's more than enough space to include this extra info

hdk5:
Implemented by #5661.

> The reason tooltips didn't have the filesize and type before was because there's not enough space for it. There's not enough room for everything and I think the size and type are the least important things here.

For me the idea was to make it easier to decide if image from twitter/misskey/\<other site that re-encodes images> is worth posting by making it possible to quickly compare file sizes/dimensions/formats from the glance at tooltip.
But, since I can't do buildering on the main site anymore... 🤷

PR #5661, opened 2024-03-27
Change media asset links to use the same format

DCrow:
Implements #5624
Since there were many places where mostly the same format was used, moved all logic to a separate component for ease of use.

Also almost the same format is used in the media_assets page, but it links to the original file so I wasn't sure if it needed to be changed.
https://github.com/danbooru/danbooru/blob/09dd86b05a23b3740190ebfc9f375cd4306b051c/app/views/media_assets/show.html.erb#L26-L33


nottalulah:
Using a whole component for something this small feels a bit overkill. Contrast something like [`link_to_user`](https://github.com/danbooru/danbooru/blob/09dd86b05a23b3740190ebfc9f375cd4306b051c/app/helpers/application_helper.rb#L245).

DCrow:
@nottalulah I thought about doing a simple helper method, but helpers readme somewhat discouraged it 
https://github.com/danbooru/danbooru/blob/09dd86b05a23b3740190ebfc9f375cd4306b051c/app/helpers/README.md?plain=1#L13-L15

I can change it to a helper method if that's preferable.

DCrow:
Changed component to a helper method.

evazion:
image

The reason tooltips didn't have the filesize and type before was because there's not enough space for it. There's not enough room for everything and I think the size and type are the least important things here.

PR #5690, opened 2024-04-27
Use natural sorting for artist URLs

viliml:
https://discord.com/channels/310432830138089472/310846683376517121/1233865675517329589


evazion:
Fixed in 8ec3abf20. We already have a `Danbooru.natural_sort_by` method defined in `app/logical/danbooru.rb` that should be used instead.

Issue #5698, opened 2024-05-02
Preserve duration when editing ban


ErikHumphrey:
When you edit a ban, often to edit the ban reason, the duration dropdown is always reset to 1 day. Editing the ban should default to the current ban duration so you don't mistakenly reduce the ban length to 1 day.

nottalulah:
This only seems to happen for one-week bans, because they get converted to "7 days" and the iso8601 entries are different (P7D vs P1W). The other lengths aren't reset to 1 day.

nonamethanks:
This was fixed by pull #5702.

PR #5702, opened 2024-05-06
#5698 Fix a bug when editing a 1 week ban, duration resetting to 1 day

DCrow:
When editing a ban and ban duration was different from selectable durations it would reset to the first selection(which was 1 day).

Judging from the comment, there are some bans that don't conform to current restrictions, which breaks duration selector in ban edit page. https://github.com/danbooru/danbooru/blob/fe73bcad3603adc3fc097018151406fc97149c51/app/models/ban.rb#L76-L77

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -171,24 +171,6 @@ module ApplicationHelper
     end
   end
 
-  def humanized_duration(duration)
-    if duration >= 100.years
-      "forever"
-    elsif duration < 0
-      # In production, the oldest bans have negative duration because in 2013 the database was migrated and the
-      # created_at field was reset to 2013, which made their creation date come after their expiration date.
-      "unknown"
-    elsif duration < 1.month
-      duration.in_days.round.days.inspect
-    elsif duration < 1.year
-      duration.in_months.round.months.inspect
-    elsif duration < 100.years
-      duration.in_years.round.years.inspect
-    else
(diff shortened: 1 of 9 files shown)


DCrow:
It seems postgres doesn't support saving week intervals. 
Under the hood it changes weeks to days. 
So when we save 1 week duration we get 7 days in return.
Changed so that we save 7 days and expect 7 days interval in dropdown.

Also on a side note. 
Maybe rename 1 week to 7 days? 
Since this also incorrectly shows 1 week as 7 days. https://github.com/danbooru/danbooru/blob/fe73bcad3603adc3fc097018151406fc97149c51/app/models/ban.rb#L79-L80

nottalulah:
If the label is 7 days why are you doing `1.week.in_days.round.days` instead of just `7.days`?

DCrow:
I thought about that, but then it would differ from
https://github.com/danbooru/danbooru/blob/3d2677a44a9ddc1cc135aa2efb3aeae15ec4b965/app/models/ban.rb#L18
And would raise questions down the line. 
Like why its `7.days` in one place and `1.week` in the other. 

This way at least its more consistent.

nottalulah:
Well right now there's the inconsistency of it being 1.week in the duration validator, but 7.days in the UI. It should just be 7 days everywhere.

DCrow:
Changed in ban model and simplified value in view.

PR #5703, opened 2024-05-07
Fix [spoiler] tags not hiding embedded images

### LINKED ISSUE #5700: DText: Emdedded images aren't hidden by [spoiler] tags.
The image isn't hidden here:

```
[spoiler]
!post #1
[/spoiler]
```

DCrow:
Fixes #5700

### DIFF
--- a/app/javascript/src/styles/common/dtext.scss
+++ b/app/javascript/src/styles/common/dtext.scss
@@ -140,6 +140,10 @@
     background: var(--dtext-spoiler-background-color);
   }
 
+  .spoiler:not(:hover) > * {
+    visibility: hidden;
+  }
+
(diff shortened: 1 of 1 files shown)


evazion:
I think we probably need to get rid of the black background after you hover over a spoiler. Otherwise they look really ugly when you hover over a spoiler image.

image

DCrow:
Maybe constrict background width? 
Instead of hiding it for the full width of comment box
Screenshot 2024-05-16 at 15 53 51
Screenshot 2024-05-16 at 15 53 58

hide only for the width of image

Screenshot 2024-05-16 at 15 53 40
Screenshot 2024-05-16 at 15 54 14

Issue #5602, opened 2023-11-24
Avoid duplicate character names in tab names


kagayakushininggate:
Can be commonly seen with character variants like in the posts below:
https://danbooru.donmai.us/posts/6770511
https://danbooru.donmai.us/posts/6913266

Simply put, the tab names will display the given character's name multiple extraneous times, which looks a little jarring.

I can see two potential variants:
1) Check name matches (for instance, houshou_marine vs houshou_marine_(1st_costume) share "houshou_marine" in their tagname), though this option doesn't cover tags that differ more than just the parenthesis, like hatsune_miku vs yuki_miku,
2) Check implications (for instance, heko_reimu_(hekoningyou) implies hakurei_reimu).

bipface:
I presume by 'tab name' you mean the page title ?

PR #5780, opened 2024-07-03
Remove duplicate unqualified tag names from essential tag string

TheLastZombie:
If a post has multiple tags with the same `unqualified_name`, its page title, file name, etc. list all of their occurrences (instead of just one per tag).

For example, [post #7667814](https://safebooru.donmai.us/posts/7667814)'s `humanized_essential_tag_string` is:

> aubrey, kel, aubrey, and kel (omori) drawn by sr_ld_fr

This PR removes duplicate values from the `chartags` and `copytags` arrays, resulting in a more readable/concise title:

> aubrey and kel (omori) drawn by sr_ld_fr

### DIFF
--- a/app/presenters/tag_set_presenter.rb
+++ b/app/presenters/tag_set_presenter.rb
@@ -24,12 +24,12 @@ class TagSetPresenter
   end
 
   def humanized_essential_tag_string
-    chartags = tags_for_category("character")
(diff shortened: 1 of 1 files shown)


hdk5:
Fixes #5602, I think?

Issue #5638, opened 2024-02-05
Add more filters to refine status:deleted search; let Gold+ users upload straight to status:deleted ahead of modqueue.


hdk5:
I know it may sound as a weird joke, but no, that's a genuine feature request from me.

Gold+ restriction is to avoid spam with garbage that should be deleted manually, I guess.

With this, I also want to search for posts that are
- posted directly to `status:deleted`
- posted to the queue, deleted by DanbooruBot
- - `flag_count:>0 is:deleted flagger:none` takes too long and breaks for older posts
- posted to the queue, deleted manually by Approver+
- - `flag_count:1 is:deleted flagger:any` should work(?), but also takes time
- uploaded/appealed by user of the specific rank
- - `user_level:30..`, according to `User.Levels` constants 

Thus, my ideal personal garbage filter ~~not bound to the taste of elitists jerks who can't handle some awesome mild knot-futa-loli on pokemon watersports action~~ should look like `~fav:hdk5 ~is:active ~is:modqueue ~user_level:>=30 ~appealer_level:>=30 ~(is:deleted -<deleted manually>)`.

I really hope you consider this. Thank you.

bipface:
I don't think posting directly to `status:deleted` will ever be implemented, for a variety of reasons.
The compromise I'd suggest is thinking of ways to soften the penalty for having deleted posts.
For example, I'd like to see a system where posting variant-sets/comics which have less than a high chance of approval can be done without the uploader taking a huge risk.

I have no objection to the other parts of the proposal though.

nonamethanks:
Back in ye olde days before the upload limit rework, comics did take 2/3rd of an upload slot, so that you could upload three comic pages with two upload slots. Unfortunately what actually happened back then was that people would just tag all uploads with comic and then remove it later when they were done dumping, effectively uploading every day more than they were allowed.

nottalulah:
Deletions should kept separate from flags (even evazion said them being the same was an old hack), so that you could search something like `has:deletions` (akin to has:flags which doesn't include DanbooruBot flags). 

[...]

PR #5789, opened 2024-07-11
post queries: add hash-like nested attributes search syntax

hdk5:
This is the proposal to allow filtering posts by related objects attributes in a similar way that is [implemented](https://danbooru.donmai.us/wiki_pages/help%3Ahash_syntax) for other models with the `search` query parameter, but integrated into the tag system instead to account for tag limits, as mentioned before at https://github.com/danbooru/danbooru/issues/5638#issuecomment-1930911388.

The syntax is similar to aforementioned `search` query parameter, and relies on the fact that tags can't start with `[` character.
Example:
- `[uploader_id]:508240`
- `[uploader][name]:Monki`
- `[uploader][level_gteq]:32`
- `[appeals][creator][id]:200729`
- etc.

### DIFF
--- a/app/logical/concerns/searchable.rb
+++ b/app/logical/concerns/searchable.rb
@@ -582,11 +582,19 @@ module Searchable
       end
 
       if params[:"#{name}_include_any_array"]
-        relation = visible(relation, name).where_array_includes_any(name, params[:"#{name}_include_any_array"])
+        items = params[:"#{name}_include_any_array"]
+        items = [items] unless items.is_a?(Array)
+        items = items.map(&:to_i) if type == :integer
+
+        relation = visible(relation, name).where_array_includes_any(name, items)
       end
 
       if params[:"#{name}_include_all_array"]
-        relation = visible(relation, name).where_array_includes_all(name, params[:"#{name}_include_all_array"])
+        items = params[:"#{name}_include_all_array"]
+        items = [items] unless items.is_a?(Array)
+        items = items.map(&:to_i) if type == :integer
+
+        relation = visible(relation, name).where_array_includes_all(name, items)
       end
 
       if params[:"#{name}_include_any_lower"]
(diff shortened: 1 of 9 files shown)


hdk5:
With this implementations there's a huge issue that lets user bypass tag limits by using search expressions that accept post query themselves (e.g. `[tags]:...` or `[replacements][post][tags]:...` or `[replacements][post_tags_match]:...` etc.). I couldn't fix that myself in a non-ugly non-hacky way.
Those realistically could be used for valid queries like, for example, `loli or [parent][tags]:"loli" or [parent][tags]:"loli"`, or `[pools][post_tags_match]:"loli"`, although I don't know how good those will perform on donmai.us scale database, given that similar searches time out for me there.
I guess those nested tag searches can also just be completely disabled, but I couldn't find a nice way to do that either.

The other small thing is that invalid searches (those with no matching relation/attribute in brackets) do not fail - similarly to how the actual search behaves. I was not able to solve that as well, so I'll just leave it like that.

The last minor problem for now is that search queries are executed independently with no way to combine them in case of `has_many` relationship searches.

hdk5:
> search queries are executed independently

by 85bda4d7025d7c86ce8d7eae468e2d516a238357 the proposal is to combine params within a single AND node into one query. Very unobvious, but that turns `[param1]: value1 [param2]: value2` and `([param1]: value1) ([param2]: value2)` into a different searches.

> lets user bypass tag limits by using search expressions that accept post query themselves 

kinda addressed by 47dac1518f739d5a62f4491ed261db4a7544aaee by hard-coding which searches would invoke tag search.
It seems to only be either `tags` for direct Post: https://github.com/danbooru/danbooru/blob/433fad84c150fdc904cbd09cc68746db0599fe42/app/models/post.rb#L1751-L1752
or `#{attr}_tags_match` for Post through association:
https://github.com/danbooru/danbooru/blob/433fad84c150fdc904cbd09cc68746db0599fe42/app/logical/concerns/searchable.rb#L647-L648
(there's also `MediaAsset`'s `ai_tags_match` that doesn't seem to have a limit, but eh, fuck it for now)

Issue #5773, opened 2024-06-22
Skeb uploads are broken


9suika:
Looks like request keys are now set from JS.

hdk5:
won't make a merge request, but here's what we did at our booru, in case that helps:

```rb
# app/logical/source/extractor/skeb.rb:60
        if response.status == 429 && response.mime_type == "text/html"
          request_key = response.to_s.match(/request_key=(\h{8}:\h{8})/)&.[](1)
          if request_key.present?
            self.cached_request_key = request_key
            response = http.cache(1.minute).get(api_url)
          end
        end
```

nonamethanks:
Should be fixed by #5795.

PR #5795, opened 2024-07-26
#5773: Fix skeb extractor

hdk5:
This solution does not involve tinkering with cookies, therefore also making https://github.com/danbooru/danbooru/pull/5644#issuecomment-1958933920 irrelevant.

An alternative would be getting the cookies the new way (https://github.com/danbooru/danbooru/issues/5773#issuecomment-2188898418).


nonamethanks:
Sorry for the big backlog of pulls, I'll see if I can go over them as I have time in the next few weeks.

PR #5798, opened 2024-08-05
posts: fix "Resize to window" for ugoira

hdk5:
There are three things incorrect as of now:

First, `Note.Box.scale_all`, which is supposed to adjust font size for embedded notes, applied it to the whole image/notes container, which also contained ugoira controls, making them incorrectly sized, and the image itself, giving it a large top padding.
The fix is to change `font-size` only for `.note-box` elements.

Second, `Post.resize_ugoira_controls` couldn't calculate the right slider width - it was trying to use the image's width, but the image itself adjusted own width to match slider.
The fix is to get rid of that function and just use CSS.

### DIFF
--- a/app/javascript/src/javascripts/notes.js
+++ b/app/javascript/src/javascripts/notes.js
@@ -358,7 +358,7 @@ class Note {
       let ratio = $container.width() / large_width;
       let font_percentage = ratio * 100;
 
-      $container.css('font-size', font_percentage + '%');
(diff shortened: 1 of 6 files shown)


hdk5:
A minor annoyance (that doesn't bother me enough to fix): the slider width jumps on play/pause due to the button not being constantly sized:
image
image

On current version, however, the slider itself jumps right/left:
image
image

PR #5806, opened 2024-08-14
sources: add e621 extractor

### LINKED ISSUE #5572: Bookmarklet compatibility with e621.net
As it stands, you can only seem to bookmarklet from e621 if you're using the image/file location for the upload. With Gelbooru, this is not the case, and you can bookmarklet straight from the post page. As e621 utilizes more up-to-date Danbooru builds, I would think this to be an easy process.

hdk5:
Follow up to #5634, closes #5572.

I've been keeping this one to myself and then found out there's a pending issue for a proper extractor.
The tests were passing months ago when I made this (trust me bro!), but as some of them link to [not-allowed-anymore](https://e621.net/forum_topics/45501) posts, they might need to be adjusted now.


evazion:
>The tests were passing months ago when I made this (trust me bro!), but as some of them link to [not-allowed-anymore](https://e621.net/forum_topics/45501) posts, they might need to be adjusted now.

Don't use such posts as test cases, I don't want them in the repo.

hdk5:
> Don't use such posts as test cases, I don't want them in the repo.

e621 does not directly give image URLs to such posts, but gives md5 and extension separately. I needed to verify such scenario.

PR #5821, opened 2024-09-23
sources: add DC Inside extractor

hdk5:
A Korean board, somewhat similar to Arca.live.

### DIFF
--- a/app/logical/source/extractor/dc_inside.rb
+++ b/app/logical/source/extractor/dc_inside.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+# @see https://dcinside.com
+# @see Source::URL::DcInside
+class Source::Extractor::DcInside < Source::Extractor
+  def image_urls
+    if parsed_url.image_url?
+      [parsed_url.full_image_url]
+    else
+      image_urls_from_commentary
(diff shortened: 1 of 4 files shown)


hdk5:
https://gall.dcinside.com/mgallery/board/view/?id=mibj&no=5554705 downloads both mp4 and gif for some reason

hdk5:
should be better now, undrafted

app/logical/source/extractor/dc_inside.rb
```diff
@@ -0,0 +1,61 @@
...
+    page&.css(".writing_view_box .write_div")&.css("img, video").to_a.filter_map do |el|
+      next if el.classes.include?("written_dccon")
+
+      if el.name == "video"
+        src = el.attr("data-src")
+      else
+        src = el.attr("src")
+      end
+
+      next unless src.match %r{\Ahttps?://dcimg\d+\.dcinside\.(?:com|co\.kr)/viewimage\.php}
```
evazion, line 24: It'd be better to move the `Source::URL.parse(src)` call up so you could do `parsed_src.image_url?` to detect this. Also `match?(regex)` is preferred over `match regex`.

PR #5824, opened 2024-09-24
Fix tests being broken by CurrentUser

nonamethanks:
See https://discord.com/channels/310432830138089472/310846683376517121/1288146168915951639

https://github.com/danbooru/danbooru/commit/df778c8585097143e3639f936604409a485f580f introduced some calls to CurrentUser that broke every factory instantiaton of post objects. There's more stuff to fix since post.update! also needs to be fixed, and probably some more methods, so this is a wip.

### DIFF
--- a/test/factories/post.rb
+++ b/test/factories/post.rb
@@ -13,6 +13,14 @@ FactoryBot.define do
     source { Faker::Internet.url }
     media_asset { build(:media_asset) }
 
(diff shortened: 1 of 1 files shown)


nonamethanks:
Fixed by 061337895bf287eeac780490322c2de7b5c882b8 after discussion on discord

PR #5830, opened 2024-09-28
Make media assets reportable

### LINKED ISSUE #5825: Allow reporting media assets

nonamethanks:
Fixes #5825.
Still a WIP, need to add tests and think of edge cases.

Media assets can only be reported if they're not associated to posts, otherwise flags must be used. Because of that, I added a different wiki for media asset report notices so that this can be explained separately.

The main change is that it changes reported_user to reported_users, because media assets can be associated with multiple uploads.

image
image

### DIFF
--- a/app/components/media_asset_component/media_asset_component.scss
+++ b/app/components/media_asset_component/media_asset_component.scss
@@ -63,6 +63,12 @@
     }
   }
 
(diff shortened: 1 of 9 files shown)


nonamethanks:
Should be g2g unless I missed something.

PR #5848, opened 2024-10-25
sources: allow ugoira uploads from danbooru

hdk5:
Rough draft as I can't properly test it at the moment.
This will let people running self-hosted booru to steal ugoiras from the main site.
Root model for api request changed from post to media asset to allow grabbing unposted media. Thus, post is also added as searchable attribute for media asset. Otherwise I have to make an extra request for media_asset to get frame delays.

### DIFF
--- a/app/logical/source/extractor/danbooru2.rb
+++ b/app/logical/source/extractor/danbooru2.rb
@@ -53,17 +53,32 @@ module Source
         api_response.dig(:media_metadata, :metadata, :"Ugoira:FrameDelays")
       end
 
-      memoize def api_response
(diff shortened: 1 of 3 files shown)


hdk5:
Flipped some stuff a bit so that the posts with known ids are fetched in a backward-compatible way that doesn't require post to be available as a searchable include for media asset. All tests for the danbooru2 extractor pass now.

nonamethanks:
@hdk5 are you still interested in finishing this, or can it be closed?

hdk5:
this should be working again

PR #5849, opened 2024-10-25
User Feedbacks: add option to hide ban messages from search results

nonamethanks:
This has been bugging me for a while. In times of heavy spam or sockpuppeting, the feedback page becomes pretty much unreadable, and there's no easy way to filter bans out of it.

This pull adds a simple dropdown that allows us to remove standard ban messages from the results, with default being no filtering (same as status quo).
image

https://danbooru.donmai.us/user_feedbacks?commit=Search&page=8&search%5Bbody_regex%5D=%5EBanned shows that the only false positives were joke bans made for april fools', so more accurate filtering doesn't seem necessary.


hdk5:
Would also be convenient to have the same for the promotions

PR #5852, opened 2024-11-01
sources: add odaibako extractor

hdk5:
A free art requests site. Used to allow only sending requests, but now also lets artists to respond with attached images.


evazion:
One of the test cases has gone bad_id and needs to be replaced (see e6542eba795332a3c452f4580f6e46ef668f983b)

PR #5869, opened 2024-11-10
bilibili: refactor the extractor

### LINKED ISSUE #5868: On bilibili articles only three images are extracted
There are actually two (or maybe more) types of text posts there:
- `DRAW` (e.g. [page](https://www.bilibili.com/opus/995617386860118016) [api](https://api.bilibili.com/x/polymer/web-dynamic/v1/detail?id=995617386860118016&features=itemOpusStyle))
- `ARTICLE` (e.g. [page](https://www.bilibili.com/opus/428178320677065986) [api](https://api.bilibili.com/x/polymer/web-dynamic/v1/detail?id=428178320677065986&features=itemOpusStyle))

The former has 4 images which are extracted correctly.

The latter used to have https://www.bilibili.com/read/cv7360489/ format url, which now redirects to `/opus/`.
While on the actual page there are 10 images, only three of them, in no specific order, are extracted.

hdk5:
Fixes #5868

- `api.bilibili.com` is not used anymore, only json from the web page is used.
- `/read/cv` urls seem to be not used anymore, and instead redirect to `/opus/`
- "canonical" post url format changed from `https://t.bilibili.com/` to `https://www.bilibili.com/opus/`, as the latter seems to be more "full" (compare https://www.bilibili.com/opus/428178320677065986 and https://t.bilibili.com/428178320677065986)
- `dtext_artist_commentary_desc` works on existing tests, but might be actually a bit broken. I can't create an account there to test out all the possible bullshit formatting supported by myself. Will gladly accept help if offered though. Just to note, parsing the html from the page is not an option either, as it is also seems to be incomplete.
- For now, reposts support is broken, and I'm too fucking mad to continue dealing with this chinese crap. Hence the draft PR status.

### DIFF
--- a/app/logical/source/extractor/bilibili.rb
+++ b/app/logical/source/extractor/bilibili.rb
@@ -9,9 +9,6 @@ module Source
           [parsed_url.full_image_url]
         elsif parsed_url.image_url?
           [parsed_url.original_url]
-        elsif post_json.present?
-          image_urls = post_json.dig("modules", "module_dynamic", "major", "opus", "pics").to_a.pluck("url")
-          image_urls.to_a.compact.map { |u| Source::URL.parse(u).full_image_url || u }
         elsif article_json.present?
           article_image_urls
         else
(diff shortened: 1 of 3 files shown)


evazion:
>For now, reposts support is broken, and I'm too fucking mad to continue dealing with this chinese crap. Hence the draft PR status.

The way this was handled was wrong before (ex: https://t.bilibili.com/723052706467414039). This was Bilibili's equivalent to a quote-retweet. When given a post quoting or replying to another post, we shouldn't grab the post being quoted, we should only grab the post that was given. Otherwise it leads to inconsistent and confusing behavior when grabbing one post actually grabs another.

Handling this post (<https://t.bilibili.com/723052706467414039>) also required going back to the old method of using the API, since the new /opus/ page isn't available for it.

Issue #5890, opened 2024-12-16
arca.live: use the original post link from cross-posts


hdk5:
When browsing /b/breaking, the link to the post on the original board is present on the page:
![image](https://github.com/user-attachments/assets/b9a11f3d-e5a4-4bd1-ab85-4c64396aa485)
(/b/commission in this case)

Not sure if it is present only on /b/breaking, however, as there might be cross-posts on other boards - couldn't find them myself.

evazion:
Fixed by #5892.

Issue #5875, opened 2024-11-26
arca.live extractor can't extract non-gif animations


hdk5:
See
https://github.com/danbooru/danbooru/blob/f4960115f306ab7aba21b8daedea44f2e85cde8c/app/logical/source/url/arca_live.rb#L25

The original file is not necessarily a gif.

For example, this one is a webp (which is also blocked by #5619)
https://arca.live/b/arknights/122263340
https://ac-o.namu.la/20241126sac/bc1f3cb388a3a2d099ab67bc09b28f0a93c2c4755152b3ef9190690a9f0a28fb.webp?type=orig

I suppose the extractor should try gif and webp, and also fallback to mp4 on failure.

PR #5892, opened 2024-12-19
arca.live extractor tweaks

hdk5:
1. Use the canonical post url when extracting from `/b/breaking` page (#5890)
2. Replace redirect links with actual urls in artist commentary
3. Use the original file extension for videos to support .gif, .webp, or actual .mp4 files (#5875)

### DIFF
--- a/app/logical/source/extractor/arca_live.rb
+++ b/app/logical/source/extractor/arca_live.rb
@@ -44,7 +44,11 @@ module Source
       end
 
       def dtext_artist_commentary_desc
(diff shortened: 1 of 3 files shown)


hdk5:
Remind me later: commentary extractor pulls "귀하의 브라우저는 html5 video를 지원하지 않습니다." from some posts, which should not be there.

hdk5:
Another note: turns out, "data-orig" attribute is not always present. For example, https://arca.live/b/bluearchive/111191955 has a gif.

hdk5:
Fixed the issues above, plus made the extractor use the API instead of parsing the page.

PR #5917, opened 2025-02-18
sources: Xiaohongshu extractor fixes/tweaks

hdk5:
1. xsec_token query parameter is now required for posts
2. include artist id in post url whenever available
3. extract formatted artist commentary from page html, as opposed to unformatted commentary from json
4. some tests (A video post, A post with /spectrum/ image URLs) still fail, as i didn't bother to find new appropriate posts to test them.

### DIFF
--- a/app/logical/source/extractor/xiaohongshu.rb
+++ b/app/logical/source/extractor/xiaohongshu.rb
@@ -34,7 +34,7 @@ class Source::Extractor::Xiaohongshu < Source::Extractor
 
   def tags
     note["tagList"].to_a.pluck("name").map do |tag|
-      [tag, "https://www.xiaohongshu.com/search_result/?keyword=#{Danbooru::URL.escape(tag)}"]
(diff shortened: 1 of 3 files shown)


hdk5:
someone helped me with the remaining tests, now all of them pass at the moment

hdk5:
Also, i'm not really sure how long links with `xsec_token` parameter live. Still, it is better to have link that worked for some time than link that is dead from beginning, I think?

evazion:
This seems to work, but all the `xsec_token` URLs in the tests are expired now so I can't actually test this.

If you google around, there are other Xiaohongshu scrapers on Github that try to deal with this, but they're all in Chinese and I don't know if any of them have a working solution to this. Apparently there's a `https://edith.xiaohongshu.com/api/sns/web/short_url` endpoint that might be able to get you a working URL from a note id, but I couldn't figure it out:

https://github.com/NanmiCoder/MediaCrawler/blob/main/media_platform/xhs/client.py#L527-L560

PR #5944, opened 2025-05-11
lofter: add secondary profile urls

hdk5:


app/logical/source/extractor/lofter.rb
```diff
@@ -48,11 +56,15 @@ def tags
       end
 
       def display_name
-        page_json.dig(:postData, :data, :blogInfo, :blogNickName)&.strip
+        blog[:blogNickName]&.strip
       end
 
       def username
-        parsed_url.username || parsed_referer&.username
+        blog[:blogName] || parsed_url.username || parsed_referer&.username
```
evazion, line 63: There's a subtle difference here between 

```
blog[:blogName] || parsed_url.username || parsed_referer&.username
```

and

```
parsed_url.username || parsed_referer&.username || blog[:blogName]
```

If `blog[:blogName]` comes first, then the username from the API will be preferred, but if `blog[:blogName]` comes last, then the username in the URL will be preferred. This makes a difference if the artist has changed their name. For example, suppose you're fetching source data for `https://mukaend.lofter.com/post/1edceeb0_2b7198f54`, but the artist has changed their name to `https://ranolll.lofter.com`. Then we'll return `https://ranolll.lofter.com` as the profile URL. If that URL isn't in the artist entry yet, then artist finding will fail.

It probably makes more sense overall for extractors to always return current profile URLs, not dead URLs, but it has consequences for artist finding and it's something we're not always consistent about.

PR #5947, opened 2025-05-12
artist urls: factor out `secondary_url?` method

hdk5:


hdk5:
Any issues here?

evazion:
I don't think the name `secondary_url` is clear enough here. That might have to change at some point. We deal with many different types of urls, but this is specifically about profile urls.

We need a better way of classifying urls. We need some kind of methods like:

- `category` - `:image`, `:page`, or `:profile`
- `subcategory` - `image_type || page_type || profile_type`
- `image_type` - `:image_sample`, `:full_size`, `:profile_picture`, `:profile_background`, `:profile_banner`, etc
- `page_type` - For Pixiv: `:illust` or `:unlisted`. For NicoSeiga: `:illust`, `:manga`, `:video`. Etc
- `profile_type` - `:primary` or `:secondary`? `:user_id` or `:username`? I think some sites like NicoSeiga have multiple types of profile url with different user ids, so maybe this would be the user id type, and `primary` vs `secondary` would be a different thing.

This ties into #6301 and having some way for urls to specify which type of work_id it is, when a site has multiple types of id.

PR #5953, opened 2025-05-15
Modernize video and audio codecs

TypeA2:
Initial codec restrictions were introduced in a9d586e.

These changes are based on the web codec guides in the Mozilla docs:
* WebM: https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/Containers#webm
* MP4: https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/Containers#mpeg-4_mp4

For WebM, only `opus` and `vorbis` should be allowed. Currently it's possible to upload files with non-standard audio codecs simply by modifying FFmpeg (specifically, `libavformat/matroska.c`), as seen in https://danbooru.donmai.us/media_assets/25818543.

For MP4, AV1 WebM support was already enabled in 2d4b20e, and HEVC (H.265) should be enabled too. Vorbis audio should be _disabled_.  There are currently 2 MP4 assets with Vorbis audio, one of which is a post. Chrome plays these fine, but Firefox doesn't play the audio (which is fair, since it's not officially supported), see https://danbooru.donmai.us/media_assets/24656948.

Support for HEVC is generally good on browsers people actually use (https://caniuse.com/hevc), and AV1 video support is a bit better (https://caniuse.com/av1).


evazion:
We're going to need to generate H.264 MP4 sample videos for anything not supported by older browsers, and make it so the `<video>` tag falls back to the MP4 if the original video isn't supported by the browser. Something like this:

```html
<video>
  <source src="original.webm" type="video/webm; codecs=av01.0.08M.08.0.110.01.01.01.1"/>
  <source src="sample.mp4" type="video/mp4"/>
</video>
```

Generating samples will slow down uploads because it means we'll have to transcode video in the background. There's also the question of what encoding parameters to use.

The main reason these codecs weren't allowed was because I wasn't sure of the best to do all this.

https://jakearchibald.com/2022/html-codecs-parameter-for-av1/
https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/codecs_parameter#using_the_codecs_parameter

Issue #5251, opened 2022-09-17
Support undo after autocompletion


nottalulah:
It would be nice if you could hit ctrl-z to undo an autocomplete.

evazion:
I spent like a full day on this once and could never get it to work. As I recall the problem was that programmatically adding text to an input field doesn't add it the undo history. You have to either reimplement the entire input field in Javascript, or try to use tricks like [execCommand](https://developer.mozilla.org/en-US/docs/Web/API/document/execCommand#inserttext) to insert text, which I could never get to work right.

https://discord.com/channels/310432830138089472/310846683376517121/1020533982800126073

PR #5954, opened 2025-05-18
Autocomplete: Improve behavior when caret is in the middle of a term

ToksT:
Sometimes people make typos when writing tags. If you're trying to write `brown_eyes` you may write `brow_e` and only then notice your mistake:
image
So you move your caret back and fix the typo:
image
But now autocomplete doesn't see the `_e` part that comes after your cursor, so it's putting `brown_hair` above `brown_eyes` in the suggestions.
So you press down a couple times and manually select the second result instead, and then press tab so it completes `brown_eyes`:
image
Now there's some junk at the end that will mess up your search, since `_e` is not a real tag. So you have to manually delete that before hitting enter.

This pull request makes a simple change to autocomplete so that it pretends the cursor is at the end of the current term rather than being in the middle. This way, it sees `brown|_e` correctly suggests `brown_eyes` at the top instead of `brown_hair`, and hitting tab won't leave the `_e` in there:
image
I also made the same change for dtext autocomplete.
(Related tags already have similar logic to find the end of the current term, so that didn't need changing.)

Note: From what I've seen, this behavior is consistent with a lot of other programs that have autocomplete, such as Firefox's native autocomplete which is used when not overridden.

### DIFF
--- a/app/javascript/src/javascripts/autocomplete.js
+++ b/app/javascript/src/javascripts/autocomplete.js
@@ -60,8 +60,10 @@ Autocomplete.initialize_dtext_autocomplete = function($fields) {
       my: "left bottom"
     },
     source: async function(req, resp) {
-      var cursor = this.element.get(0).selectionStart;
-      let match = req.term.substring(0, cursor).match(/([ \r\n/"\\()[\]{}<>]|^)([@:])(\S*)$/);
+      let caret = this.element.get(0).selectionStart;
+      let term_after_caret = req.term.substring(caret).match(/\S*/)[0];
+      caret += term_after_caret.length;
+      let match = req.term.substring(0, caret).match(/([ \r\n/"\\()[\]{}<>]|^)([@:])(\S*)$/);
 
       let prefix = match?.[1];
       let type = match?.[2];
@@ -99,6 +101,8 @@ Autocomplete.initialize_tag_autocomplete = function() {
 Autocomplete.current_term = function($input) {
   let query = $input.get(0).value;
   let caret = $input.get(0).selectionStart;
+  let term_after_caret = query.substring(caret).match(/\S*/)[0];
+  caret += term_after_caret.length;
   let regexp = new RegExp(`^[-~(]*(${Autocomplete.tag_prefixes().join("|")})?`);
   let match = query.substring(0, caret).match(/\S*$/)[0].replace(regexp, "").toLowerCase();
   return match;
@@ -107,9 +111,13 @@ Autocomplete.current_term = function($input) {
 // Update the input field with the item currently focused in the
 // autocomplete menu, then position the caret just after the inserted completion.
 Autocomplete.insert_completion = function(input, completion) {
+  let caret = input.selectionStart;
+  let term_after_caret = input.value.substring(caret).match(/\S*/)[0];
+  caret += term_after_caret.length;
+
   // Trim all whitespace (tabs, spaces) except for line returns
-  var before_caret_text = input.value.substring(0, input.selectionStart).replace(/^[ \t]+|[ \t]+$/gm, "");
-  var after_caret_text = input.value.substring(input.selectionStart).replace(/^[ \t]+|[ \t]+$/gm, "");
+  var before_caret_text = input.value.substring(0, caret).replace(/^[ \t]+|[ \t]+$/gm, "");
+  var after_caret_text = input.value.substring(caret).replace(/^[ \t]+|[ \t]+$/gm, "");
 
   var regexp = new RegExp(`([-~(]*(?:${Autocomplete.tag_prefixes().join("|")})?)\\S+$`, "g");
   before_caret_text = before_caret_text.replace(regexp, "$1") + completion + " ";


ghost:
I just merged these changes into my local instance and this way of doing it is significantly worse to me. Deleting an extra _e is not much of an issue, you just hit ctrl+delete. This change makes correcting _previous_ terms significantly more annoying, because now you can end up in a situation like `blakc_hair|red_eyes`, where autocomplete will not pop up anything because it's trying to autocomplete `blakc_hairred_eyes` instead of the previous behavior of autocompleting `blakc_hair`. You now have to insert a space, press left to go to the previous word, and hope autocomplete works right (it doesn't, it autocompletes red_eyes instead, which means you have to also re-trigger autocomplete—this also compounds with another bug that makes autocomplete trigger on the word on the other side of a space, which means you're deleting the other tag if you complete it). Even if it did work right it would still be significantly more annoying than the previous behavior.

This commit also introduced multiple other bugs (or at least made previous bugs that did nothing start causing problems), see #5958. I'd rather just see it reverted.

ToksT:
> This change makes correcting _previous_ terms significantly more annoying, because now you can end up in a situation like `blakc_hair|red_eyes`,

How would you wind up in that specific situation though? If you typed `blakc_hair`, then a space, then `red_eyes` there wouldn't be an issue when you go back to fix it to `black_hair` because there would be a space between them. You would have to not only mistype `black_hair` but *also* forget the space between the two tags. Is that a common occurrence?

If you had typed `red_eyes` first, then repositioned your caret to the beginning and typed `blakc_hair` there and noticed the typo, then yes, that would put you in that situation which wouldn't be ideal. But if you hadn't repositioned your cursor to the beginning and instead typed `red_eyes blakc_hair|` in that order, then that would work fine.

When you press Q or E or whichever shortcut to focus a text field, your caret is placed at the end by default. So my thinking was that making editing from the end of a text field better at the cost of making editing from the beginning of a text field worse would be an okay tradeoff as I didn't think that people typed from the beginning. Do you type from the beginning?

ghost:
> > This change makes correcting _previous_ terms significantly more annoying, because now you can end up in a situation like `blakc_hair|red_eyes`,
> 
> How would you wind up in that specific situation though?

Ctrl-left after a term jumps to the beginning of the word rather than to before the space. Previously you could ctrl-left then ctrl-delete to start writing the term over, or ctrl-left then backspace to autocomplete the previous term (because left arrow doesn't trigger autocomplete correctly—in this case it would autocomplete red_eyes and if you hit tab it would overwrite blakc_hair with red_eyes, though that problem existed even before this change). Even if the latter bug were fixed, it would still prevent the former case, meaning you'd have to jump back, delete word back, jump forward, and make sure there's a space left of the caret before you can start typing again.

ToksT:
> Ctrl-left after a term jumps to the beginning of the word rather than to before the space. Previously you could ctrl-left then ctrl-delete to start writing the term over, [...] meaning you'd have to jump back, delete word back, jump forward, and make sure there's a space left of the caret before you can start typing again.

If I'm understanding you, this is pretty easy to fix by programmatically inserting an extra space after the caret when ctrl-deleting/ctrl-backspacing a word like that, similar to the change made in PR #5962.

The only issue I can see with this is that it would mean you can't undo the deletion of a tag anymore because browsers don't allow users to undo programmatic text edits made by javascript, just like how you already can't undo an autocomplete insertion: #5251.

Although looking into the execCommand trick evazion mentioned in that issue, it actually seems to work for me. I'll need to test it some more before creating a PR in case it introduces new bugs, but it may be possible to have full undo support for both deletions and autocomplete insertions.

> or ctrl-left then backspace to autocomplete the previous term (because left arrow doesn't trigger autocomplete correctly—in this case it would autocomplete red_eyes and if you hit tab it would overwrite blakc_hair with red_eyes, though that problem existed even before this change).

Are you referring to how autocomplete only updates when you type something and not when you move your selection to a different tag? That would be a simple change.

ghost:
> If I'm understanding you, this is pretty easy to fix by programmatically inserting an extra space after the caret when ctrl-deleting/ctrl-backspacing a word like that, similar to the change made in PR #5962.
> 
> The only issue I can see with this is that it would mean you can't undo the deletion of a tag anymore because browsers don't allow users to undo programmatic text edits made by javascript, just like how you already can't undo an autocomplete insertion: #5251.
> 
> Although looking into the execCommand trick evazion mentioned in that issue, it actually seems to work for me. I'll need to test it some more before creating a PR in case it introduces new bugs, but it may be possible to have full undo support for both deletions and autocomplete insertions.

I guess that could work, assuming the execCommand thing will work properly for all. You probably wouldn't want undo to only remove the inserted space.

Edit: Actually, looking at it, MDN seems to say execCommand is deprecated.

> Are you referring to how autocomplete only updates when you type something and not when you move your selection to a different tag? That would be a simple change.

I'm not sure if that issue is caused by this, but when you don't have autocomplete triggered and move the cursor to another term, it'll autocomplete the wrong term:

https://github.com/user-attachments/assets/a1f6113a-2438-4a3b-9e48-38cdae3ebfbb

ToksT:
> I guess that could work, assuming the execCommand thing will work properly for all. You probably wouldn't want undo to only remove the inserted space.

For autocomplete insertions, execCommand seems to allow undo to work exactly as expected. Programmatically deleting the original text, inserting the completion, and inserting the space are all done at once and undone at once.

For deletions, it seems to be undone in two steps instead of one, which is somewhat unexpected. First it undoes the programmatic insertion of the space, then it undoes the user's manual deletion of the tag with ctrl+delete/backspace, as these are considered separate actions. So these can both be undone, but the user would need to press ctrl+z twice as the first press would only undo the space, while the second press would undo the deletion itself.

The two undo steps could be combined into one by overriding the behavior of ctrl+delete/backspace entirely so that the tag is deleted programmatically as well. This would also allow more control over how those shortcuts work, like allowing entire tags to be deleted at once instead of stopping at a parenthesis. e.g. `fate_(series)` currently takes two ctrl+backspace presses to delete, but could be done in one if we're overriding the behavior of those shortcuts. This idea may be overcomplicating it though, it might introduce more subtle unexpected behaviors in some way.

> I'm not sure if that issue is caused by this, but when you don't have autocomplete triggered and move the cursor to another term, it'll autocomplete the wrong term:

Yeah, that can be fixed by triggering a new autocomplete search on "selectionchange".

PR #5964, opened 2025-05-23
Fix #5900: Fix Reddit extractor

### LINKED ISSUE #5900: Reddit uploads broken
Getting `Upload failed: (link) doesn't contain any images.` on all attempts to upload from reddit. Tried both /s/ links from the mobile app and the basic desktop links

TypeA2:
Fixes #5900 for the most part, using a different JSON endpoint instead of scraping HTML. Also slightly improved commentary parsing while fixing the tests.

One slight deviation is that posts with external images (see test `A reddit post with an external image` now return the linked URLs in `image_urls`. The old extractor returned `external-preview` Reddit URLs, but I could not get these to reliably return an actual image, while this one at least works for sites like Twitter.
Another option would be to return no URLs, which would be trivial since the API response contains a boolean for it.

### DIFF
--- a/app/logical/source/extractor/reddit.rb
+++ b/app/logical/source/extractor/reddit.rb
@@ -9,10 +9,19 @@ module Source
           [parsed_url.full_image_url]
         elsif parsed_url.image_url?
           [parsed_url.to_s]
-        elsif data.present?
+        elsif post_data.present?
+          # images += [data.dig("media", "content")].compact unless crosspost? || data.dig("media", "type") == "embed"
+          return [external_image] if external_image.present?
+          return [] if crosspost?
+
           images = []
-          images += [data.dig("media", "content")].compact unless crosspost? || data.dig("media", "type") == "embed"
-          images += ordered_gallery_images
+
+          if post_data.include?("preview") && post_data["is_reddit_media_domain"]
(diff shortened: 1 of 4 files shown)


evazion:
> One slight deviation is that posts with external images (see test `A reddit post with an external image` now return the linked URLs in `image_urls`. The old extractor returned `external-preview` Reddit URLs, but I could not get these to reliably return an actual image, while this one at least works for sites like Twitter.

This is difficult because on the one hand, if you're uploading https://www.reddit.com/r/yuri_jp/comments/1kis50l/七海_x_たきうみ/ you would think it should work because there's no indication it's an external image. On the other hand, this breaks one of our core assumptions about extractors, which is that they only return URLs belonging to the same site. If the Twitter extractor only returns Twitter URLs, then the Reddit extractor should only return Reddit URLs. Otherwise if Twitter URLs can belong to Reddit posts, then that makes things difficult when it comes to eventually saving source data.

To put it another way: if you upload that post, what should the source be? Should it be the Twitter image because the image originally came from Twitter, or should it be the Reddit post because the image was posted on Reddit? Ideally you would just upload it from the original source instead of Reddit.

This is why I think it's better for extractors not to return external images to avoid these difficulties.

Issue #5492, opened 2023-04-04
Feature: BUR previews


hdk5:
Currently, BUR edit page (i.e. `https://danbooru.donmai.us/bulk_update_requests/:id/edit`) has a "Preview" button that does absolutely nothing.

I believe it would be nice to be able to validate BURs and print them in the same format they are displayed on forum (with colored and/or underlined clickable links) before actually creating/editing them.

hdk5:
#5965 did not solve what i wanted, keeping it open

PR #5965, opened 2025-05-23
Bulk Update Requests: display better errors

nonamethanks:
This basically just adds `[[]]` and `{{}}` around each tag name in BUR/implication errors, so that they are automatically styled by the site. 

The errors then become like this:
image

Which makes it infinitely easier to parse the error message, especially in the case of large BURs.

Compare with the status quo:

image

This partially mitigates the problems in issue #5492.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -300,17 +300,17 @@ module ApplicationHelper
     end
   end
 
-  def edit_form_for(model, validate: false, error_notice: true, warning_notice: true, **options, &block)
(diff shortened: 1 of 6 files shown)


nonamethanks:
This pull adds a `formatted_errors` parameter to edit_form_for that defaults to false and is only active for the BUR form, because it seemed to be the easiest way to deal with this, but thinking about it it might not be a bad idea to also use this elsewhere.

Compare the artists form:

Currently:
image

If `formatted_errors: true`
image

PR #5984, opened 2025-05-29
Add sqs mock for local development

### LINKED ISSUE #5255: Is there a more updated version of the archives service, that current production danbooru is meant to work with?
I've previously been running (for a personal instance) the [archives service](https://github.com/danbooru/archives) for versioning by running it alongside SQS and the `danbooru:production` container (etc.), with a shared database (created by the archives service) pointed to using `POSTGRES_DB` on the archives side, and `ARCHIVES_DATABASE_URL` on the danbooru side.

However, this stopped working after 1d2bac7b95d516b79e44b3e95427a41ed5c4063c as danbooru no longer provides `updater_ip_addr` over SQS, and while the migrations have updated danbooru's database, there's no relevant migration for the archives service so updates fails with a null for a non-nullable field.

I noticed then that the danbooru database also has `post_versions` etc. and that [app/models/post_version.rb](https://github.com/danbooru/danbooru/blob/master/app/models/post_version.rb) (among others) have
```ruby
  def self.database_url
    ENV["ARCHIVE_DATABASE_URL"] || ENV["DATABASE_URL"]
  end
```
hinting that perhaps I should be pointing both at the main danbooru database rather than a separate archives database. However, when I try that, the archives service deals badly with the main danbooru database's schema, now complaining that `created_at` is null but is a non-nullable field.

My questions are:
1. Should I be pointing both at the main danbooru database (generated by danbooru), or a separate archives database (generated by the archives service)? If the latter, are there recent relevant migrations?
2. If I should be pointing both at the main danbooru database, is there an updated version of the archives service that works with the current schema?

Thank you, and sorry for the trouble. Please let me know if there's more information that would be helpful for me to provide.

nonamethanks:
This pull adds a local SQS mock that makes post versions available locally and to codespaces for development and testing.

Still a WIP, I need to add it to the non-dev docker compose file as well and figure out the various options.

This pull also contains migrations that align the schema to the one used by danbooru in production.

Fixes #5255.

### DIFF
--- a/db/migrate/20250601164355_align_pool_versions_to_prod.rb
+++ b/db/migrate/20250601164355_align_pool_versions_to_prod.rb
@@ -1,0 +1,40 @@
+class AlignPoolVersionsToProd < ActiveRecord::Migration[7.1]
+  def up
+    change_column :pool_versions, :id, :integer
+    change_column_null :pool_versions, :created_at, true
+    change_column_null :pool_versions, :updated_at, true
+    change_column :pool_versions, :pool_id, :integer
+    change_column :pool_versions, :updater_id, :integer, null: true
+
+    change_column_null :pool_versions, :name, true
+    change_column_default :pool_versions, :name, from: '', to: nil
+
+    change_column_null :pool_versions, :description, true
+    change_column_default :pool_versions, :description, from: '', to: nil
+
+    change_column_null :pool_versions, :category, true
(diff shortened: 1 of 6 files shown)


nonamethanks:
Should be ready now. I'll take another look at it tomorrow with fresh eyes to make sure everything's working correctly and merge it.

PR #5998, opened 2025-06-02
Xiohongshu: Use all credentials

TypeA2:
I hope you like giving them your phone number to get these.

From what I can tell they at least don't expire very quickly (~1week and counting)


TypeA2:
I've been testing this on Cosbooru for a few weeks now, it seems to ask me to reauthenticate my session using the mobile app roughly every week. The cookies stay the same, so it's jut a matter of opening any post and scanning the QR.

PR #6048, opened 2025-06-22
feat(artist): add `other_names=` method to prevent duplicate artist names

NekoAria:
- Override `other_names=` to automatically remove elements that exactly match the artist name
- Add comprehensive test cases covering various scenarios including case sensitivity
- Ensure other_names array stays clean without main artist name duplicates


ghost:
I didn't comment on this before but I think this should have been done in a before_save or before_validation callback instead of overriding the setter function. This leads to weird quirks like #6066.

Issue #4727, opened 2021-02-17
Color tags in tag/upload box according to their categories


nonamethanks:
I thought this was a very Javascript heavy thing, but apparently it's possible to do it with minimal code.

Some examples and implementations:
https://stackoverflow.com/a/37160584/11558993
https://jsfiddle.net/Rino_Raj/j4dcbovo/8/ (colors words at will)

It would look much nicer and make it easier to edit a post's tags visually.

bipface:
an alternative approach:
https://github.com/bipface/galkomponents/blob/master/readme.md

PR #6052, opened 2025-06-23
Various fixes and improvements for autocomplete

### LINKED ISSUE #5251: Support undo after autocompletion

### LINKED ISSUE #5958: pull #5954 has had unintended annoying consequences
https://github.com/danbooru/danbooru/pull/5954

This is a good change but it's created some new behaviour which was unintended and I really don't like.

When you're tagging a post and you go back to delete or change an earlier tag, previously you could simply keep on tagging from that point (in the middle of the tags) right away.  Because of this pull, now after you delete a tag your caret remains before the first character of the next tag, and you're stuck, unable to use autocorrect, and you have to go back to the end of the tags if you want to keep tagging.

Say I realize this post actually had 3 girls in it:
Image

I delete 1girl, and now the caret sits next to the first character of the next tag:
Image

Try to start typing a new tag and you're stuck with autocomplete trying to correct your new tag to be the next tag:
Image

There is no easy way to get back to tagging except to return the caret back to the bottom of the tags.  If you go back a space you're still caught by autocorrect and the surrounding tags.  It's a minor inconvenience but it's annoying.  You should be able to edit a tag in the middle of the field and just keep typing from there, like you could before.

Could be fixed by returning to the old functionality when you're at the start of a tag, i.e. there's whitespace left of the caret.

ToksT:
The first five commits are several bugfixes and quality-of-life improvements for autocomplete. Details are included in each commit. Between these five commits, I believe the following issues should now all be resolved (as well as some other issues I haven't seen reported anywhere):

* Fixes #5251
* Fixes #5958: When deleting a tag, the caret remains before the next tag, interfering with autocomplete for the next term you type
* Fixes the issue reported in the second half of [this comment](https://github.com/danbooru/danbooru/pull/5954#issuecomment-2954619440): Autocomplete results not updated when moving caret to a different term

The sixth commit is slightly different from the others, as it completely replaces the native undo/redo implementation with a custom one for tag input fields. My reasoning for this is that while the native browser undo works 'well enough' thanks to the `execCommand` trick used in several of the above commits, there are some minor issues with that trick which can be fixed by avoiding `execCommand` entirely and instead reimplementing undo from scratch so that it still works when changing an input's value.

Specifically:
* When you undo an `execCommand`, the browser's native undo is not smart enough to also revert the user's selection to its original placement. This means that the entire text field's contents get selected when you undo, which can be annoying, although it's better than undo not working at all.
* `execCommand` is technically deprecated. It currently works in all major browsers that I'm aware of, but they could decide to drop support for it at some point in the future.
* `execCommand` does not work in some minor browsers (such as Pale Moon).

I've tested my custom undo stack on Firefox and Chrome and it seems to work well. But if completely reimplementing the undo stack is considered overkill for those three problems mentioned above, then the sixth commit can be reverted and the other five will still work without it by using `execCommand` instead.

### DIFF
--- a/app/javascript/src/javascripts/autocomplete.js
+++ b/app/javascript/src/javascripts/autocomplete.js
@@ -7,6 +7,11 @@ let Autocomplete = {};
 Autocomplete.VERSION = 2; // This should be bumped whenever the /autocomplete API changes in order to invalid client caches.
 Autocomplete.MAX_RESULTS = 20;
 
+Autocomplete.WHITESPACE = ' \\t';
+Autocomplete.WORD_SEPARATORS = Autocomplete.WHITESPACE + '_()\\[\\]{}<>`\'"-/;:,.?!';
+Autocomplete.PREV_WORD_REGEXP = new RegExp(`[^${Autocomplete.WORD_SEPARATORS}]+[${Autocomplete.WORD_SEPARATORS}]*$`);
+Autocomplete.NEXT_WORD_REGEXP = new RegExp(`^[${Autocomplete.WORD_SEPARATORS}]*[^${Autocomplete.WORD_SEPARATORS}]+`);
+
 Autocomplete.initialize_all = function() {
   $.widget("ui.autocomplete", $.ui.autocomplete, {
     options: {
@@ -87,9 +92,9 @@ Autocomplete.initialize_tag_autocomplete = function() {
         var before_caret_text = target.value.substring(0, caret);
         var after_caret_text = target.value.substring(caret);
         if (event.inputType == "deleteWordBackward") {
-          before_caret_text = before_caret_text.replace(/\S+[ \t]*$/, "");
+          before_caret_text = before_caret_text.replace(Autocomplete.PREV_WORD_REGEXP, "");
         } else if (event.inputType == "deleteWordForward") {
-          after_caret_text = after_caret_text.replace(/^[ \t]*\S+/, "");
+          after_caret_text = after_caret_text.replace(Autocomplete.NEXT_WORD_REGEXP, "");
         }
         if (after_caret_text.match(/^\S/)) {
           // There's a tag after the caret, so add a space between them so it doesn't interfere with autocomplete.
@@ -112,7 +117,7 @@ Autocomplete.initialize_tag_autocomplete = function() {
     }
     let caret = target.selectionStart;
     var before_caret_text = target.value.substring(0, caret);
-    let match = before_caret_text.match(/\S+[ \t]*$/);
+    let match = before_caret_text.match(Autocomplete.PREV_WORD_REGEXP);
     if (match) {
       target.selectionStart = target.selectionEnd = match.index;
     }
@@ -130,7 +135,7 @@ Autocomplete.initialize_tag_autocomplete = function() {
     let caret = target.selectionStart;
     var before_caret_text = target.value.substring(0, caret);
     var after_caret_text = target.value.substring(caret);
-    let match = after_caret_text.match(/^[ \t]*\S+/);
+    let match = after_caret_text.match(Autocomplete.NEXT_WORD_REGEXP);
     if (match) {
       target.selectionStart = target.selectionEnd = before_caret_text.length + match[0].length;
     }
--- a/app/javascript/src/javascripts/autocomplete.js
+++ b/app/javascript/src/javascripts/autocomplete.js
@@ -1,4 +1,5 @@
 import { isBeforeInputEventAvailable }  from './utility'
+import UndoStack from './undo_stack';
 
 let Autocomplete = {};
 
@@ -68,6 +69,8 @@ Autocomplete.initialize_tag_autocomplete = function() {
       resp(results);
     }
   });
+
+  UndoStack.initialize_fields($fields_multiple);
 
   if (isBeforeInputEventAvailable()) {
     $fields_multiple.on("beforeinput", function(e) {
--- a/app/javascript/src/javascripts/autocomplete.js
+++ b/app/javascript/src/javascripts/autocomplete.js
@@ -1,5 +1,6 @@
 import { isBeforeInputEventAvailable }  from './utility'
 import UndoStack from './undo_stack';
+import Utility from './utility';
 
 let Autocomplete = {};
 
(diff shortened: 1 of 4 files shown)


evazion:
> [Tag autocomplete fields: Override the default Ctrl+Backspace and Ctrl+Delete shortcuts](https://github.com/danbooru/danbooru/pull/6052/commits/05d3cf0b7a4442613fc1eac5cf1d7c38d152555b)
> Improve the behavior of Ctrl+Backspace (deleteWordBackward) and Ctrl+Delete (deleteWordForward). These two shortcuts will now always delete the entire tag before/after the caret.

I'm not sure about this. Overriding the native Ctrl+Backspace behavior could be frustrating to people who are used to it.

I could argue the opposite and say that Ctrl+Backspace should stop at all punctuation, including underscores. This would make it easier to delete parts of tags to fix mistakes. A common mistake is for people to choose the wrong tag in autocomplete, like `white_panties` when they meant `white_pants` or `white_pantyhose`. Being able to delete each word instead of the entire tag would make these types of mistakes easier to fix.

Another issue is that this makes Ctrl+Backspace inconsistent with Ctrl+Left and Ctrl+Right. Both of these things should have the same sense of what a word is.

It also makes Ctrl+Backspace in the tag editor inconsistent with Ctrl+Backspace in the DText editor. But giving the DText editor the same behavior would be even more undesirable because you often have things like `[[artoria_pendragon_(fate)|Artoria]]`, and you don't want to delete the whole thing.

> [Autocomplete: Update search results when selection changes](https://github.com/danbooru/danbooru/pull/6052/commits/64faeae73cb42858f6e40dd95f75ce44f97ce03a)
> Currently the autocomplete search results only update when the user types something and not when they simply move the caret somewhere else in the existing text. This can cause an issue where the results shown will become outdated if the user moves their caret to a different tag, and if the user then presses tab to complete the results, it will result in the wrong tag being overwritten.
> 
> This can be fixed by manually triggering an autocomplete search when the selection changes if the autocomplete menu is already open.

I think it would be better to close the autocomplete menu when the cursor moves outside the current word. I think autocomplete should only open in response to you typing something, not just when you move the cursor to another tag.

> [Implement custom undo stack for tag autocomplete](https://github.com/danbooru/danbooru/pull/6052/commits/fb0056b1d3c4d083f8d451553c115d380234b8e7)

Re: the custom undo stack, implementing our own undo is something I wanted to avoid, but I think it's probably necessary. I experimented with replacing the DText editor's `<textarea>` with a contenteditable `<div>` (which is necessary to be able to do things like syntax highlighting or coloring tags, cf #4727), but the problem you run into is that you can't rely on `execCommand` for reasons you mentioned, but if you don't use `execCommand` then undo doesn't work, so you have to implement undo yourself.

The problem with implementing our own undo is that it really needs to emulate native behavior as closely as possible, so it works the way people are used to, which I think will be hard.

One example of native behavior that's hard (or impossible) to emulate is that you can right click a `<textarea>` and select undo/redo from the context menu, but that only knows about events in the native undo history, so the options will be greyed out if you're not using the native undo history.

Implementation-wise, I think it would be better to keep an undo log instead of snapshots of editor state. Every time we perform an editing command, such as inserting text, record an undo event like `[{ action: "insertText", text: "abc", start: 42, end: 45 }]`. Then it can be undone by doing the opposite. 

I think this approach is necessary to emulate native undo behavior. The normal behavior is that when you type something, then backspace over it, then type something else, that will generate three separate undo events. That requires keeping track of each type of command and merging adjacent commands of the same type together, while letting different commands create new undo events.

evazion:
I should mention that I'm planning to get rid of the jQuery autocomplete widget at some point and replacing it with our own autocomplete component. We spend too much time fighting against it or trying to layer our own behavior on top of it. jQuery UI itself is in maintenance mode and is pretty much dead at this point, so we need to move off it sooner or later.

I also have an idea about factoring out a `PlaintextEditor` component from the `DTextEditor` component. It would basically be a wrapper around a `<textarea>` that's used everywhere where a `<textarea>` is used, and it would contain things like the undo history and all the non-DText specific code from the DText editor, such as the stuff for positioning autocomplete menus beneath the cursor. This would make things like #5994 (adding autocomplete to the BUR field) easier. Then we could build a TagEditor component on top of it, and that would make dealing with autocomplete in the tag editor a lot easier to deal with.

ToksT:
> I could argue the opposite and say that Ctrl+Backspace should stop at all punctuation, including underscores.

> Another issue is that this makes Ctrl+Backspace inconsistent with Ctrl+Left and Ctrl+Right. Both of these things should have the same sense of what a word is.

I hadn't thought about that before, but it makes sense considering Danbooru tags treat underscores as spaces. Pushed two commits that give Ctrl+Left/Right the same behavior as Ctrl+Backspace/Delete and make them both stop at underscores. It does seem useful for correcting individual words as you said.

> I think it would be better to close the autocomplete menu when the cursor moves outside the current word. I think autocomplete should only open in response to you typing something, not just when you move the cursor to another tag.

I'm not sure how we would detect if the cursor has moved to a different tag or not. I don't think we want to close the autocomplete menu when the cursor moves to another position within the same tag, right?

> One example of native behavior that's hard (or impossible) to emulate is that you can right click a `<textarea>` and select undo/redo from the context menu, but that only knows about events in the native undo history, so the options will be greyed out if you're not using the native undo history.

This had been bothering me. I did implement detection of the "historyUndo" and "historyRedo" input events, so if the user somehow managed to click those buttons in the context menu they would trigger the custom undo instead of the native undo. But as you mentioned, the buttons are greyed out because the browser isn't aware undo has been overridden by the website, so they can't be clicked in the first place.

If we were to use `execCommand`, this allows the "Undo" button to be enabled in the context menu, but "Redo" will still be permanently greyed out. I wasn't able to think of a way to fix the context menu "Redo" and didn't want them to be inconsistent.

> Implementation-wise, I think it would be better to keep an undo log instead of snapshots of editor state. Every time we perform an editing command, such as inserting text, record an undo event like `[{ action: "insertText", text: "abc", start: 42, end: 45 }]`. Then it can be undone by doing the opposite.
> 
> I think this approach is necessary to emulate native undo behavior. The normal behavior is that when you type something, then backspace over it, then type something else, that will generate three separate undo events. That requires keeping track of each type of command and merging adjacent commands of the same type together, while letting different commands create new undo events.

The reason I went with the snapshots approach is to keep the undo implementation simple as I was worried the complexity would start to balloon with a different approach.

You make a good point about native undo grouping event types together though, so I think I've come up with a good compromise: I pushed a commit that still uses snapshots like before, but includes the type of action in those snapshots, and only merges actions of the same type together. This appears to get the native behavior you describe while keeping the implementation just as simple (actually, it's slightly simpler than the original implementation now).

evazion:
> I'm not sure how we would detect if the cursor has moved to a different tag or not. I don't think we want to close the autocomplete menu when the cursor moves to another position within the same tag, right?

It would require tracking the previous cursor position and doing something like `Autocomplete.current_term(currentPosition) !== Autocomplete.current_term(oldPosition)` to see if the current tag has changed.

(This isn't quite right because it's possible for the same tag to appear multiple times if someone adds it multiple times, but it's probably good enough. Figuring out when we moved from the Nth tag to the Mth tag would be harder.)

>The reason I went with the snapshots approach is to keep the undo implementation simple as I was worried the complexity would start to balloon with a different approach.

It would be overkill if it were just for autocomplete purposes, but I'd like to use the undo stack for the DText editor too, since it has the same problems with `execCommand` that autocomplete has. But undo is used a lot more heavily in writing things like wiki pages and forum posts than in tag editing, so any problems with undo there will be a lot more noticeable.

PR #6053, opened 2025-06-23
Add pending posts link to contributor users profile page

hdk5:
The issue here is that I am used to click the upload limits on own profile page to quickly access own pending posts.
And now, after promotion, I can't do that for the posts that I occasionally send to the queue and have to search for `user:hdk5 status:pending` manually instead.

Also fix Posts/Uploads wording on the upload page.


ghost:
This just looks dumb in the more common case of contributors not uploading as pending. If you absolutely need such a thing it would be better to only show if the user actually has pending posts.

hdk5:
yeah, whatever

PR #6074, opened 2025-07-01
Another minor tweak to Ctrl+Right

ToksT:
Related to commit 35494be3c60e6540b2da847348dbf777857acb2f

> Now Ctrl+Right moves past spaces at the end of tags, which matches normal browser behavior.

I was confused by this statement as when I tested it on both Firefox and Chrome Ctrl+Right did not move past the final space. After some more testing, it would seem this behavior actually varies not on your browser but your operating system. On Windows it skips the final space in both browsers, but on Linux it doesn't. (Unsure about Mac.)

Furthermore, desktop GUI frameworks like Qt seem to have the Windows behavior even on Linux. So not skipping the space is something that specifically only browsers do, and only on Linux.

If we wanted to emulate native behavior as closely as possible we would need to detect the user's operating system and change the regex, but I'm not sure if this is necessary. I think the behavior of skipping the space makes more sense, and it's unclear why browsers conditionally disable this anyway.

Instead this PR is just further tweaking Ctrl+Right to be closer to the Windows behavior, as on master it is now currently a hybrid between the two. Ctrl+Right normally would only skip one run of separators and one run of non-separators on both platforms, and only the order of the two runs varies between platforms.

For example, if your caret was here:
`1girl| 1boy `
The current behavior of Danbooru's Ctrl+Right  override on master would place it here, skipping the first space, the word, and the final space:
`1girl 1boy |`
But the native browser behavior on Windows would place the caret here, moving only a single space, so that's what this new PR does:
`1girl |1boy `
For comparison, the native browser behavior on Linux would place the caret here, moving one space and one word, but not the final space, which is what my original PR did:
`1girl 1boy| `

### DIFF
--- a/app/javascript/src/javascripts/autocomplete.js
+++ b/app/javascript/src/javascripts/autocomplete.js
@@ -92,12 +92,13 @@ Autocomplete.initialize_tag_autocomplete = function() {
         let caret = target.selectionStart;
         var before_caret_text = target.value.substring(0, caret);
         var after_caret_text = target.value.substring(caret);
+        let orig_after_caret_text = after_caret_text;
         if (event.inputType == "deleteWordBackward") {
           before_caret_text = before_caret_text.replace(Autocomplete.PREV_WORD_REGEXP, "");
         } else if (event.inputType == "deleteWordForward") {
           after_caret_text = after_caret_text.replace(Autocomplete.NEXT_WORD_REGEXP, "");
         }
-        if (after_caret_text.match(/^\S/)) {
+        if (after_caret_text.match(/^\S/) && orig_after_caret_text.match(/^\S/)) {
           // There's a tag after the caret, so add a space between them so it doesn't interfere with autocomplete.
           after_caret_text = " " + after_caret_text;
         }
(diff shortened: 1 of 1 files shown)


ToksT:
This fix caused a new bug with Ctrl+Delete to appear due to the feature that automatically inserts a space after the caret, as Ctrl+Delete would only remove the single space and then that feature would add it back, causing Ctrl+Delete to appear to not do anything if the caret was positioned before a space. Fixed that bug, so now it actually seems to work.

evazion:
I think this will require some further tuning to get things working just right. One issue is that it's currently hard to delete words in the middle of a tag. For example, if you have `hand_on_own_hip`, you should be to easily delete `own` and replace it with `another's` to get `hand_on_another's_hip`. Or if you have `artoria_pendragon_(lancer)_(fate)`, you should be able to easily delete `lancer` and replace it with `swimsuit_ruler`.

Currently it's hard because it inserts spaces after deleting words, which breaks deleting words inside tags. But even if you get rid of that, there are still issues with it not working how you would like in some cases. For example, if you're at the start of `lancer` in `artoria_pendragon_(lancer)_(fate)`, and you try to Ctrl+Delete it hoping to get `artoria_pendragon_()_(fate)` so you can insert `swimsuit_ruler`, instead you get `artoria_pendragon_(|fate)`.

I experimented with different ways to solve this problem, but couldn't find something that felt good. One way is to make Ctrl+Left/Right move between the ends of words, instead of the beginnings of words. That way you can use Ctrl+Left to move to the end of a word and use Ctrl+Backspace to delete it. But then it feels weird because this is not how Ctrl+Left/Right normally works.

PR #6082, opened 2025-07-06
Artist Name Validation Enhancement

NekoAria:
**Description**
This PR adds comprehensive name format validation for artist entries to prevent ambiguous and low-quality naming practices.

**Changes**
- Added `validate_name_format` method to Artist model
- Implemented three validation rules:
  - Reject numeric-only names (e.g., "123456")
  - Reject names starting with "user_" prefix
  - Check for existing similar name patterns (`name_*`) and warn users

**Motivation**
- Reduces naming ambiguity and conflicts
- Prevents low-quality auto-generated usernames as artist names
- Proactively guides users toward more descriptive naming
- Saves maintainer time by preventing problematic entries upfront

### DIFF
--- a/app/models/artist.rb
+++ b/app/models/artist.rb
@@ -191,6 +191,36 @@ class Artist < ApplicationRecord
       if tag_alias.present?
         errors.add(:name, "'#{name}' is aliased to '#{tag_alias.consequent_name}'")
       end
+
+      validate_name_format
+    end
+
+    def validate_name_format
+      return unless name.present?
+
+      # Reject numeric-only names
+      if name.match?(/\A\d+\z/)
+        errors.add(:name, "cannot be a number. Use a descriptive name instead")
+      end
+
+      # Reject user_ prefix (auto-generated username pattern)
+      if name.match?(/\Auser_/i)
+        errors.add(:name, "cannot start with 'user_'. Use the artist's actual name")
+      end
+
+      # Check for existing similar name patterns (name_*)
+      escaped_name = ActiveRecord::Base.sanitize_sql_like(name)
+      similar_artists = Artist.where("name ILIKE ?", "#{escaped_name}\\_%")
+
+      # Exclude current artist during updates
+      similar_artists = similar_artists.where.not(id: self.id) if persisted?
+
+      similar_artists = similar_artists.limit(3)
+
+      if similar_artists.any?
+        similar_names = similar_artists.pluck(:name).join(", ")
+        errors.add(:name, "may conflict with existing artists: #{similar_names}. Use a more specific name")
+      end
     end
 
     def update_tag_category


ghost:
And what happens if a user literally uses "user" in their name? https://danbooru.donmai.us/artists/387456

Or one who exclusively uses numbers in their name?

Ignoring both of those points, it's still better to have an artist tag that has a bad name than not have an artist tag at all. Adding a barrier like this just makes it more likely that users won't bother to make artist tags, which makes it harder to fix later.

NekoAria:
> And what happens if a user literally uses "user" in their name? [danbooru.donmai.us/artists/387456](https://danbooru.donmai.us/artists/387456)
> 
> Or one who exclusively uses numbers in their name?
> 
> Ignoring both of those points, it's still better to have an artist tag that has a bad name than not have an artist tag at all. Adding a barrier like this just makes it more likely that users won't bother to make artist tags, which makes it harder to fix later.

In this specific situation, I think renaming it back to `weibo_7841564805` would make more sense.
Because `用户7841564805` is also a system auto-generated username. The Weibo account's UID is `7841564805`.

---

I don't know what your current Danbooru account is, nor do I know if you've been involved in maintaining artist entries.
At the very least, I think a more meaningful name is better than leaving a mess behind.
The specific details can be modified, but I disagree with your opinion on the general direction.

ghost:
> In this specific situation, I think renaming it back to `weibo_7841564805` would make more sense. Because `用户7841564805` is also a system auto-generated username. The Weibo account's UID is `7841564805`.

Surely you can understand why weibo_12345 is not better than user_12345, though? Why should such a validation only extend to pixiv and not other sites with similar behavior? (that one was just an example, but it's not too far-fetched to imagine someone has a username that begins with user_ which would be disallowed now, in a futile effort to "prevent a mess")

> I don't know what your current Danbooru account is, nor do I know if you've been involved in maintaining artist entries. 
> At the very least, I think a more meaningful name is better than leaving a mess behind.

Ideally, yes, but adding such strict validations will only create more of a mess. Many users are extremely lazy and won't bother making an artist entry at all if it's too hard to do so. The site should make it easy as possible to make one, even if that means it's not perfect, because not perfect is still better than nothing.

NekoAria:
You raise valid points about `user_something` legitimate cases - I didn't consider that properly.

However, I still think `weibo_12345` is more meaningful than `user_12345` because:
- It indicates the platform source (traceability)
- Provides context for maintainers

I'm planning to adjust the validation: change `user_*` prefix to warnings instead of errors. This way legitimate cases like `user_something` won't be blocked, but users will still get guidance about potentially problematic naming.

The goal isn't to be unnecessarily strict, but to guide users toward more maintainable naming when the current input looks auto-generated.

TypeA2:
+1 on what talulah said, this seems overly restrictive and _will_ just lead to people not creating artist entries because that's easier. This is already somewhat common, we don't need to encourage it. Also:
> Reject numeric-only names (e.g., "123456")

Who decided that is wholly undescriptive? Artists have weird names, being restrictive leads to people using weird other names that don't actually refer to the artist.

> Reject names starting with "user_" prefix

Same as abouve, this will have false positives. Even if you make it a warning instead of an error, it'll lead to people thinking they're doing something wrong even when they aren't.

> Check for existing similar name patterns (name_*) and warn users

This will also lead to a lot of false positives, https://danbooru.donmai.us/artists/328313 vs https://danbooru.donmai.us/artists/78340 for example. Even if it's just a warning, it's more false positives people will have to know they can ignore (they won't know they can ignore them...). Not to mention this only takes artist names into account (tags are qualified regardless of the category they clash with), and even if it did take all tags into account, sometimes it's not worth qualifying very popular tags (see https://danbooru.donmai.us/forum_topics/28579 for an example).

> The goal isn't to be unnecessarily strict, but to guide users toward more maintainable naming when the current input looks auto-generated.

Then I think you missed the mark because these seem incredibly strict. You're probably better off making it easier to find and fix these issues (where it can be handled by people who know what they're doing). Some extractors could probably be modified to give more meaningful results in many cases, and maybe even allow some extra search options, though these are just examples.

NekoAria:
> > Check for existing similar name patterns (name_*) and warn users
> 
> This will also lead to a lot of false positives, [danbooru.donmai.us/artists/328313](https://danbooru.donmai.us/artists/328313) vs [danbooru.donmai.us/artists/78340](https://danbooru.donmai.us/artists/78340) for example. Even if it's just a warning, it's more false positives people will have to know they can ignore (they won't know they can ignore them...). Not to mention this only takes artist names into account (tags are qualified regardless of the category they clash with), and even if it did take all tags into account, sometimes it's not worth qualifying very popular tags (see [danbooru.donmai.us/forum_topics/28579](https://danbooru.donmai.us/forum_topics/28579) for an example).

Other parts aside, I believe you didn't fully understand the scope of the changes I made to the module.  
Regarding the situation you mentioned, adding `kanade_sakura` when `kanade_(kanade_3344)` already exists will not cause any issues; only ambiguous names like `kanade` will trigger a warning.  
Additionally, this modification will only affect the artist tag, and is unrelated to the character tag and artist tag situation you mentioned.

> Some extractors could probably be modified to give more meaningful results in many cases, and maybe even allow some extra search options, though these are just examples.

For this part, I don't understand what you're trying to express. In the case of a user's Pixiv stacc being `user_abc123`, how else can it be optimized?

TypeA2:
> Other parts aside, I believe you didn't fully understand the scope of the changes I made to the module.
> Regarding the situation you mentioned, adding `kanade_sakura` when `kanade_(kanade_3344)` already exists will not cause any issues; only ambiguous names like `kanade` will trigger a warning.
> Additionally, this modification will only affect the artist tag, and is unrelated to the character tag and artist tag situation you mentioned.

I phrased my example poorly, what I meant is that if we have `kanade_sakura`, and if `kanade` isn't taken, then having a name such as `kanade` wouldn't be ambiguous per se (contrived example because this doesn't really work with normal names like this).


> Additionally, this modification will only affect the artist tag, and is unrelated to the character tag and artist tag situation you mentioned.

That's specifically the problem. If there's a copytag called `foo` then an arttag called `foo_(bar)` would also have to be taken into account when deciding whether there's _potential_ ambiguous other tags. In the example I linked, the artist tag `hoshino_ai` already existed, so `hoshino_ai_(oshi_no_ko)` was made. Later, `hoshino_ai` was qualified since, despite it being an artist tag, it's still technically ambiguous with `hoshino_ai_(oshi_no_ko)`, which is how we got `hoshino_ai_(nature3111)`.


> For this part, I don't understand what you're trying to express. In the case of a user's Pixiv stacc being `user_abc123`, how else can it be optimized?

I didn't say Pixiv per se, the usage of the stacc name is already as good as it's likely going to get, but I recall there being a few other extractors that give less than optimal results.

NekoAria:
> > Additionally, this modification will only affect the artist tag, and is unrelated to the character tag and artist tag situation you mentioned.
> 
> That's specifically the problem. If there's a copytag called `foo` then an arttag called `foo_(bar)` would also have to be taken into account when deciding whether there's _potential_ ambiguous other tags. In the example I linked, the artist tag `hoshino_ai` already existed, so `hoshino_ai_(oshi_no_ko)` was made. Later, `hoshino_ai` was qualified since, despite it being an artist tag, it's still technically ambiguous with `hoshino_ai_(oshi_no_ko)`, which is how we got `hoshino_ai_(nature3111)`.

I think there might be some confusion about the scope of this PR. The `hoshino_ai` example you mentioned addresses a fundamentally different problem than what this validation is trying to solve.

**This PR's scope:**
- **Input validation**: When a user is creating or updating an artist entry, check if the input looks problematic (auto-generated patterns, etc.)
- **Immediate feedback**: Warn users about potentially low-quality naming before they commit

**The `hoshino_ai` case:**
- **Post-creation evolution conflict**: The artist tag `hoshino_ai` was perfectly valid when created
- **Cross-category collision**: Issues arose later when character tags appeared
- **Dynamic ecosystem problem**: Requires ongoing maintenance and conflict resolution

These are two entirely different problem domains with different technical solutions.

TypeA2:
> **This PR's scope:**
> 
> * **Input validation**: When a user is creating or updating an artist entry, check if the input looks problematic (auto-generated patterns, etc.)
> * **Immediate feedback**: Warn users about potentially low-quality naming before they commit
> 
> **The `hoshino_ai` case:**
> 
> * **Post-creation evolution conflict**: The artist tag `hoshino_ai` was perfectly valid when created
> * **Cross-category collision**: Issues arose later when character tags appeared
> * **Dynamic ecosystem problem**: Requires ongoing maintenance and conflict resolution
> 
> These are two entirely different problem domains with different technical solutions.

The `hoshino_ai` case was also just a contrived example that at least relates to a real-world example where removing a qualifier (or alternatively, not having it in the first place) is valid. It also points out how, if you're going to check `<name>_*` to determine potential name conflicts, you also have to take non-artist tags into account, since those same conflicts exist there. If `char:foo_(bar)` exists, this code won't warn if you try to make an artist named `foo`, while if this validation is actually something we want, it should.

NekoAria:
> It also points out how, if you're going to check `<name>_*` to determine potential name conflicts, you also have to take non-artist tags into account, since those same conflicts exist there. If `char:foo_(bar)` exists, this code won't warn if you try to make an artist named `foo`, while if this validation is actually something we want, it should.

I understand your point about logical consistency, but extending this validation to all tag categories would fundamentally change the scope and feasibility of this feature.

**Why artist tags warrant special treatment:**

1. **Auto-generated name patterns**: The problems this PR addresses (numeric-only names, `user_*` prefixes) are predominantly found in artist tags due to how various extractors work
2. **Maintenance burden**: Artist tags with poor naming create disproportionate maintenance overhead compared to other tag types
3. **User creation patterns**: Users frequently create artist entries from extracted data, making input validation particularly valuable here

**Cross-category validation challenges:**

Implementing comprehensive cross-category conflict checking would require:
- Global tag namespace queries for every tag creation
- Complex conflict resolution logic across different tag semantics
- Significantly increased system complexity and performance overhead
- Potentially overwhelming users with warnings across all tag types

This would transform a focused improvement into a major architectural change affecting the entire tagging system.

**Maintaining reasonable scope:**

This PR aims to solve a specific, observable problem: low-quality artist naming patterns that create maintenance issues. While perfect logical consistency across all tag types would be ideal in theory, it's not practical to implement as part of this targeted improvement.

The validation can always be extended to other tag categories in future PRs if there's demonstrated need and community consensus, but I believe the current scope provides meaningful value without introducing systemic complexity.

If this approach doesn't seem worthwhile for the limited scope it addresses, I'm open to closing this PR entirely rather than expanding it beyond practical implementation boundaries.

nonamethanks:
As the others mentioned, it's better to have a bad artist tag than no artist tag. Bad names can be fixed afterwards, but for a lot of sites once the artwork is deleted it becomes impossible to find the original artist if an entry wasn't created while it was live.

PR #6092, opened 2025-07-12
sources: add Huajia extractor

asgembed:
A Chinese commission/art website similar to Mihuashi.


asgembed:
I decide to open this PR for review as I have no plans on credential support.

PR #6094, opened 2025-07-14
sources: add Mihuashi extractor

### LINKED ISSUE #5973: Add Mihuashi support
Chinese commission/art website.

Example profile:
https://www.mihuashi.com/profiles/29105?role=painter

Example page:
https://www.mihuashi.com/artworks/15092919

Example image URL:
https://image-assets.mihuashi.com/permanent/29105%7C-2024/05/29/16/FuE-9jWo-aPKXOq2KP2ZsR5Nxnqa.jpg

asgembed:
Fixes #5973

### DIFF
--- a/app/logical/source/extractor/mihuashi.rb
+++ b/app/logical/source/extractor/mihuashi.rb
@@ -1,0 +1,154 @@
+# frozen_string_literal: true
+
+# @see Source::URL::Mihuashi
+module Source
+  class Extractor
+    class Mihuashi < Source::Extractor
+      def image_urls
+        if parsed_url.full_image_url.present?
+          [parsed_url.full_image_url]
+        elsif parsed_url.image_url?
+          [parsed_url.to_s]
+        elsif work.present?
+          image_url = work[:url]
+          [Source::URL.parse(image_url).try(:full_image_url) || image_url]
+        elsif stall.present?
+          [
+            stall[:cover_url],
+            *stall[:example_images].to_a,
+          ].uniq.map do |url|
+            Source::URL.parse(url).try(:full_image_url) || url
+          end
+        elsif project.present?
+          project[:example_images].to_a.pluck(:url).map do |url|
+            Source::URL.parse(url).try(:full_image_url) || url
+          end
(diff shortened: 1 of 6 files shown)


test/unit/source/url/mihuashi_url_test.rb
```diff
@@ -32,8 +32,8 @@ class MihuashiUrlTest < ActiveSupport::TestCase
         )
 
         url_parser_should_work(
-          "https://www.mihuashi.com/profiles/29105?role=painter",
-          profile_url: "https://www.mihuashi.com/profiles/29105",
+          "https://www.mihuashi.com/profiles/29105",
+          profile_url: "https://www.mihuashi.com/profiles/29105?role=painter",
```
evazion, line 37: I don't see the difference between https://www.mihuashi.com/profiles/29105 and https://www.mihuashi.com/profiles/29105?role=painter. Using `?role=painter` for the profile URL means all existing artist URLs will have to be normalized to that form (see <https://danbooru.donmai.us/artist_urls?search[url_matches]=*mihuashi*>).

app/logical/source/extractor/mihuashi.rb
```diff
@@ -0,0 +1,154 @@
...
+      def account_url
+        "https://www.mihuashi.com/profiles/#{user_id}?role=painter" if user_id.present?
+      end
+
+      def profile_urls
+        [profile_url, account_url].compact
+      end
+
+      def display_name
+        username&.strip
```
evazion, line 51: `display_name` and `username` should not be the same thing. `display_name` is for sites like Twitter or Pixiv where users have two different names, a username that uniquely identifies the account and a display name that doesn't. If they only have a username, then `display_name` should be nil.

asgembed:
Since the website uses `https://www.mihuashi.com/profiles/*` by default, should it be the primary profile url?

evazion:
> Since the website uses `https://www.mihuashi.com/profiles/*` by default, should it be the primary profile url?

Yes, the primary URL should be the one that the site normally shows to the viewer.

PR #6095, opened 2025-07-16
mihuashi: use `/profiles/*` as primary profile url

asgembed:
https://github.com/danbooru/danbooru/issues/6094#issuecomment-3074903620


asgembed:
Wait, this is already changed?

PR #6103, opened 2025-07-23
ugoira: tweak webm sample conversion

hdk5:
* convert underlying .gif frames to .png, as otherwise ffmpeg tries to treat them as animations;
* remove the last frame duplication, as modern browsers and players already play the produced files correcly;
* force `yuv420p` output pixel format, as some frames may now have RGB format, which is not widely supported.


TypeA2:
`x264` supports a specific animation tuning mode (https://trac.ffmpeg.org/wiki/Encode/H.264). It might be worth doing a comparison, since the current vp9 samples have the issue of regularly not playing on iOS.

It also might be possible to convert the gif frames to a sample directly with the concat format (https://ffmpeg.org/ffmpeg-formats.html#concat), but apparently some ugoiras don't work well with this so it'd require some testing.

hdk5:
superseded by #6241

PR #6115, opened 2025-08-18
Replace wiki links with bulk update request links for tag aliases

NekoAria:
## Problem
Tag aliases on artist pages currently link to empty wiki pages, providing no useful information to users.

## Solution
Replace wiki links for tag aliases with links to approved bulk update requests. This allows users to:
- See when the alias was added
- View the actual bulk update request that created the alias
- Access meaningful information instead of empty wiki pages

## Changes
- Added `link_to_bulk_update_requests` helper method
- Updated tag alias display to use bulk update request links instead of wiki links

## Example
**Before:** `https://danbooru.donmai.us/wiki_pages/abc` (empty page)
**After:** `https://danbooru.donmai.us/bulk_update_requests?commit=Search&search[status]=approved&search[tags_include_any]=abc` (shows actual bulk update history)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -224,6 +224,10 @@ module ApplicationHelper
     link_to text, wiki_page_path(title), class: "wiki-link #{classes}", **options
   end
 
+  def link_to_bulk_update_requests(text, tag_name = text, classes: nil, **options)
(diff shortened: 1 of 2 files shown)


nonamethanks:
I'd say this would be unexpected behavior. All other tags' aliases point to the previous name's wiki, so we'd be making an exception only for artists, and for no clear reason. 



Consider these links that you find on the wiki sidebar when you click an alias: 
image
These are pretty useful to tell what happened to the previous iteration of a tag. We'd be making navigation to these links harder just for artist aliases, and for no benefit.

PR #6116, opened 2025-08-18
feat(views/artist): add create wiki link for artists without wiki pages

NekoAria:


nonamethanks:
This was removed on purpose to dissuade people from making pointless artist wikis.

NekoAria:
> This was removed on purpose to dissuade people from making pointless artist wikis.

I was wondering why I seemed to remember this link existing before.

However, I think we could add a restriction, showing it only to users with a specific level of permissions (e.g., Builder). What do you think?

Currently, I’m using a userscript to add this link.

nonamethanks:
Builders are not much smarter than other users, and there's been many times where we had to mass revert builder edits like these.

NekoAria:
> Builders are not much smarter than other users, and there's been many times where we had to mass revert builder edits like these.

What about contributors?

If you think only approvers or higher-level users are smart enough, it might be better to just keep things as they are.

Those who need it can continue using the userscript.

PR #6125, opened 2025-09-19
Mihuashi: image url improvement and activity artworks support

asgembed:
- Extracts profile id from image url.
- Legacy image url support for old artworks.
- Activity artworks support.


asgembed:
Just found that a Mihuashi user can be both painter and employer, like https://www.mihuashi.com/profiles/613986?role=painter. I think this should be handled in the URL parser.

nonamethanks:
This seems to be broken now @asgembed, the api returns:

```
{"error":"signature invalid","msg":"签名错误"}
```

Example: https://www.mihuashi.com/api/v1/artworks/13982141

asgembed:
> This seems to be broken now @asgembed, the api returns:
> 
> ```
> {"error":"signature invalid","msg":"签名错误"}
> ```
> 
> Example: https://www.mihuashi.com/api/v1/artworks/13982141

You missed a `/` at the end. https://www.mihuashi.com/api/v1/artworks/13982141/ works for me.

nonamethanks:
Ah I see, I mistook this error for another. Running on codespaces I actually get `"{\"message\":\"Your IP address is blocked.\",\"reason\":\"ip_denied\"}"`, probably a blanket ban on certain server providers. Works on another server though.

Hopefully danbooru's servers are not blocked.

PR #6133, opened 2025-10-15
-

Moebytes:
-

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,3 @@
 lib/dtext_rb/ext/dtext/Makefile
 lib/dtext_rb/ext/dtext/dtext.cpp
 .history
-.DS_Store
-*.plist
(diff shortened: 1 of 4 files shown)


ghost:
Doing it this way is a pretty bad idea because it's not really sustainable to have a separate column and operator for every different source site Danbooru supports. There are thousands of uploads a week from sites that aren't pixiv or Twitter. It's also not that bad with source, you can do `source:*<twitter id>*` without having to use their username. You should know this since your implementation did not require their username either. Danbooru also has its own IQDB at /iqdb_queries which isn't rate limited like SauceNAO is.

Moebytes:
If you want to support all sources then using generic site_id and site_name columns will be better. I don't actually have a major need for this anymore so I doubt I'll work on that.

PR #6138, opened 2025-11-02
Bilibili: fix commentary fetching when falling back to Bilibili API

asgembed:
`page` function (using HTML) may fail due to the CAPTCHA.


asgembed:
Replaced by #6139.

PR #6139, opened 2025-11-07
Bilibili: Some extractor improvements

asgembed:


hdk5:
In case you are interested in improving `article_commentary_desc` to support all possible article formatting, you can test on this article 

https://www.bilibili.com/opus/1127456471773085717

app/logical/source/extractor/bilibili.rb
```diff
@@ -22,86 +21,190 @@ def image_urls
...
 
-        urls
+        urls.compact.map { |u| Source::URL.parse(u).full_image_url || u }
+      end
+
+      memoize def post_image_urls
+        urls = []
+
+        urls += post_json.dig("modules", "module_dynamic", "major", "opus", "pics").to_a.pluck("url")
+        urls += post_json.dig("modules", "module_dynamic", "desc", "rich_text_nodes").to_a.select do |node|
```
hdk5: Not really sure if applicable here, but the order of extracted images should be the same as they appear in the post

nonamethanks:
At some point we should probably split the bilibili extractor into different strategies for different subdomains, because otherwise it becomes really hard to follow.

PR #6144, opened 2025-11-15
mode menu: include old post tag string in edit mode form

hdk5:
Fixes the error where edit mode would completely overwrite existing tags if another edit occurred after the page load.

### DIFF
--- a/app/views/posts/update.js.erb
+++ b/app/views/posts/update.js.erb
@@ -1,7 +1,9 @@
 <% if @post.valid? %>
-  var postElement = $("#post_<%= @post.id %>").get(0);
+  var $postElement = $("#post_<%= @post.id %>");
+  var postElement = $postElement.get(0);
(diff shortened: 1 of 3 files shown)


app/views/posts/update.js.erb
```diff
@@ -1,7 +1,9 @@
 <% if @post.valid? %>
-  var postElement = $("#post_<%= @post.id %>").get(0);
+  var $postElement = $("#post_<%= @post.id %>");
+  var postElement = $postElement.get(0);
   var html = "<%= j post_preview(@post, fit: :compact, show_deleted: true, show_votes: @show_votes, size: @preview_size) %>";
   Alpine.morph(postElement, html);
+  $postElement.removeData();
```
hdk5: this clears jQuery's internal data cache, otherwise it would load stale tags value when, for example, reopening the edit mode for the same post.

Issue #6038, opened 2025-06-15
make `parent:` accept media_asset/upload_media_asset id


hdk5:
with some special syntax like `parent:asset12345` or something
this should make posting variant sets a bit easier by not having to copy and paste the first in set post's id after it is made

PR #6149, opened 2025-12-14
posts: resolve parent_id from object URLs

hdk5:
post parent_id may now be assigned using post, media asset, upload, or upload media asset URLs.

Solves #6038.


nonamethanks:
I don't think this level of complexity is what we need, especially given the amount of parsing going on.

The easiest way to do this imo is to just allow something like `parent:asset:123`.

PR #6157, opened 2025-12-30
artist urls: add create/update/destroy actions

hdk5:
Editing artist URLs with API requires clients to reconstruct url_string themselves. Providing a REST-like interface should simplify that.
Notes:
1. `create` really acts like upsert/create_or_update;
2. `update` is intentionally made to not allow changing `url` field, to match how `url_string=` in `Artist` model works; 
3. there are no bulk create/update methods, or a replace method;
4. Version entry creation in controller feels kinda hacky.

### DIFF
--- a/app/controllers/artist_urls_controller.rb
+++ b/app/controllers/artist_urls_controller.rb
@@ -24,23 +24,18 @@ class ArtistURLsController < ApplicationController
     
     attrs = permitted_attributes(ArtistURL)
     @artist_url = @artist.urls.find_or_initialize_by(url: attrs[:url])
+    status = @artist_url.new_record? ? :created : :ok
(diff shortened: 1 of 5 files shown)


hdk5:
>> Version entry creation in controller feels kinda hacky.

> evazion said:
> `@artist.reload.create_version(true)` isn't safe, it can potentially fail and if it does the version won't be saved but the edit won't be rolled back
> someone else could also edit the artist in between the time you update the url and the time you create the version, i.e. `@artist.reload` could return a new version of the artist after someone else has edited it
> this is why it's better to do these things in model callbacks where everything is in a transaction by default

Artist version saving moved to ArtistURL callback for the reason above

PR #6161, opened 2026-01-03
BURs: add the convert command

### LINKED ISSUE #6146: make alias/rename support pools
Instead of having to use mass updates + nuke, we should be able to move the description of a pool to a tag along with its posts. It's not technically an alias, but there's probably no need for extra syntax just for this... Unless we want `convert pool:123 -> tag`. 

Though that's not a bad idea I guess, given that it could also apply to the opposite.

nonamethanks:
Fixes #6146.

Adds a `convert` command to convert pools to and from tags, that takes care of pool/wiki descriptions as well as moving posts.

With this, no more mass update && nuke && admin has to fix descriptions.

### DIFF
--- a/app/logical/bulk_update_request_processor.rb
+++ b/app/logical/bulk_update_request_processor.rb
@@ -384,9 +384,6 @@ class BulkUpdateRequestProcessor
     else
       convert_pool_to_tag(tag_or_pool1, tag_or_pool2)
     end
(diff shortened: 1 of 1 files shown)


nonamethanks:
TODO: add order:id_asc to tag -> pool so that we have some semblance of order.

PR #6179, opened 2026-01-15
Dmails: send the entire comment/forum post when a user is mentioned

### LINKED ISSUE #5717: Mentioning a user in a message with a quote breaks formatting in the dmail
[Example](https://testbooru.donmai.us/dmails/132?key=eyJfcmFpbHMiOnsiZGF0YSI6MTMyLCJwdXIiOiJkbWFpbF9saW5rIn19--8a6e4d7f46dc8900d8ecefa40f888c5265009c2f66b14bee6395b00821277537)

It looks like for some reason it strips out one of the opening [quote]s so it doesn't keep the main comment in a quote and has a unmatched closing element.

nonamethanks:
The previous behavior was to send an excerpt, but this is confusing because it's not clear that it's an excerpt unless you go open the post/comment itself and compare it with the dmail.

This also makes it so that if you have email notifications you can read the whole post via email.

Fixes #5717.

### DIFF
--- a/app/logical/concerns/mentionable.rb
+++ b/app/logical/concerns/mentionable.rb
@@ -27,7 +27,7 @@ module Mentionable
   def queue_mention_messages
     message_field = self.class.mentionable_option(:message_field)
     return if !send(:saved_change_to_attribute?, message_field)
-    return if self.skip_mention_notifications
+    return if skip_mention_notifications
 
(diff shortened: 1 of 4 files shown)


nonamethanks:
See https://github.com/danbooru/danbooru/issues/5717#issuecomment-3758895173

PR #6180, opened 2026-01-15
Dmails: don't show the "Respond" button for system dmails

### LINKED ISSUE #5411: Disallow users sending dmails to themselves/DanbooruBot
I've seen some people dmailing DanbooruBot trying to reply to the dmails it sends for mentions in comments/forum posts by mistake. I've also seen people dmail themselves trying to reply to someone else. No idea how they accomplished it but it has happened. Given there is little reason to message yourself and no reason to message the bot, I'd suggest disallowing it as it might help prevent confusion/lost replies.

A suggestion was given in Discord to have a flag on the users table for whether a user is capable of receiving dmails or not. That way it would be able to extend to non-DanbooruBot bots, but I don't know if it is worth implementing.

nonamethanks:
Fixes #5411.

This is to prevent users from unknowingly responding to the bot when they receive an automated dmail such as a feedback notice or mention notice.
This change is mostly cosmetic, users can still send dmails to DanbooruBot as a way to keep notes, which was one of the concerns in the linked issue.

### DIFF
--- a/test/functional/dmails_controller_test.rb
+++ b/test/functional/dmails_controller_test.rb
@@ -1,11 +1,12 @@
-require 'test_helper'
+require "test_helper"
 
 class DmailsControllerTest < ActionDispatch::IntegrationTest
   context "The dmails controller" do
     setup do
-      @user = create(:user, id: 999, unread_dmail_count: 1)
+      @user = create(:member_user, id: 999, unread_dmail_count: 1)
       @unrelated_user = create(:moderator_user, id: 1000, name: "reimu")
       @dmail = create(:dmail, owner: @user, from: @user)
+      @system_dmail = create(:dmail, owner: @user, from: User.system)
     end
 
     teardown do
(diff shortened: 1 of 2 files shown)


test/functional/dmails_controller_test.rb
```diff
@@ -72,7 +73,7 @@ class DmailsControllerTest < ActionDispatch::IntegrationTest
             create(:ban, user: @user)
           end
 
-          should respond_to_search({}).with { [@received_dmail, @dmail] }
+          should(respond_to_search({}).with { [@received_dmail, @dmail] })
```
nottalulah: Mistakenly-added parentheses?

nonamethanks: It's actually what the linter wants, but I keep forgetting to remove them after they're autoadded... 
https://docs.rubocop.org/rubocop/cops_lint.html#lintambiguousblockassociation
Anyway, fixed

Issue #5832, opened 2024-09-29
Show BUR controls directly into the forum post


nonamethanks:
We already have a really nice form in the BUR page: 
![image](https://github.com/user-attachments/assets/2a8a1f3b-de8f-4c25-b53d-0006fdc2886c)

We should just display it embedded in the forum post so that BURs can be processed without leaving the thread.

ngl, after a few years of being the primary handler of BURs on the site I've kinda gotten tired of having to go back and forth to make sure I'm approving the right BUR people are talking about. Having a button to press right in the forums would be a godsent.

This also make sure people know they can edit their own BURs, because I can't track how many times someone didn't even know they could do that.

nonamethanks:
fixed in pull #6186

PR #6186, opened 2026-01-19
BURs: show controls on forum post dropdown menu

### LINKED ISSUE #5214: Add "edit BUR" link to forum posts containing BURs
Sometimes I see superfluous BURs being made because the user was not aware they could just edit their initial BURs (something like changing a rename to an alias). It's not super intuitive that you have to click the BUR link to edit it since it looks like part of the forum post. Having an easier to access link could be helpful.

nonamethanks:
Fixes #5214, #5832.

firefox_JL5GQff5tG


ghost:
#5214 is also open where evazion mentioned having the "edit forum post" button let you edit the BUR inline.

PR #6189, opened 2026-01-19
Avoid clobbering commentary when fetching source data.

### LINKED ISSUE #5516: Manually added commentary is removed if rating is not selected
To replicate:

1. Go to upload an image
2. Manually add text into the commentary field
3. Post without selecting commentary
4. Page will reload with the usual "rating not selected" notification
5. Commentary field will be empty if it was blank before. If there was text in the commentary field from the bookmarklet, then only that text will be there after reloading.

Source links that were manually changed are not affected. This is mainly an issue when uploading from unsupported sites (Instagram, Hoyolab, etc.) where the commentary has to be added manually.

nottalulah:
Fixes #5516.

Potential caveat: if the commentary is fetched/inputted and you enter a different source to fetch again it won't be overwritten.

### DIFF
--- a/app/views/sources/show.js.erb
+++ b/app/views/sources/show.js.erb
@@ -5,7 +5,13 @@ Danbooru.RelatedTag.update_selected();
 
 if ($("#c-uploads #a-show").length) {
   Danbooru.Utility.alpineReady(() => {
-    $("#post_artist_commentary_original_title").val(<%= raw @source.dtext_artist_commentary_title.to_json %>);
-    $(".post_artist_commentary_original_description .dtext-editor").get(0).editor.dtext = <%= raw @source.dtext_artist_commentary_desc.to_json %>;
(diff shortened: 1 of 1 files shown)


nonamethanks:
That's a pretty big caveat. We're trading one bug for the other. 

You're not supposed to edit the original commentary, and in the future the ability to do so will be removed, so I think the current "bug" is only a bug in the case where the source is not returning any commentary. I've added a different fix in 4985c809340c33dc1501bbb046dc94a45ce3cf87.

ghost:
I figured the caveat was unimportant given it would only be relevant if the original source had commentary but was for some reason the wrong source, which seems an uncommon problem.
Additionally, there's always the problem of extractors not working (like xhs) or uploading from an unsupported site that would require commentary/etc be manually editable, or it risks losing the data entirely.

PR #6190, opened 2026-01-19
Fix #5368: Use /i/user/:id for Twitter user ID links

### LINKED ISSUE #5368: Use /i/user/:id for Twitter user ID links
[From the forums.](https://danbooru.donmai.us/forum_posts/227639)

Using /i/user/:id instead of /intent/user?user_id=:id allows visiting the link without triggering a "follow user" popup. If this change is made a script would have to be ran to update existing URLs.

nottalulah:
Fixes #5368.


evazion:
Doing this requires changing every artist URL to the new format for the artist finder to work, which is painful because it amounts to 418k artist edits. If we're going to do this we might as well rip off the bandaid and switch to `x.com` at the same time to avoid having to do another 418k edits.

https://danbooru.donmai.us/artist_urls?search[url_matches]=*twitter.com%2Fintent*&limit=1000

ghost:
I tested the artist finder on my local instance in the opposite direction, i.e. looking up /i/user/123 to find /intent/user?user_id=123, which seemed to work. Nonetheless, I opened #6191 for the domain change.

PR #6191, opened 2026-01-20
Replace twitter.com urls with x.com

nottalulah:
cf #6190

This commit replaces twitter.com normalization with x.com links in artist URLs, post sources, and commentary entities. I left it alone in some other places (Twitter link in footer, for example) for now. I'm not sure if the favicon should be replaced, so I left the Twitter one.

Fixes https://github.com/danbooru/danbooru/issues/5368.


nonamethanks:
We should change the logo too if we're doing this.

ghost:
Added.

nonamethanks:
@evazion do we also need a migrate script to convert all sources/artist urls?

evazion:
Yes, see `script/fixes/106_normalize_artist_urls.rb` for example for converting artist urls. Post sources don't need to be converted.

ghost:
I added [this script](https://github.com/danbooru/danbooru/pull/6191/changes/829234d3c79b5d68a46d1686ae87337d942b7d33#diff-4b10335df47ef35318c7d17a18db87a13235b8130ccbbcef61924fbfb85d4d13), hopefully it is sufficient.

BrokenEagle:
> evazion (<https://github.com/danbooru/danbooru/pull/6191#issuecomment-3800563660>):
> Post sources don't need to be converted.

They don't need to be, but should they? It does cause some hardship for those searching through posts, as they have to check both `source:https://twitter.com/` and `source:https://x.com/` . If this isn't something to be done by Danbooru, then should the users bot this out to ensure conformity with post sources? That would inevitably generate additional post versions though.

Another aspect of this is artist commentaries. Should Twitter links in artist commentaries be updated as well? This isn't as important as the prior item, since the data is unstructured and isn't really that searchable.

ghost:
That's already sort of an issue. There are 61k posts with x.com sources right now, because people will upload from the image url then change the source manually, or from disk, etc. Especially when the extractor was broken due to rate-limiting.

PR #6196, opened 2026-01-20
Github: skip extractor test

### LINKED ISSUE #6195: Skip or restrict extractor tests on github
We shouldn't really be stress-testing these, plus the datacenter IPs are often blocked, plus they often fail due to flaky connections, plus we tend to only notice that these are failing when our users report they're broken anyway.

Removing them will also make sure we can focus on the rest of the failing tests, which shouldn't actually fail.

nonamethanks:
These tests are very prone to failing on github, and they're a bad idea to run repeatedly, for various reasons:

* github runners use datacenter IPs, which are frequently blocked
* they create noise and tend to make us ignore other failed tests
* the tests only run on new commits, so we only notice when a strategy breaks because our users report it

This commit adds an environment variable, DANBOORU_SKIP_EXTRACTOR_TESTS, that when present simply causes all extractor tests to skip. URL parsing tests still run because they don't make external requests.

Fixes #6195

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -55,6 +55,7 @@ jobs:
       DATABASE_URL: postgresql://danbooru:danbooru@postgres/danbooru
       DANBOORU_SECRET_KEY_BASE: 1234
       DANBOORU_AWS_SQS_ENABLED: false
+      DANBOORU_SKIP_EXTRACTOR_TESTS: true
       DANBOORU_PIXIV_PHPSESSID: ${{ secrets.DANBOORU_PIXIV_PHPSESSID }}
       DANBOORU_NIJIE_LOGIN: ${{ secrets.DANBOORU_NIJIE_LOGIN }}
       DANBOORU_NIJIE_PASSWORD: ${{ secrets.DANBOORU_NIJIE_PASSWORD }}


ghost:
It might be worth considering something like [vcr](https://github.com/vcr/vcr) instead of skipping them completely. That way we could still test for regressions (it also avoids the headache of some breakage like a username change or post replacement on e621). Of course the downside is if the site changes its API the tests won't catch it, but they usually get caught by users before that anyway.

evazion:
> It might be worth considering something like [vcr](https://github.com/vcr/vcr) instead of skipping them completely. That way we could still test for regressions (it also avoids the headache of some breakage like a username change or post replacement on e621). Of course the downside is if the site changes its API the tests won't catch it, but they usually get caught by users before that anyway.

We used VCR in the past but it was a real pain. It led to a bunch of huge files in the repo since every request and file downloaded gets saved, it was hard to keep secrets like cookies, passwords, and IPs out of the recorded requests (VCR has ways to deal with this but it's easy to mess up and not realize), there were issues with VCR being sensitive to the exact order of requests, and there were potential merge conflicts if multiple people were working on extractors at the same time and they both needed to update the same cached requests.

PR #6204, opened 2026-01-28
Avoid double-quoting Time objects in HTML data attributes.

### LINKED ISSUE #5005: Export time attributes as HTML attributes
One use case that is immediately applicable would be the post show page.  The post has the attribute `last_noted_at` , however this information is currently only available through the API.  This information could be used to tell a translator when the last time someone has done notes on a post, so that they don't translate at the same time someone else has just started translating. So as to not force an unnecessary network call in cases such as this, it would be useful if this information was on the `<body>` element for show pages, and the `<tr>` element for index pages.

https://github.com/danbooru/danbooru/blob/a7dc05ce63bbf5129df8ea4517f11db76d335042/app/policies/application_policy.rb#L79-L85

The best format for the time values would be the standard ISO8601 strings that the site already uses for the JSON/XML, normalized to the GMT time zone.

nottalulah:
Call strftime on Time objects so that Rails doesn't call to_json on them, which causes them to have two sets of quotation marks. I used strftime instead of to_s or inspect, because to_s loses subsecond precision, and inspect can't be parsed automatically by JavaScript.

Old: `"2026-01-28T09:17:33.064+02:00"` (note the quotes)
New: `2026-01-28T09:17:33.064312+02:00`

Fixes #5005, but this is a breaking change for scripts that currently use the incorrectly double-quoted version.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -372,7 +372,9 @@ module ApplicationHelper
         value = record[attr]
       end
 
-      if value.nil?
+      if value.is_a?(Time)
+        value = value.strftime("%Y-%m-%dT%H:%M:%S.%6N%:z")
+      elsif value.nil?
         value = "null"
       end
 


BrokenEagle:
> Fixes 5005, but this is a breaking change for scripts that currently use the incorrectly double-quoted version.

I came up with my own `toTimestamp` function a while back due to this, which accounts for any amount of quotes.

```javascript
Utility.toTimeStamp = function (time_value) {
    while (typeof time_value === 'string') {
        var tmp;
        try {
            tmp = JSON.parse(time_value);
        } catch(e) {
            break;
        }
        time_value = tmp;
    }
    return (typeof time_value === 'string' ? new Date(time_value).getTime() : time_value);
};
```

PR #6205, opened 2026-01-28
Fix blacklisted post count on post page.

### LINKED ISSUE #6085: blacklist shows wrong count for posts with parent
If the blacklisted post has a parent the blacklisted items count on the sidebar is off by one.
Example:
blacklist = furry -rating:g
https://danbooru.donmai.us/posts/9513845 child post blacklisted parent post not blacklisted, shows as 2 should be 1
https://danbooru.donmai.us/posts/9480011 child and parent posts blacklisted, shows as 3 should be 2 (this also happens with the parent post)

nottalulah:
The post itself and the thumbnails in the parent/child bar are counted separately, so count unique post ids instead. I'm unsure about using `Alpine.raw` but `blacklist.posts` is a set which isn't compatible with Alpine's proxies, so calling `.map` on it throws an error. It worked in my testing, anyway.

Fixes #6085.

### DIFF
--- a/app/components/blacklist_component/blacklist_component.html.erb
+++ b/app/components/blacklist_component/blacklist_component.html.erb
@@ -11,7 +11,7 @@
       Blacklisted
     </label>
 
-    <span class="whitespace-nowrap post-count" x-text="blacklist.blacklistedPosts.length"></span>
+    <span class="whitespace-nowrap post-count" x-text="blacklist.blacklistedPostCount"></span>
     <span class="flex items-center flex-grow-1 justify-end gap-1">
       <%= render PopupMenuComponent.new do |menu| %>
         <% menu.with_item do %>
@@ -62,7 +62,7 @@
            x-show="rule.visible">
         <input type="checkbox" class="toggle-switch" x-model="rule.enabled">
         <%= link_to rule, posts_path(tags: rule), title: rule, class: "truncate", tabindex: "-1", "x-bind:class": "{ 'inactive-link': !rule.enabled }", "x-on:click.prevent": "rule.toggle()" %>
-        <span class="whitespace-nowrap post-count" x-text="rule.posts.size"></span>
+        <span class="whitespace-nowrap post-count" x-text="Alpine.raw(rule.posts).map(post => post.post.dataset.id).size"></span>
       </div>
     <% end %>
   <% end %>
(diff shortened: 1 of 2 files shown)


nonamethanks:
Something like `[...rule.posts].map(post => post.post.dataset.id).size` also works I guess (spread operator), but it might just be a matter of style at this point. At least I'm not aware of downsides to using Alpine.raw.

Would older browsers have a problem with this? I guess not since we're already using Alpine.* on other places.

ghost:
`[...rule.posts].map(post => post.post.dataset.id).size` turns it into an array which would mean we have the same problem as now, where duplicate IDs slip in. I guess an alternative to Alpine.raw would be to do `new Set(rule.posts)` which should work since it's not proxied. It would be slightly less efficient (having to construct a new set), but avoid potential reactivity issues. But I don't know whether those issues are even possible to arise in this situation.

nonamethanks:
Ok, I don't understand something in this pull then. Isn't rule.posts already a set? 

https://github.com/danbooru/danbooru/blob/9a4fc5656803662a14c93627266b4959e35a3cb7/app/javascript/src/javascripts/blacklists.js#L192

Changing line 65 back to rule.posts.size still seems to show correct blacklist counts.

ghost:
I'm not sure where you tested, but the problem in the linked issue is that if you have a post in a parent/child set, and the current post is blacklisted, the blacklist counts are off by one. 
image

There are only two images here, but it says "furry -rating:g" has three entries. This is because it counts both the thumbnail in the parent/child bar and the actual full-size image as separate entries, given the items in the set are the HTML elements themselves. To solve this, I want to do `posts.map(post => post.post.dataset.id)`, because both the thumbnail and full-size image have the post id on their dataset.

The problem is `this.posts` is not actually a set when it's used by the blacklist template, because Alpine has wrapped it in a Proxy to handle reactivity. But sets are not compatible with proxies (see https://stackoverflow.com/a/43947472), which means calling `this.posts.map` in the handler throws `TypeError: has method called on incompatible Proxy`. For whatever reason, passing the proxy to `new Set` doesn't throw that error. Using `Alpine.raw` to get the internal set out of the proxy also works.

nonamethanks:
https://ptb.discord.com/channels/310432830138089472/310846683376517121/1467544286231920650 relevant discussion on discord.

PR #6210, opened 2026-01-31
Fix #5716: one-hour drawing challenge tag not suggested in TL tags

### LINKED ISSUE #5716: one-hour drawing challenge tag not suggested in TL tags
Twitter extractor cuts off the 版深夜の真剣お絵かき60分一本勝負 part of one-hour drawing challenge hashtags (intentionally), which makes it suggest the base tag but means one-hour drawing challenge will never be suggested in TL tags. アークナイツ版深夜の真剣お絵かき60分一本勝負 should suggest both "arknights" (アークナイツ is in its wiki), and one-hour drawing challenge (アークナイツ版深夜の真剣お絵かき60分一本勝負 is in its wiki). Currently it only suggests arknights.

nottalulah:
The Twitter extractor's normalize_tag function will now try to match the full tag, the common suffix, and the base tag, instead of just the base tag. For example, in `アークナイツ版深夜の真剣お絵かき60分一本勝負`, it will match `アークナイツ`, `版深夜の真剣お絵かき60分一本勝負`, and `アークナイツ版深夜の真剣お絵かき60分一本勝負`. The base extractor's normalize_tags function was changed to use flat_map to allow this.

Fixes #5716.

### DIFF
--- a/app/logical/source/extractor.rb
+++ b/app/logical/source/extractor.rb
@@ -324,11 +324,7 @@ module Source
     end
 
     def normalized_tags
-      tags.map { |tag, _url| normalize_tag(tag) }.sort.uniq
-    end
-
-    def normalize_tag(tag)
-      WikiPage.normalize_other_name(tag).downcase
+      tags.flat_map { |tag, _url| TagNormalizer.normalize(tag) }.sort.uniq
     end
 
     def translated_tags
--- a/app/logical/source/extractor/newgrounds.rb
+++ b/app/logical/source/extractor/newgrounds.rb
@@ -62,9 +62,8 @@ module Source
         end
       end
 
-      def normalize_tag(tag)
-        tag = tag.tr("-", "_")
-        super(tag)
+      def normalized_tags
+        super.map { [it, it.tr("-", "_")] }.flatten.uniq
       end
 
       def display_name
--- a/app/logical/source/extractor/pixiv.rb
+++ b/app/logical/source/extractor/pixiv.rb
@@ -181,10 +181,6 @@ module Source
         tags
       end
 
-      def normalize_tag(tag)
-        tag.gsub(/\d+users入り\z/i, "")
-      end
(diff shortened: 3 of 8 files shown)


nonamethanks:
TBH I think it would be better for normalize_tag to always return an array of strings, or for the twitter extractor to override normalized_tags. It's a bad idea in general to have mixed type returns like this.

Given that this kind of tags is not limited to twitter but could also appear in mastodon/bluesky/other social media, option 1 is probably the better option.

But then again... Why are we only normalizing these for twitter? They're used on places like pixiv too: https://www.pixiv.net/en/tags/%E6%B7%B1%E5%A4%9C%E3%81%AE%E7%9C%9F%E5%89%A3%E3%81%8A%E7%B5%B5%E6%8F%8F%E3%81%8D60%E5%88%86%E4%B8%80%E6%9C%AC%E5%8B%9D%E8%B2%A0/artworks

ghost:
I refactored the normalize_tag method. How's this? I think some extractors replacing `-` with `_` would cause some tags (like x-men) to not be suggested.

nonamethanks:
As mentioned in discord, the issue here is that sites like tumblr or newgrounds lack precision, so while `one-piece` on newgrounds could mean both the series and the dress, on pixiv it only means the dress. So the dash normalization should only happen on sites like that.

ghost:
I was actually looking at Tumblr for unrelated reasons the other day (#6229) and Tumblr seems to use spaces and not hyphens for tags. https://github.com/danbooru/danbooru/blob/c1c54b02e33baacfaa8a98553a73ea6913f92f3d/test/unit/source/extractor/tumblr_extractor_test.rb#L70-L74

Should I move it just to the Newgrounds extractor, then?

ghost:
I moved the hyphen-to-space normalization to Newgrounds extractor and rebased the branch.

PR #6211, opened 2026-01-31
Fix #6181: Don't reload forum/comment page unnecessarily...

### LINKED ISSUE #6181: Don't reload forum/comment page unnecessarily when navigating to another forum post/comment
Now that quotes include a link to the forum post or comment they're quoting, it'd be nice if clicking on that would jump to the post/comment in question and highlight it if it's already visible, instead of reloading the page.

nottalulah:
Jump to the forum post/comment on the current page if it's visible. I wanted to use history.replaceState to avoid extra history entries, though that doesn't properly update the selection (not sure if it would be more desirable, anyway).

Comments could be done with HTML since we always know what page they'll be on, but forum posts need JS since we can't know ahead of time what page it'll be rendered on.

Fixes #6181.


evazion:
I think this won't work on the comment index when comments are hidden behind the `Show N more comments` link and loaded asynchronously.

ghost:
I'm not sure what to do about that. Fetch the comment to get the post ID, check if the post ID is on the current page and has hidden comments, try to load the rest of the comments?

PR #6216, opened 2026-02-02
Fix #6153: Calculate maxed? based on slots instead of points

### LINKED ISSUE #6153: Upload limit tooltip sometimes fails to show "Maximum amount of upload slots reached"
When the user has enough upload points to reach the maximum number of slots but less points than `MAXIMUM_POINTS`, the tooltip still shows "n approved posts needed for next level" instead of "Maximum amount of upload slots reached".
Image

Seems the logic was implemented in issue #4564 , but this edge case wasn't taken into account.

I think the solution would look something like this:
```rb
def maxed?
  upload_level >= UploadLimit.points_to_level(MAXIMUM_POINTS)
end
```

nottalulah:
Fixes #6153.

### DIFF
--- a/test/components/user_profile_component_test.rb
+++ b/test/components/user_profile_component_test.rb
@@ -1,0 +1,32 @@
+require "test_helper"
+
+class UserProfileComponentTest < ViewComponent::TestCase
+  def render_user(user)
+    as(@user) { render_inline(UserProfileComponent.new(user: user, current_user: user)) }
+  end
+
+  context "The UserUploadComponent" do
+    setup do
+      @user = create(:user)
+      @post = create(:post, uploader: @user, created_at: 10.days.ago)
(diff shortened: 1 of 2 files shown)


hdk5:
I don't like it: I remember as non-unres you could push it further than 40 slots to 40.99 or something, so that some deletions would still keep you at max slots as long as you have enough approved posts in 3 day span.

ghost:
It's confusing to the average user to see "posts needed for next level" only to have the next level actually be unattainable.

bipface:
any unittests?

hdk5:
> It's confusing to the average user to see "posts needed for next level" only to have the next level actually be unattainable.

average user does rarely reaches max slots

ghost:
> any unittests?

I didn't add them initially because I was looking for a component test, which didn't exist. I added some now.

> > It's confusing to the average user to see "posts needed for next level" only to have the next level actually be unattainable.
> 
> average user does rarely reaches max slots

Then you have no reason to complain. Though there have been numerous users who didn't get promoted for a long time despite having max slots.

PR #6224, opened 2026-02-03
Add icon to dtext search links.

nottalulah:
People requested differentiation between dtext searches `{{tag}}` and wiki links `[[tag]]`. This commit adds a magnifying glass icon similar to the current external link to search links.
image

I worry it would confuse people into thinking it's a Google link or something but I guess it's worth a shot.


evazion:
Changed the icon to use the same `magnifying-glass` icon we already have (which I thought came from Font Awesome but it's not the same as the one they currently have) and to tweak the spacing.

image

PR #6227, opened 2026-02-03
Fix #5950: Rework translated commentary buttons

### LINKED ISSUE #5950: Rework translated commentary buttons
There is no real need for 4 checkboxes. `commentary` is mutually exclusive with `commentary_request`, and `check_commentary`/`partial_commentary` imply `commentary_request`.

nottalulah:
[From Discord](https://discord.com/channels/310432830138089472/310846683376517121/1468282761440923711), it was suggested they be radio buttons. That requires us including an option for "None". The new "Untranslatable Commentary" was also requested.

Fixes #5950.

### DIFF
--- a/app/models/artist_commentary.rb
+++ b/app/models/artist_commentary.rb
@@ -1,14 +1,15 @@
 # frozen_string_literal: true
 
 class ArtistCommentary < ApplicationRecord
-  class RevertError < StandardError; end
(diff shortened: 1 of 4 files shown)


evazion:
The verdict is still out on whether `untranslatable_commentary` will be a tag but I'll go ahead with this anyway.

PR #6229, opened 2026-02-04
Fix #3503: Additional site links for translated tags

### LINKED ISSUE #3503: Additional site links for translated tags
Discussed this on Discord the other day.  Basically, it would be nice if we could have links to sites other than Pixiv with the translated tags on wikis.

Putting a link for every site after every tag would start to get very cluttered though, so one option discussed was to make the site links collapsible.  One idea for this would be to have a pair of parentheses with something in it to indicate that it expands, such as an ellipsis.

``ガールズ&パンツァー (...)``

After that, there were a couple of thoughts on how to indicate the different site links.

### List the full site name for every link
``ガールズ&パンツァー`` (**Seiga**,**Tinami**,**Nijie**)
However that could get very long as well if there are a lot of sites.

### List a letter for every link, with hover text for the full site name
``ガールズ&パンツァー`` (**S**,**T**,**N**)
The above would have more name collisions though.

### List a site icon for every link, with hover text for the full site name
``ガールズ&パンツァー`` (image)
Though there would be more network requests to grab every icon.

Since Pixiv is probably the most visited, it could remain as the main link, and all of the other sites could be in the expandable portion.

nottalulah:
Changes the other names links to show a dropdown with other sites.
image

These six were picked based on Discord users' feedback.

Fixes #3503.

### DIFF
--- a/config/danbooru_default_config.rb
+++ b/config/danbooru_default_config.rb
@@ -915,6 +915,19 @@ module Danbooru
     def reactions
       {}
     end
+
+    # A hash of site names to a function taking a tag name and returning a URL for
+    # tag searches on that site. Used to render other names links on wiki pages.
+    def tag_lookup_sites
+      {
+        pixiv: ->(name) { "https://www.pixiv.net/tags/#{name}/artworks" },
+        Twitter: ->(name) { "https://x.com/hashtag/#{name}" },
+        Bluesky: ->(name) { "https://bsky.app/hashtag/#{name}" },
+        Weibo: ->(name) { "https://s.weibo.com/weibo?q=%23#{name}%23" },
+        Lofter: ->(name) { "https://www.lofter.com/tag/#{name}" },
+        Tumblr: ->(name) { "https://www.tumblr.com/tagged/#{name.tr('_', ' ')}" },
+      }
+    end
   end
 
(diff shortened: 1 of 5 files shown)


config/danbooru_default_config.rb
```diff
@@ -915,6 +915,19 @@ def dtext_emojis
...
+    # A hash of site names to a function taking a tag name and returning a URL for
+    # tag searches on that site. Used to render other names links on wiki pages.
+    def tag_lookup_sites
+      {
+        pixiv: ->(name) { "https://www.pixiv.net/tags/#{name}/artworks" },
+        Twitter: ->(name) { "https://x.com/hashtag/#{name}" },
+        Bluesky: ->(name) { "https://bsky.app/hashtag/#{name}" },
+        Weibo: ->(name) { "https://s.weibo.com/weibo?q=%23#{name}%23" },
+        Lofter: ->(name) { "https://www.lofter.com/tag/#{name}" },
+        Tumblr: ->(name) { "https://www.tumblr.com/tagged/#{name.tr('_', ' ')}" },
```
evazion: The names should be escaped with `Danbooru::URL.escape(name)`, otherwise `?` characters will break the url.

Issue #4113, opened 2019-07-30
Store full source metadata


evazion:
We know how to extract lots of metadata from sources using source strategies, but most of this data gets thrown away. Instead it should be permanently stored in a table somewhere.

Specifically, for every source we should extract and store the following:

* site name (Pixiv, Twitter, etc)
* work id (Pixiv illust id, Twitter status id, etc)
* artist id
* artist username
* tags (original Pixiv tags, Twitter hashtags, etc)
* upload date (date the post was uploaded to Twitter / Pixiv / etc)
* image url
* page url
* page number (for images part of a Pixiv / Twitter batch)
* status (whether the source is active or deleted)

This would facilitate a number of things:

* Searching posts by Twitter/Pixiv/etc tags.
* Searching posts by Twitter/DeviantArt/etc IDs (#3924).
* Sorting posts by the date they were originally posted on Twitter/Pixiv/etc (#3899).
* Finding posts that are part of the same Pixiv or Twitter gallery.
* Finding untagged posts by the same artist.
* Simplifying artist lookups (find artists by artist id instead of trying to match profile urls).

r888888888:
Maybe this is a good candidate for Postgres's JSON indexing support.

GlassedSilver:
I love this idea. Especially things like upload dates can really come in handy.

nonamethanks:
https://discord.com/channels/310432830138089472/310846683376517121/1372311624643510293 recent discussion on the possible details of this implementation, linking it here for future reference


Pseudo tables with a list of stuff that would allow for all required features:
![Image](https://github.com/user-attachments/assets/94f54044-9586-42b7-a573-ffb65f0c5927)


Still not 100% sure on how to store tags. 

`source_tags (site_id, tag_name, post_count) -> post_source_tags (post_id, source_id, source_tags_array)` was the mentioned solution.

PR #6230, opened 2026-02-04
Implement tracking of source publication date for posts

ToksT:
Add a "published at" field for posts. This is the time that the artist first published the image at the source. Some examples:
* For web sources, this is the date that the post was created on the original website (which can usually be determined automatically by a source extractor).
* For game CGs, this is the date that the game was released.
* For magazine scans, this is the date that the issue was published (not when it was scanned).

Keeping track of this has several benefits:
* Knowing the surrounding context of when the artist posted an image. e.g. For art that references current world events or a new meme, not knowing the publication date makes it more difficult to determine what the image is referencing.
* Sorting posts by when the artist made them instead of when they got uploaded to Danbooru. This can fix images appearing out of order if they weren't uploaded to Danbooru in the right order. (Putting them in a series pool works to fix the order too, but publication date is a more general solution than pools.)
* Searching for art published within a certain time period, e.g. if you want to only look at art published in the 2010s.
* Archival purposes. The original source of an image can often be deleted (bad_id has almost 1.4 million posts), in which case the original metadata is often lost along with it. Danbooru already archives the artist's original commentary that accompanied the image, so archiving the original date it was posted as well makes sense.

Relevant forum discussion: https://danbooru.donmai.us/forum_topics/34323

---

It's important to note that the publication date is *not* the same as the time that the image was actually drawn, but is instead some amount of time later (because artists can wait a while before posting their work). Although Danbooru could theoretically keep track of the creation date that some artists date their work with, I think that idea causes too many issues in practice so this PR only deals with publication date. For more details about my thoughts on creation date click below.

<details>
<summary>Issues with creation date tracking</summary>

Let's say that image is posted to Twitter on 2026-03-15, but the artist commentary says "I drew this a couple years ago", and the image itself is dated as "03/09/24". If we only track publication date then we simply automatically extract that 2026-03-15 date and don't even need to think about anything else. But if Danbooru were to track creation date too, then a large number of problems appear:

* It cannot be automated at all, even for web sources. So it would need to be filled in manually for all posts, which would take an incredible amount of time.
* Danbooru users can misread dates. Is "03/09/24" March 9th or September 3rd? It probably depends what country the artist is from since date formats differ by country. Having to figure out what country every artist is from would create even more work.
* Artists can misremember or lie about when they drew something and put the wrong date. Unlike publication date, there's no way to verify creation date.
* The precision of creation date varies wildly, from the exact day, to the month, to just the year, to "I drew this a while ago".
* There's generally no way to get the creation time more specifically than just the date (no hour/minute/second or time zone) so sorting images drawn on the same day wouldn't work.
* Artists don't always date literally every single image they draw, so it wouldn't be very useful for sorting an individual artist tag as many posts would be missing the date.
* It wouldn't be useful for archival purposes either, as Danbooru already saves both the artist commentary and the image itself, so the original source being deleted already won't result in either of those being lost.

Rather than trying to deal with all of that, I think it's simpler to track only publication date, and creation date doesn't provide much search benefit over publication date anyway.

</details>

---

This PR has several parts:
* Add a published_at column for posts (and post versions) and display this date on the post show page if present. By default this publication date is null, which means it is currently unknown when the image was first published and this information needs to be filled in.
* Implement automatic date parsing for most source extractors so that the publication date for images from web sources can be filled in automatically without manual user intervention.
* Add a new `published:` edit metatag (as well as a separate text field in the post edit partial) to allow users to manually input publication dates (e.g. for non-web sources). Example: `published:2026-01-01` or `published:2026-01-01T12:34:56Z`
* Add new `published:` and `publishedage:` search metatags for searching for posts published within a certain range of dates/ages. (e.g. `published:2010-01-01..2020-01-01` to see images published in the 2010s.)
* Add a new `order:published` order metatag for sorting posts by their publication date. Posts with a null (unknown) publication date will be automatically excluded when using `order:published` (as if you had put `published:any` in the search).
* Add a fix script for existing Pixiv/Twitter posts to have DanbooruBot populate their publication dates. Because these two sites include publication date information within the URL itself, this does not require the script to make any external HTTP requests to sources.

Additional things that cannot be done on migration and instead will need to be done by users, after the update is live on Danbooru:
* Manually fill in the publication date for non-web sources. (Scans, game CGs, etc.)
    * To speed up setting many posts that were published together to the same publication date, it is possible to use the `published:` metatag in a tag script.
* For existing posts from websites other than Pixiv and Twitter, the publication date can be determined automatically by the source extractor, but this requires an HTTP request for every post, so users will need to fetch source data and click the "Fill" button to populate the publication date field for these posts.
    * It should be possible to speed up populating the backlog by writing an automated script that makes requests locally instead of through Danbooru, and then fills in the publication date via a bot account.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ For help, ask in the [#technical](https://discord.com/channels/31043283013808947
 
 ## Services
 
-Danboou depends on a couple of cloud services and several microservices to
+Danbooru depends on a couple of cloud services and several microservices to
 implement certain features.
 
 ### Amazon Web Services
--- a/app/components/source_data_component/source_data_component.html.erb
+++ b/app/components/source_data_component/source_data_component.html.erb
@@ -51,6 +51,18 @@
             <% end %>
           </td>
         </tr>
+
+        <tr class="source-data-publication-date">
+          <th>Published</th>
+          <td>
+            <% if @source.published_at.blank? %>
+              <em>None</em>
+            <% else %>
+              <%= helpers.compact_time(@source.published_at) %>
+              (<%= link_to "Fill", "javascript:void(0)", "x-on:click": "$('#post_published_at_string').val(#{@source.published_at_string.to_json});" %>)
+            <% end %>
+          </td>
+        </tr>
       </tbody>
     </table>
   <% end %>
--- a/app/helpers/post_versions_helper.rb
+++ b/app/helpers/post_versions_helper.rb
@@ -10,6 +10,7 @@ module PostVersionsHelper
     added_tags << "rating:#{post_version_value(post_version.rating)}" if post_version.rating_changed
     added_tags << "parent:#{post_version_value(post_version.parent_id)}" if post_version.parent_changed
     added_tags << "source:#{post_version_source(post_version.source)}" if post_version.source_changed
+    added_tags << "published:#{post_version_time(post_version.published_at)}" if post_version.published_at_changed
 
     removed_tags = post_version.removed_tags.compact
 
@@ -21,6 +22,7 @@ module PostVersionsHelper
       other_tags << "rating:#{post_version_value(other.rating)}"
       other_tags << "parent:#{post_version_value(other.parent_id)}"
       other_tags << "source:#{post_version_source(other.source)}"
+      other_tags << "published:#{post_version_time(other.published_at)}"
       obsolete_added_tags = added_tags - other_tags
       obsolete_removed_tags = removed_tags & other_tags
     end
@@ -40,11 +42,18 @@ module PostVersionsHelper
   end
 
   def post_version_field(post_version, field)
-    value = post_version_value(post_version.send(field))
-    prefix = ((field == :parent_id) ? "parent" : field.to_s)
+    if field == :published_at
+      value = post_version_time(post_version.send(field))
+      prefix = "published"
+      title = "Published"
+    else
+      value = post_version_value(post_version.send(field))
+      prefix = ((field == :parent_id) ? "parent" : field.to_s)
+      title = field.to_s.titleize
+    end
     search = "#{prefix}:#{value}"
     display = ((field == :rating) ? post_version.pretty_rating : value)
-    %{<b>#{field.to_s.titleize}:</b> #{link_to(display, posts_path(:tags => search))}}.html_safe
+    %{<b>#{title}:</b> #{link_to(display, posts_path(:tags => search))}}.html_safe
   end
 
   def post_version_value(value)
@@ -60,5 +69,9 @@ module PostVersionsHelper
       # This turns non-web sources with spaces (e.g., "File provided by the artist") into a single clickable entity.
       %{"#{source}"}
     end
+  end
+
+  def post_version_time(time)
+    time&.iso8601 || "none"
   end
 end
--- a/app/logical/post_edit.rb
+++ b/app/logical/post_edit.rb
@@ -14,7 +14,7 @@ class PostEdit
   CATEGORIZATION_METATAGS = TagCategory.mapping.keys
 
   # Pre-metatags affect the post itself, so they must be applied before the post is saved.
-  PRE_METATAGS = %w[parent -parent rating source status] + CATEGORIZATION_METATAGS
+  PRE_METATAGS = %w[parent -parent rating source published status] + CATEGORIZATION_METATAGS
 
   # Post-metatags rely on the post's ID, so they must be applied after the post is saved to ensure the ID has been created.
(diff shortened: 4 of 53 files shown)


ghost:
See also #4113 (and to a lesser degree, #6133). Published timestamp being an attribute of posts doesn't seem like a great idea given it's more an attribute of the source itself than the post or even the media asset.

ToksT:
> See also #4113 (and to a lesser degree, #6133). Published timestamp being an attribute of posts doesn't seem like a great idea given it's more an attribute of the source itself than the post or even the media asset.

#4113 plans to store full source information extracted from web sources, but I don't believe it plans to handle non-web sources as there's no automated way to extract and store that information from a plaintext source like "Ripped from game". I think that it's worth storing publication time for non-web sources as well.

This column doesn't necessarily have to be on the posts table, but trying to associate a publish time with a source URL that can then be shared by multiple posts introduces some new issues. What about multi-image posts where only some of the images get revised after the initial upload (or new images get added entirely)? Each image can theoretically have its own publish time independent from the source post itself, though the exact details depend on the specific website. So for some websites the model would need to store publish time on the "source post" and for others store it on the "source file" depending on how that website handles it. This would make searching and sorting posts on publish time more complicated compared to just storing it on the post table.

I don't see how #6133 is relevant.

ghost:
Non-web and web sources both have the same considerations to be made. A non-web source still has to come from somewhere. 

> What about multi-image posts where only some of the images get revised after the initial upload (or new images get added entirely)?

This is the is ends up being a never-ending rabbit-hole of more and more things you have to consider when implementing such things. Such cases mean we need to store attributes for each image individually in addition.

> I don't see how #6133 is relevant.

I mentioned that issue because it's ultimately trying to do the same thing as this PR: implement a change people want in a way that is not sustainable and ultimately does not address the root cause. Trying to do them the "correct" way just leads us back to #4113. Storing random extra data on the posts table is the "easy" way, but it's a bad idea and just leads to a lot of churn later. If we do this now, then what happens in the future if we get a more sensible system? Now the post/versions table has a "published" column that shouldn't exist, but can't really be gotten rid of, and we've made millions of pointless post versions.

ToksT:
> Non-web and web sources both have the same considerations to be made. A non-web source still has to come from somewhere.

So the new sources table would also include non-web sources like games and have metadata about those, sort of like Wikidata. Then a particular game release would be a single source, and every game CG ripped from that game would share that single source.

> I mentioned that issue because it's ultimately trying to do the same thing as this PR: implement a change people want in a way that is not sustainable and ultimately does not address the root cause. Trying to do them the "correct" way just leads us back to #4113. Storing random extra data on the posts table is the "easy" way, but it's a bad idea and just leads to a lot of churn later. If we do this now, then what happens in the future if we get a more sensible system? Now the post/versions table has a "published" column that shouldn't exist, but can't really be gotten rid of, and we've made millions of pointless post versions.

That makes sense. I'm going to close this and split the source extractor changes into a separate PR, as the extractors will need to be updated to parse this information regardless of what table it eventually gets stored in.

evazion:
I'd rather pull the extractor changes by themselves and get that part right before dealing with how to store the data.

Note that the date inside Pixiv URLs is the image revision date, not the post publication date. A Pixiv post can be published on one date and the individual images can be revised on later dates. [pixiv #127078599](https://www.pixiv.net/en/artworks/127078599) for example was first posted 2025-02-09 but later revised 2026-01-07.

This means that extractors need to distinguish between the `published_at` date of the post itself and the `published_at `date of each individual image in the post. Unfortunately extractors would need a big refactor to handle this properly. Right now we have `image_urls` and it returns a list of strings, when what we need is a list of extractors. That way we could do something like `Source::Extractor.find("https://www.pixiv.net/en/artworks/127078599").images.map(&:published_at)` to get the date of each image in the post. This would also solve the problem of some sites having per-image tags and per-image alt text, which we currently can't handle.

In other words, what we need is:

* For extractors to have `published_at` and `updated_at` methods (we want `updated_at` so we can detect revisions).
* For extractors to have some method that's like `image_urls` but that returns a list of extractors instead (something like `image_urls.map { Source::Extractor.find(image_url, parent: self) }`, where `parent` is so that these child extractors can get to their parent extractor).
* For extractors to be very careful to return the right `published_at` date for the given URL. If you do `Source::Extractor.find(page_url).published_at` on a Pixiv post or a tweet, that should return the publication date of the post or tweet itself. If you do `Source::Extractor.find(image_url).published_at` on a Pixiv image URL, that should return the date when the image itself was published (the revision date), not the date when the Pixiv post was published. This way we can keep both dates straight.

PR #6235, opened 2026-02-05
Fix #6183: Strip SiteCredentials

### LINKED ISSUE #6183: Strip SiteCredentials
Leading whitespace in Twitter CSRF token causes 403 with the error `{"errors" => [{"code" => 353, "message" => "This request requires a matching csrf cookie and header."}]}`.

It can be difficult to debug why because the leading whitespace is not shown on the UI and it's hard to dig into the API response because of all the parsed_get abstraction.

nottalulah:
Fixes #6183.


evazion:
A couple of sites use passwords and it's unlikely but possible that someone could have a space at the end of their password.

PR #6258, opened 2026-02-10
Source extractors: Implement post/image time parsing

ToksT:
Several source extractors now have `published_at` and `updated_at` methods. The behavior of these methods depends on whether the parsed URL was a post URL or direct image URL:
* Given a post URL, `published_at` will be the time the post was first published.
    * Implemented for: Pixiv, Twitter, DeviantArt, Tumblr, Mihuashi, Mastodon, Bluesky, Newgrounds, Weibo.
* Given a post URL, `updated_at` will be the time the post was last revised.
    * Implemented for: Pixiv, Mihuashi activity works.
* Given a direct image URL, `published_at` will be the time that specific image was published/uploaded, independent of the post it belongs to.
    * Implemented for: Pixiv, Twitter, Mihuashi. Most other sites don't include timestamps within the file URL.
    * For Pixiv, an image's publish time can be arbitrarily later than the post itself (if the post was revised).
    * For Twitter, an image's publish time can be arbitrarily earlier than the post itself (if the tweet was scheduled or drafted in advance).
* Given a direct image URL, `updated_at` will always be `nil`.
    * If we wanted to, we could do an HTTP HEAD request for the file to check the `last-modified` header time. But this time may not accurately represent when the artist last revised the image. For some sites like Mastodon/Mihuashi/Newgrounds, very old posts have last-modified times that are too new to realistically be when the artist edited the image, so I suspect all of the file mtimes across the site may have been touched when some kind of server migration happened. I'm not sure if these dates are worth keeping track of, so I didn't implement anything relating to this.

These times are shown to the user when fetching source data, but otherwise not used for anything or stored anywhere.

---

Furthermore, the base extractor class now has an `image_sources` method which returns extractors for each of the direct image URLs in the post. This is currently only used by tests and the `/source` HTML page to get the publish time of each individual image within the post.

Example of a Pixiv source that was revised (image date is later than the post date):
image

Example of a Twitter source that was scheduled a month in advance (video date is earlier than the post date):
image

---

Related:
#4113
https://github.com/danbooru/danbooru/pull/6230#issuecomment-3850655154

### DIFF
--- a/app/logical/source/extractor.rb
+++ b/app/logical/source/extractor.rb
@@ -411,6 +411,11 @@ module Source
       http_downloader.head(url).status.success?
     end
 
+    # @return [Array<Source::Extractor>] Return the list of image URL extractors.
+    def image_sources
+      image_urls.map { |image_url| Source::Extractor.find(image_url, parent_extractor: self) }
+    end
+
     # @return [Enumerator] An enumerator that lets you iterate across the chain of parent extractors.
     def each_parent
       return enum_for(:each_parent) unless block_given?
--- a/app/views/sources/show.html.erb
+++ b/app/views/sources/show.html.erb
@@ -75,14 +75,27 @@
 
             <% if @source.image_urls.present? %>
               <div class="grid grid-cols-<%= @source.image_urls.size.clamp(1..4) %> gap-2">
-                <% @source.image_urls.each do |url| %>
-                  <% if Source::URL.parse(url)&.file_ext.in?(%w[mp4 webm]) %>
-                    <%= tag.video src: url, referrerpolicy: "no-referrer", class: "max-w-full max-h-screen rounded object-cover", loop: true, controls: true %>
-                  <% else %>
-                    <%= external_link_to url, class: "w-fit h-fit" do %>
-                      <%= tag.img src: url, referrerpolicy: "no-referrer", class: "max-w-full max-h-screen rounded object-cover #{"w-full aspect-square" if @source.image_urls.size > 1}" %>
+                <% @source.image_sources.each do |image_source| %>
+                  <div style="width: fit-content;">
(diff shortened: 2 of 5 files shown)


evazion:
This looks good, I'll go ahead and pull this so we can iterate on it.

>If we wanted to, we could do an HTTP HEAD request for the file to check the last-modified header time. But this time may not accurately represent when the artist last revised the image.

We could potentially do it anyway, and treat `updated_at` for files as "when the file was last modified (for any reason)" rather than "when the file was last revised by the artist". This could be useful for detecting when files have changed. For example, Pixiv losslessly recompressed all their files in 2020, which caused all our existing files to become `md5_mismatch`es and new posts to become `pixel-perfect_duplicate`s. Knowing the modification date could help us tell when a file has potentially changed at the source for any reason.

In fact, we could also have an `md5` method to calculate the md5 of an image URL to help detect `md5_mismatch`es (this would of course be slow if called on the `/source` page, but it could let us make a background job or something to automate the tagging of `md5_mismatch`).

ToksT:
> For example, Pixiv losslessly recompressed all their files in 2020, which caused all our existing files to become `md5_mismatch`es and new posts to become `pixel-perfect_duplicate`s. Knowing the modification date could help us tell when a file has potentially changed at the source for any reason.

For Pixiv specifically I'm not sure how trustworthy the `last-modified` header is. e.g. [post #3389312](https://safebooru.donmai.us/posts/3389312) appears to be one of the files you're referring to as it was uploaded in January 2019 and then had the `md5_mismatch` tag added to it in 2020. But Pixiv claims the `last-modified` time for that image is also January 2019, which must be a lie if they recompressed it in 2020.

Might be useful for other sites that use `last-modified` more accurately though.

PR #6266, opened 2026-02-12
Ugoira MP4 samples

### LINKED ISSUE #5090: Change ugoira samples to mp4
Turns out that .webm VP9 is slow to encode and has compatibility problems with some browsers.

TypeA2:
Fixes #5090. Benefits include better (or any, for that matter) iOS compatibility.

It uses a single pass encoding using `libx264` with `-preset medium -crf 14`. I haven't tested these numbers in particular they just felt kinda right. Timecodes are set using ffmpeg's `concat` demuxer, so no need for external tools and only one external call is required.

Some notes:
* After extensive testing, the old problem of browsers discarding the final frame still exists. The last frame is duplicated with 0 length to avoid this. This seems to work on FF and Chrome.
* H.264 only supports dimensions divisible by 2, so images where this isn't the case are padded by a single row or column of white pixels (to match the default Pixiv background color of white).

# **Important**
I added some code to fall back the ugoira player to use WebM samples if the MP4 one does not exist, so that all existing images will continue to work for the user. **_HOWEVER_** the `sample` video URLs will return `.mp4` extensions even when only a `.webm` exists. I don't know if there's a good way to deal with this, it might just be better to remove it from the API until all samples have been regenerated.

### DIFF
--- a/app/logical/media_file/ugoira.rb
+++ b/app/logical/media_file/ugoira.rb
@@ -345,7 +345,7 @@ class MediaFile::Ugoira < MediaFile
 
         # H.264 only supports even dimensions, so pad by 1 pixel in width or height to fit
         ffmpeg_out, status = Open3.capture2e(
-          "ffmpeg -f concat -i #{concat_path} -c:v libx264 -preset medium -profile:v high -crf 14 -pix_fmt yuv420p -vf 'pad=ceil(iw/2)*2:ceil(ih/2)*2' -fps_mode:v passthrough -movflags +faststart -f mp4 -y #{output_file.path}"
+          "ffmpeg -f concat -i #{concat_path} -c:v libx264 -preset medium -profile:v high -crf 14 -pix_fmt yuv420p -vf 'pad=ceil(iw/2)*2:ceil(ih/2)*2:0:0:white' -fps_mode:v passthrough -movflags +faststart -f mp4 -y #{output_file.path}"
         )
         raise Error, "ffmpeg failed: #{ffmpeg_out}" unless status.success?
 


evazion:
Closing this in favor of b3874f6c737ca7e9cedc2c61c87dcefa520aad9b. This added the ability to convert ugoiras to any format (.webm, .mkv, .mp4) or codec (h264, h265, vp8, vp9, av1), and added a `bin/media-conv` tool for converting ugoiras from the command line.

The default is still to generate VP9 webms instead of MP4. I'm still not sure what encoder settings we should use for h264, and we still probably need to generate constant frame videos instead of variable frame rate to avoid issues with browsers that don't like VFR videos. In particular, I've noticed that even when Chrome plays a VFR video correctly, it's hard to seek the video because Chrome can only seek to frame boundaries, and for slideshow type ugoiras frames can be multiple seconds apart.

There were a few issues here:

* Padding odd dimensions with white causes a noticeable line on ugoiras with dark backgrounds. Duplicating the edge pixels with the `fillborders` filter avoids this.
* The `concat` demuxer rounds durations to the nearest 40ms. The simplest workaround is to multiply the durations by 40 then use `-itsscale 0.025` to divide by 40 and get the right number. But this doesn't really matter because we have to use `mkvmerge` to fix the frames anyway.
* FFmpeg ignores the duration of the last frame and sets it to 40ms. I tried numerous things but the only way I found to avoid it was to use `mkvmerge` to fix the timestamps afterwards. `mkvmerge` also works on .mp4 files, but it converts them to .mkv, so you have to remux them back to .mp4.

PR #6268, opened 2026-02-13
github: only run tests for the master branch

nonamethanks:
Also kill all in-progress tests when a new commit is pushed.

Right now [they start and get queued at every push no matter where](https://github.com/danbooru/danbooru/actions/workflows/test.yaml).


evazion:
Might have to redo this later if we want to get tests working on PRs.

PR #6281, opened 2026-02-18
Fix #3503: Additional site links for translated tags

### LINKED ISSUE #3503: Additional site links for translated tags

nonamethanks:
Reopening pull #6229, but the additional moving of logic for tag url generation to Source::URL. We have a lot of duplicate logic all over the place, I'll do a cleanup of our extractors to simplify the logic.


nonamethanks:
Gonna reopen another pull with just the original commit and some adjustments, otherwise this will take forever, as a refactor ended up being a lot more involved than I thought.

PR #6298, opened 2026-03-01
Uploads: Search source, referer, and source_url fields together.

### LINKED ISSUE #5960: Support searching for file uploads on "My Uploads" page
If it already exists, it's not obvious to me. File uploads have `null` in their source field, so trying `file://*` in source search doesn't work, and manually writing `search[source_present]=false` also doesn't work because that does `where(source: "")`, which returns nothing. It seems the only way at present is to add `search[upload_media_assets][source_url_like]=file://*` which is very non-obvious.

### LINKED ISSUE #6257: "Source" search in uploads should search by both upload's source, referer_url, as well as upload_media_asset's source_url
For example, now if I search for `*twitter.com*` on the page, I only find images I uploaded using tweet link, but not those with`?url=https://pbs.twimg.com/...&ref=https://twitter.com/...`. 
Surely, I can do `search[referer_url_ilike]=*twitter.com*` myself, but I want the results merged together in one search.
Similarly, I want to do the opposite and search for `*pbs.twimg.com*` and get the results where the url was extracted from only the provided tweet link.
The option to search for each of those separately should also remain available.

nottalulah:
Fixes #6257 and fixes #5960.

Since `referer_url` can be anything, sometimes you can have false positives.

### DIFF
--- a/app/models/upload.rb
+++ b/app/models/upload.rb
@@ -122,8 +122,16 @@ class Upload < ApplicationRecord
     where(upload_media_assets.where("upload_media_assets.upload_id = uploads.id").arel.exists)
   end
 
+  def self.any_source_matches(source)
+    upload_media_assets = UploadMediaAsset.where_ilike(:source_url, source)
+    q = where_ilike(:source, source)
+    q = q.or(where_ilike(:referer_url, source))
+    q = q.or(where(upload_media_assets.where("upload_media_assets.upload_id = uploads.id").arel.exists))
+    q
+  end
+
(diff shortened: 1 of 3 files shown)


evazion:
This could be slow in production if you have a large number of uploads. Currently these fields are unindexed; adding trigram indexes on `uploads.source`, `uploads.referer_url`, `upload_media_assets.source_url`, `upload_media_assets.page_url` may help (but they might need to be `(user_id, source)` indexes to actually be chosen by Postgres).

PR #6300, opened 2026-03-02
Prevent changing artist tags' category.

### LINKED ISSUE #5326: Don't allow category change for tags with active artist entries
I regularly (multiple times a week) find artist tags changed to chartags because our users don't bother to check if the tag exists when pretending a category change, which ends up causing a mess for artists.

Trying to change the category of tags with an active artist entry should stop you and not add the tag if done via tagging box, because it's guaranteed you're trying to tag something else.

nottalulah:
Fixes #5326. It doesn't give you an error if the category change fails (also currently does not happen if the tag is too large to change the category), but I wasn't sure how to add that.

### DIFF
--- a/app/logical/bulk_update_request_processor.rb
+++ b/app/logical/bulk_update_request_processor.rb
@@ -47,7 +47,7 @@ class BulkUpdateRequestProcessor
         [:rename, Tag.normalize_name($1), Tag.normalize_name($2)]
       when /\A(?:mass update|update) (.+?) -> (.*)\z/i
         [:mass_update, $1, $2]
-      when /\Acategory (\S+) -> (#{Tag.categories.regexp})\z/i
+      when /\Acategory (\S+) -> (.*)\z/i
         [:change_category, Tag.normalize_name($1), $2.downcase]
       when /\Aconvert (.+?) -> (.*)\z/i
         [:convert, $1, $2]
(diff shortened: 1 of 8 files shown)


app/logical/bulk_update_request_processor.rb
```diff
@@ -160,7 +160,13 @@ def validate_remove_implication(antecedent, consequent)
   def validate_change_category(tag_name, category)
     tag = Tag.find_by_name(tag_name)
     if tag.nil?
-      errors.add(:base, "Can't change category of [[#{tag_name}]] to #{category} ([[#{tag_name}]] doesn't exist)")
+      errors.add(:base, "Can't change the category of [[#{tag_name}]] to #{category} ([[#{tag_name}]] doesn't exist)")
+    elsif TagCategory.mapping[category].nil?
+      errors.add(:base, "Can't change the category of [[#{tag_name}]] to #{category} (#{category} is not a valid category)")
+    elsif Tag.categories.value_for(category) == tag.category
+      errors.add(:base, "Can't change the category of [[#{tag_name}]] to #{category} ([[#{tag_name}]] is already in that category)")
```
evazion: The `<tag>is already in that category` error should only apply when the BUR is created, not when the BUR is approved (i.e. not when `validation_context == :approval`). Approving a BUR should be idempotent so that you can approve failed BURs again without errors about things that succeeded. If a BUR is trying to change a tag to an artist tag and it fails halfway through, you should be able to approve it again because it's okay if the tag is already an artist tag, nothing will change, just like it's okay if it's trying to create an alias or implication that already exists.

nottalulah: [Like this](https://github.com/danbooru/danbooru/compare/danbooru:a65f3b2...danbooru:5021c40#diff-f95182d20e35cffdd7c5d8dbf8607ea016ec6d88b2a5f04f9258929cc0791c98R167-R172)? I'm not exactly sure which should be excluded from approval context since it seems inconsistent—deprecate ignores tags being nil in approval context, for example. Granted, I am not sure that path could ever be reached.

Issue #6231, opened 2026-02-05
help:bookmarklet_notice should generate the list automatically, not be edited manually


nonamethanks:
https://danbooru.donmai.us/wiki_pages/help:bookmarklet_notice

A while ago I noticed that this was out of date by like a year at least.

ghost:
I think the URL classes would need a sort of "base_url" method for this.

nonamethanks:
It'd have to be a function that outputs dtext because some strategies (like youtube) require to specify which posts are supported. It makes more sense for it to be in extractors because those are the ones that actually support post extraction - we have plenty of sites that we parse but we don't extract from.

evazion:
<img width="905" height="991" alt="Image" src="https://github.com/user-attachments/assets/2486e26b-3db9-4561-b649-8038f45c336d" />

This is what it looks like now. It still needs a way to add help text for sites to explain which things are supported.

This took a lot of refactoring because sites and extractors aren't 1-to-1. Some sites have multiple extractors (see: Reddit, Ko-fi), and some extractors handle multiple sites (see: Gelbooru, Moebooru, Misskey). So this necessitated introducing a new `Source::Site` concept and having each `Source::URL` class define which sites it supports (if it supports multiple) and which extractors it can use (if it can use multiple).

evazion:
![Image](https://github.com/user-attachments/assets/67065e21-61ae-414b-b8fd-df131e3afa42)

I think we could use an animation like this to show how to actually add the bookmarklet, because I think a lot of people still won't get it.

Issue #4958, opened 2022-01-15
pixiv strategy - handle unlisted pages


bipface:
Example:

* [page url](https://www.pixiv.net/en/artworks/unlisted/D2wOwQiM8yFkN5Nn1Tiq) → [testbooru](https://testbooru.donmai.us/uploads/new?url=https://www.pixiv.net/en/artworks/unlisted/D2wOwQiM8yFkN5Nn1Tiq)
  * `https://www.pixiv.net/en/artworks/unlisted/{alphanumeric-id}`
* [file url](https://i.pximg.net/img-original/img/2020/12/11/16/36/36/86229794-4263f0dcf6cabb2e6121e71a77885c1a_p0.png) → [testbooru](https://testbooru.donmai.us/uploads/new?url=https://i.pximg.net/img-original/img/2020/12/11/16/36/36/86229794-4263f0dcf6cabb2e6121e71a77885c1a_p0.png)
  * `https://i.pximg.net/img-original/img/{timestamp}/{digits}-{32-hex-digits}_p{page}.png`

I was able to upload using the direct file URL without issues, so it should be feasible to implement support for the page URLs.
There's a number of angles to consider though — such as whether there's a way to map from file url to page url (normalisation) like with ordinary pixiv works, and whether the `pixiv:` metatag needs to be extended to cater for the alphanumeric IDs

BrokenEagle:
Do you have an example that can be used for testing purposes? If needed, you can post that information in `#technical-private` on Discord.

bipface:
is there a problem with the example above ?

BrokenEagle:
Ah, sorry. I didn't notice the links you had included above.

nottalulah:
Posts with these types of URLs are being tagged bad_source automatically now. See [this post](https://danbooru.donmai.us/posts/5439159) (NSFW)

PR #6301, opened 2026-03-04
post: replace pixiv_id with source_site/source_id metatags and generic site ids

hdk5:
- Store posts' source site name and external ID as a string in separate columns
- - Also store numeric external ID for numeric range searches, and keep storing pixiv_id to keep the old behavior
- Add posts search metatags `source_site` and `source_id`
- - If `source_id` argument looks numeric, range searches are supported
- - otherwise, only full string match is supported
- - additional string match options (i.e.. wildcard) should be considered
- Only few actual extractor behavior is implemented (for testing and as a proof of concept)
- - Source URL parsers need to implement `site_name` and `site_id` methods;  the former is already done
- Add source site grouping option for posts reports
- - to consider: include unrecognized and non-web sources as separate rows

### DIFF
--- a/app/logical/uint128_type.rb
+++ b/app/logical/uint128_type.rb
@@ -2,6 +2,7 @@
 
 class Uint128Type < ActiveRecord::Type::Value
   def cast_value(value)
+    return if value.nil?
     case value
     in Integer
       value
@@ -11,6 +12,7 @@ class Uint128Type < ActiveRecord::Type::Value
   end
 
   def serialize(value)
+    return if value.nil?
     value.to_s(16).rjust(32, "0")
   end
 end


bipface:
will pixiv be the only source where IDs are treated as integers?
for example currently i can search `pixiv:<10000` to get posts with pixiv id 9999 and below, but if i wanted to search say `source_site:nijie source_id:<10000`, 9999 is not less than 10000 lexically.

additionally, if pixiv posts are cast to integers unconditionally, how will we add support for unlisted posts in the future (issue #4958)?


when i thought about this feature in the past my instinct was to have two source_id columns, like `source_id_int` and `source_id_str`, where the site determines which one is used -- or in the case of pixiv, both would have to be used.

hdk5:
> will pixiv be the only source where IDs are treated as integers? for example currently i can search `pixiv:<10000` to get posts with pixiv id 9999 and below, but if i wanted to search say `source_site:nijie source_id:<10000`, 9999 is not less than 10000 lexically.

I keep the old `pixiv_id` behavior and numeric searches in mind, i.e.
https://github.com/danbooru/danbooru/pull/6301/changes#diff-fd0046e2ea37f7ce3e91c417c57233c4b71a3324c477b5070196305ba363835cR1393, https://github.com/danbooru/danbooru/pull/6301/changes#diff-fd0046e2ea37f7ce3e91c417c57233c4b71a3324c477b5070196305ba363835cR1400

I might actually change internal implementation back to just storing three separate columns (i.e. `site_id`, `site_id_num`, `pixiv_id`), as converting those on the fly and ensuring it keeps working as intended (and also fast) turned out to be much bigger pain.
@evazion give me your thoughts on this.

> additionally, if pixiv posts are cast to integers unconditionally, how will we add support for unlisted posts in the future (issue #4958)?

I imagine `pixiv:` metatag would remain only not to break people's shit and work only with old numeric IDs.

evazion:
* The search syntax should be something like `site:pixiv/1234`, `site:twitter/1234`, `site:pixiv`, `site:any`, `site:none` instead of `source_site:pixiv source_id:42`.
* Ordering by site id (i.e. `site:pixiv order:site/pixiv`) could be useful (as long as the site has orderable ids).
* I'd rather store the name of the site as an integer enum instead of as a string. This is slightly more efficient and makes it easier to change the name of a site if another Twitter->X situation ever happens again. This would require a big list of sites somewhere with a site_id -> site_name mapping, but we need that anyway for #6231, and we already have a partial one in site_credentials.rb.
* I think it should be called `Source::URL#work_id` instead of `Source::URL#site_id`. `site_id` could mean multiple things: the id of the site itself (0 for Pixiv, 1 for Twitter, etc), the id of the work (i.e. the Pixiv post id, the Twitter status id), or the id of the image (i.e. the Twitter image filename, the Pixiv date+post_id+page_number combo). We should be clear on this. For our purposes here we want the work id.
* The work id isn't always an integer and can't always be converted to a 64-bit integer. Sometimes it's a 128-bit UUID, sometimes it's a long base36 string, sometimes it's a username+post_id combo (i.e. `evazion/posts/1234`). Going through and defining `Source::URL#work_id` for every site first would make it easier to tell what we're up against.
* For unlisted Pixiv posts there are two ids, i.e. https://www.pixiv.net/en/artworks/unlisted/ntQchboUi1CsqMhDpo5j has `ntQchboUi1CsqMhDpo5j` in the url, but the actual id is `90237981`. So the choices are: store `90237981` as the id (works, but you can't reach the page or find the post in the api using that id), store `ntQchboUi1CsqMhDpo5j` as the id (works, but then you can't compare it against other pixiv ids), or make `source_ids` an array and store both (complicated, makes it hard to filter or sort by id).
* Another idea would be to have a `source_metadata` JSON column that lets us store arbitrary JSON (in addition to dedicated source_name, source_id columns). This is less efficient than dedicated columns, but it might make handling corner cases like unlisted pixiv ids easier, since we could stuff extra data here without needing a migration to add new columns. If the column is NULL by default, then it's essentially free.
* I think having both `source_id_str` and `source_id_num` is ugly but probably necessary. It would technically be possible to do something like `CREATE INDEX index_posts_on_source_id_num ON posts ((CAST(source_id_str AS bigint))) WHERE source_id_str ~ '^\d+$';` to only store the column as a string and use an expression index to search it as an int, but it still wouldn't be able to handle things like base36 ids or UUIDs.
* `source_id_num` could potentially be a `uuid` type instead of a bigint. That would let us store up to 128-bit ids if we need it. Media asset pixel hashes are stored this way.
* The migration will have to be broken up into multiple commits to deploy this to production without downtime. It would involve first marking the new columns as `ignored_columns`, then creating them, then unignoring them and storing data into both the old and new columns simultaneously, then switching to using the new columns only and marking the old `pixiv_id` column as ignored, then dropping it, then finally removing it from the `ignored_columns` list. (see https://docs.gitlab.com/development/database/avoiding_downtime_in_migrations/#renaming-columns)

hdk5:
Changed the syntax to `site:pixiv/1234`, added the `order` option.
Retouched some extractors to use `work_id`. Some of them already take this name. Some use it for what we need. Reviewing all URL parsers sounds like a chore I am not interested in yet.
For that reason, `order` autocomplete suggests only `site` for now.
Storing source_id_num as uuid seems not that beneficial. We would have to convert it to back to `numeric` for comparison operations, no? Would be easier to just use numeric in that case if we actually need better precision.
As for `source_id_num` column, I tried using index for source_id bigint cast first (it is still in the commit history), but scraped it in favor of separate column which is easier to use in rails code, which otherwise would look even more ugly.
For the same reason pixiv_id also remains as a separate column, as its actual behavior (in production) is somewhat non obvious and behaves differently for null/not-null values and for equality/comparison/range queries. New asserts is existing PostQueryBuilder pixiv_id test reflect that. That is unless we want to throw away `pixiv_id:` query metatag completely, which probably is important to some user who uses it to heat his computer or something.

evazion:
> Retouched some extractors to use work_id. Some of them already take this name. Some use it for what we need. Reviewing all URL parsers sounds like a chore I am not interested in yet.

I mention this because I don't want to get backed into a corner because there's some site we forgot to consider. Here are some examples:

* https://skeb.jp/@kotora_hu/works/1 (work_id: `kotora_hu/1`, username+id combo)
  * https://skeb.jp/works/133404 (redirects to above)
* https://www.artstation.com/artwork/1NzJ4K (random base36 string, order by id is meaningless)
* https://odaibako.net/odais/e03a9ff1-4e3c-4168-8ef2-b888f51c25f2 (128-bit UUID)
  * https://odaibako.net/posts/01957504f709fe2bb0c9aa6338aabbee (different id type)
* https://caswac1.tistory.com/entry/Emera1Reboot-Rest-in-Peace (work_id: `caswac1/Emera1Reboot-Rest-in-Peace`)
* https://opensea.io/item/ethereum/0x495f947276749ce646f68ac8c248420045cb7b5e/25498143383868488060407396481663496375452486694447065582311815598428410347521 (two really long ids)
* https://medibang.com/picture/4b2112261505098280008769655/
* https://medibang.com/book/f72107281839259430002176282/
  * Two different id types for the same site (`picture_id` and `book_id`). For this we'd either need a column for the id type (`picture_id`, `book_id`), have the site column indicate both the site name and the id type (`medibang_picture`, `medibang_book`), or namespace the id to distinguish between types (`picture/4b2112261505098280008769655`, `book/f72107281839259430002176282`).
* https://seiga.nicovideo.jp/seiga/im3521156
* https://seiga.nicovideo.jp/watch/mg925907
* https://www.nicovideo.jp/watch/sm36465441 
  * Three different id types like above, but they're already namespaced by the site.

> Storing source_id_num as uuid seems not that beneficial. We would have to convert it to back to numeric for comparison operations, no? Would be easier to just use numeric in that case if we actually need better precision.

`uuid` can be ordered or used in range comparisons without casting, in the same way as a bigint. It just gets formatted as a UUID (but see app/logical/md5_hash_type.rb to see how it could be serialized/deserialized as a regular integer). I guess `numeric` would also be an option if we wanted arbitrary precision integers.

https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL

> For the same reason pixiv_id also remains as a separate column, as its actual behavior (in production) is somewhat non obvious and behaves differently for null/not-null values and for equality/comparison/range queries. New asserts is existing PostQueryBuilder pixiv_id test reflect that. That is unless we want to throw away pixiv_id: query metatag completely, which probably is important to some user who uses it to heat his computer or something.

For backwards compatibility all we need to worry about is `pixiv:id`. It's okay if we have to throw away other stuff like `-pixiv:id` and `pixiv:<id`, those were never really guaranteed to work to begin with.

hdk5:
Basically, each source url can be broken down into components like `[source_site, *namespace, source_id]`
Keeping namespace/work type in a separate column and have source id represent only the final identification component would probably be the best.
Parsers would have to implement `work_kind` and `work_id` in that case.
Namespaces in that breakdown would represent either user (skeb seems to be the only case, some other sites may have user name or id in URLs, but work IDs on them are still global and not scoped to user) or work type. 
Not sure what about the case when the namespace would be composite (e.g., each work has id scoped to both user and post type). I don't see any known cases like there yet, but who knows what site we might want to support in the future.
Regarding unlisted pixiv posts, since it works on `Source::URL` level, not at `Source::Extractor` level, it will rely on a single chosen page url representation. In this case, since we want to have a link that user can actually go by, it would be the unlisted alphanumerical id.
As for numeric id representation, I suggest the conversion to be moved into each individual parser subclass, so it could implement non-decimal conversions like base36 or uuid. I don't know, however, how it should work for actual searching, which only supports decimal ranges/comparison queries.

hdk5:
I don't really have a clear vision on what we want to have as a final result, so I'll close it in favor of someone more competent taking over.

evazion:
I think you're on the right track, we just need to have some method on `Source::URL` that knows how to return the right IDs for each site.

In https://github.com/danbooru/danbooru/commit/808b29e423c3db3afd7a6bb812b69dd73c08acf7 I added a system for generating IDs for each site, so you can say

```ruby
site = Source::URL.parse("https://x.com/BOW999/status/1261877313349640194").source_site
site.internal_name # => "twitter"
site.site_id # => 1408455283 (generated based on `internal_name`)
```

This should make it possible to at least store the `site_id` for each post so that `site:pixiv`, `site:twitter` work.

(Federated sites like Mastodon or Misskey are another complication though. There are many Misskey instances, but they're all linked together, so if we have a Misskey post, should the site be the particular instance it's from, or should it just be "Misskey"?)

evazion:
Thinking about it further, maybe what we need is this:

```ruby
Source::URL.parse("https://x.com/BOW999/status/1261877313349640194").components
# => { site: "twitter", status_id: 1261877313349640194, username: "BOW999" }

Source::URL.parse("https://x.com/i/web/status/1261877313349640194").components
# => { site: "twitter", status_id: 1261877313349640194 }

Source::URL.parse("https://www.pixiv.net/en/artworks/49270482").components
# => { site: "pixiv", illust_id: 49270482 }

Source::URL.parse("https://www.pixiv.net/en/artworks/unlisted/ntQchboUi1CsqMhDpo5j").components
# => { site: "pixiv", unlisted_id: "ntQchboUi1CsqMhDpo5j" }

Source::URL.parse("https://skeb.jp/@kotora_hu/works/1").components
# => { site: "skeb", username: "kotora_hu", work_id: 1 }

Source::URL.parse("https://medibang.com/picture/4b2112261505098280008769655/").components
# => { site: "artstreet", picture_id: "4b2112261505098280008769655" }

Source::URL.build(site: "pixiv", illust_id: 49270482)
# => "https://www.pixiv.net/en/artworks/49270482"

Source::URL.parse("pixiv/49270482")
# => { site: "pixiv", illust_id: 49270482 }

Source::URL.parse("twitter/1261877313349640194")
# => { site: "twitter", status_id: 1261877313349640194 }
```

That way each site just returns everything in the URL, and in the worse case we could just store it all in a JSON column. If we figure out a unified scheme for the `work_id` we could make a dedicated column for it later on, otherwise we could just search the JSON.

PR #6317, opened 2026-03-10
Modqueue: Sort by least disapprovals first by default.

### LINKED ISSUE #5862: Order the modqueue by amount of disapprovals rather than age
This would effectively be the same as the current situation, except that posts with fewer disapprovals that go past page 10 aren't forgotten forever.

The one annoying part about this would be that posts that are blacklisted by a lot of people will come up on top, but those posts are going to always be mixed in anyway so they would end up on top in any case.

### LINKED ISSUE #5863: Add order:disapproval_count
Useful to have, and also a blocker for issue #5862. Even if that isn't implemented, it'd just be neat to be able to sort the posts in the queue by this.

nottalulah:
Fixes #5862 and fixes #5863.

I switched from `paginated_search` to `user_tag_match` because `order_matches` has a bug where it doesn't apply joins (these are instead done in `PostQueryBuilder`), which makes `search` unusable for any of the association counts.

### DIFF
--- a/app/controllers/modqueue_controller.rb
+++ b/app/controllers/modqueue_controller.rb
@@ -8,10 +8,14 @@ class ModqueueController < ApplicationController
     authorize :modqueue
     @mode = params.fetch(:mode, "gallery")
     @limit = params.fetch(:limit, CurrentUser.user.per_page).to_i.clamp(0, PostSets::Post::MAX_PER_PAGE)
+    @order = params.dig(:search, :order) || "disapprovals_asc"
+    @tags = params.dig(:search, :tags)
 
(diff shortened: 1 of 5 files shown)


evazion:
I'm afraid this will turn people off the modqueue by effectively putting the least appealing / most boring posts first. It also means that things like comics, parent/child sets, and sequences of related posts by the same uploader won't necessarily be in order.

Also, I'm afraid of rule-breaking uploads getting less visibility. Every time a post gets disapproved, it will get pushed further down the modqueue, so if a few people just disapprove a post without dealing with it, then it's less likely that others will see it and deal with it.

evazion:
https://github.com/danbooru/danbooru/commit/ff6dafd6aa0a0aa3aa80247c4b50bb9cf97f9359

I had to revert back to defaulting to order by newest first because `is:modqueue order:disapprovals` times out in production.

PR #6318, opened 2026-03-10
Increase tag limit.

nottalulah:
Gold has not available for four years anyway. mashallah


Hinatsumi98:
@evazion onegai senpai

hdk5:
Let the plebs see the good stuff too while at it

PR #6334, opened 2026-03-18
add `pool:none` edit metatag

hdk5:


evazion:
This will remove both series pools and collection pools, which I think is not usually what you want.

hdk5:
How about `-pool:series` and `-pool:collection` then?

evazion:
That would be fine.

Issue #5124, opened 2022-04-15
Standardize how line breaks get saved for all text fields


BrokenEagle:
I see this switch a lot in diffs, between Linux style `\n` and Windows style `\r\n`. Not only is this distracting to the actual change, but it could be interpreted as an actual change, which could unnecessarily create a new version. Also, I haven't experienced it myself, but apparently Macs use `\r` for line breaks, which would also need to be handled the same way.

1. Use either style as the standard (except for the Mac style), i.e. `\n` or `\r\n`
2. Convert all existing records (minus versions) to the new line break (optional)
3. Convert all existing versions to the new line break (optional)

ghost:
Fixed by #6335.

PR #6335, opened 2026-03-18
chore #5124: Standardize how line breaks get saved for all text fields

steve-redka:
#5124
Have dtext_attribute automatically normalize whitespace (NFC, normalize_whitespace, strip) for all DText fields. Remove redundant manual normalizes from WikiPage, Ban, IpBan, NewsUpdate, and ArtistCommentary.

### DIFF
--- a/app/logical/concerns/d_text_attribute.rb
+++ b/app/logical/concerns/d_text_attribute.rb
@@ -57,6 +57,9 @@ module DTextAttribute
         memoize "dtext_#{name}_was"                  # memoize :dtext_body_was
 
         prepended do
+          # normalizes :body, with: ->(value) { value.to_s.unicode_normalize(:nfc).normalize_whitespace.strip }
+          normalizes name, with: ->(value) { value.to_s.unicode_normalize(:nfc).normalize_whitespace.strip }
+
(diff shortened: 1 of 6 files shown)


evazion:
Changing line endings from CRLF to LF should be a separate issue. Existing wikis use CRLF, so it would mean that every time a wiki is edited, every line ending gets changed, even when the user didn't touch that line. This would pollute diffs unless diffs are taught to ignore line ending changes. CRLF was chosen because most users are on Windows, so that's what most existing DText uses.

It would be better to have `dtext_attribute` normalize whitespace globally by having it automatically add `normalizes attribute, with: ->(value) { value.to_s.unicode_normalize(:nfc).normalize_whitespace.strip }`. See here for how it adds `validates` to every model: https://github.com/danbooru/danbooru/blob/40b5c0a99a97df822db39d690cca48e0b4922193/app/logical/concerns/d_text_attribute.rb#L59-L68

steve-redka:
Ok

Instead of changing the default EOL, I added `normalizes` directly to dtext_attribute so every dtext field gets whitespace normalization automatically. Removed the redundant manual declarations from the 5 models that had them.

Issue #6324, opened 2026-03-15
Two posts cannot be edited at all due to recursive parenting


withanorange:
GitHub issue version of [forum topic 35217](https://danbooru.donmai.us/forum_topics/35217).

Two posts ([10935774](https://danbooru.donmai.us/posts/10935774) and [10934881](https://danbooru.donmai.us/posts/10934881)) are completely broken edit-wise because they're recursively set as each other's parent. Attempting to edit the post in any way results in a guaranteed error. It likely comes from a race condition because a pixel-perfect duplicate bot attempted to set the parent of post 10934881 immediately (not like we haven't told the person running it to make it wait a bit or anything, to deaf ears).

The stack trace is as follows:

```
SystemStackError exception raised
app/models/post.rb:769:in 'Post::ParentMethods#child_height'
app/models/post.rb:770:in 'Enumerable#map'
... (the above two lines are repeated for a while)
app/models/post.rb:763:in 'Post::ParentMethods#parent_hierarchy_depth'
app/models/post.rb:1821:in 'Post::ValidationMethods#validate_parent_depth'
app/controllers/posts_controller.rb:82:in 'PostsController#update'
app/logical/rack_server_timing.rb:19:in 'RackServerTiming#call'
```

PR #6347, opened 2026-03-20
Fix duplicate post versions issues

### LINKED ISSUE #6341: Duplicate post versions when using artist commentary tags and doing a tag edit together
It's possible to create duplicate post versions when using the commentary tag option in the artist commentaries API and then performing a post tag edit right after. [In the case of bot users, this ends up creating countless bloat post versions for no good reason](https://danbooru.donmai.us/post_versions?search%5Bupdater_name%5D=hdk4.9).

The way this works is that a commentary tag option in the artist commentaries API is used, and then the post's tags are edited again using post_tag_string. Instead of merging the new tags into the post version created by using the tag option in the artist commentaries API, it creates a new version that has both the new tags and the previously added commentary tags.

It's another angle of creating duplicate post versions. Something else that comes to mind is #6273.

hdk5:
The first part of the fix for https://github.com/danbooru/danbooru/issues/6341.

1. If two concurrent post `tag_script` edits occur, the later edit may read a stale `tag_string` when merging. As a result, the later edit can overwrite the earlier one. This change adds a row lock for the post during the update transaction. Another suggestion, not implemented here, is to avoid creating a `PostEdit` entirely if the `tag_string` was not changed.
2. Messages sent to the archive service were serialized in ISO 8601 format with 1-second precision, and the service failed to find the matching version for merging because it used a strict `updated_at < ?` condition. The fix here is to increase the precision to the maximum supported by Ruby of 1ns. https://github.com/danbooru/archives/pull/2 will make it use a different approach to select the appropriate earlier version.

This change works "good enough" to fix #6341 and can be applied independently from the second half, or without the second half at all.


nonamethanks:
Reposting what evazion mentioned in the discord:
> it interferes with issue #6324
> for that issue every post in the parent chain needs to be locked in ID order, otherwise you can get deadlocks
> if edit A locks post 99 then post 100, and edit B locks post 100 then post 99, then you get deadlock
> A waits on post 100 to unlock while B waits on post 99 to unlock, and neither can progress because they're both waiting on each other 
> so you can't just lock the post, you have to find out if the parent is being updated then lock it first if it has a lower ID

PR #6348, opened 2026-03-23
Fix DCinside extractor

hdk5:
Spoof the correct referer header for http downloader.

Note that there's another issue with the host not supporting resuming downloads and closing the connection too early before the whole file is downloaded, which seems to be solvable by using a proxy close to Korea, but that's on infra, not the code.

### DIFF
--- a/app/logical/source/extractor/dc_inside.rb
+++ b/app/logical/source/extractor/dc_inside.rb
@@ -66,6 +66,6 @@ class Source::Extractor::DcInside < Source::Extractor
 
   def http_downloader
     # downloads are prone to being interrupted, especially for large files
-    super.use(retriable: { max_retries: 10 })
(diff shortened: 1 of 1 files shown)


evazion:
Downloads work for me without the referer.

I tried using a Korean residential proxy but it didn't seem to make much difference. The site works for me in the browser though, so maybe we need to do a better job of spoofing a browser.

hdk5:
Large downloads (5+ MB) fail for me in browser in the middle of downloading.
On danbooru, both real and local without referer spoof, it fails with 403, and before that - with "unexpected EOF", i.e. same as in browser.

evazion:
<https://gall.dcinside.com/mgallery/board/view/?id=zenless_zone_zero&no=322381> (20MB)

I can reproduce it now. A Korean residential proxy worked with this image, but residential proxies are relatively expensive for downloads because they charge per GB. I need to see if datacenter IPs will work.

Issue #6350, opened 2026-03-25
Prevent tabbing out of tag box


nottalulah:
Sometimes if autocomplete is too slow or I made a typo I'll accidentally tab out of the tag box instead of triggering autocomplete. It would be nice if the tag box could prevent tabbing about by mistake.

ghost:
Fixed by d62ce390a186ca8e4a79d09ef00219c633755943.

PR #6356, opened 2026-03-27
feat: change tab behavior with tag autocomplete (#6350)

steve-redka:
If menu is invisible and caret is mid-tag, prevents tab from moving focus. 

If menu is not visible and caret is at clean boundary, tab works normally.

Solves #6350

### DIFF
--- a/app/javascript/src/javascripts/autocomplete.js
+++ b/app/javascript/src/javascripts/autocomplete.js
@@ -296,21 +296,9 @@ Autocomplete.on_tab = function(event) {
       autocomplete.menu.select();
       autocomplete.close();
     }
-
-    event.preventDefault();
-    return;
   }
 
-  // If the caret is in the middle of a tag, prevent tab from moving focus
-  // out of the field. This avoids accidentally tabbing away when autocomplete
-  // is still loading.
-  let caret = input.selectionStart;
-  let term_before_caret = input.value.substring(0, caret).match(/\S*$/)[0];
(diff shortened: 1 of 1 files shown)


ghost:
This behavior is sort of confusing. It can feel like the tab button just doesn't work in some cases given it still has default behavior after a space. IMO it's better to just disable tabbing entirely. Not like there is any reason to since you submit the form with ctrl+enter anyway.

steve-redka:
Ok, it's disabled.

evazion:
This will disable tab inside DText boxes (comments, forum posts, wiki pages, etc) and inside search forms (e.g. the name field on the tag search page), even when you haven't typed anything yet and autocomplete isn't active. I'm not sure this is a good idea either.

I think a better fix would be to buffer tab until the autocomplete menu finishes loading. If you hit tab while the menu is loading, wait until it finishes loading before inserting the next tag. If you hit tab while autocomplete is inactive (the menu is closed and not loading), move focus to the next element as normal.

ghost:
I just want it disabled in the tag box, because when I'm tagging quickly I sometimes hit tab even when autocomplete would not be triggered (either because I already hit space or I typoed the tag or something like that). Other forms don't have the same problem because I'm not trying to type 100 words in 30 seconds. Given you can hit ctrl-enter to submit the form and tab in the tag box would switch input to the submit button, it didn't seem like much of an issue there.

evazion:
I changed it to only disable the tab key in the tag edit box. I experimented with buffering the tab key while autocomplete was loading, but I had trouble getting it to work properly.

Issue #6256, opened 2026-02-10
Allow "E" keyboard shortcut on single-term post searches


hdk5:
For search queries that show additional tab beside "Posts" (such as "Artist", "Wiki", "Pool" and maybe some others), pressing "E" will have the same effect as doing so on a wiki/artist/pool/etc. page directly, i.e. open the entry edit page.

BrokenEagle:
Also favgroups.

ghost:
Related #5175. Some of the shortcuts should be applied globally.

hdk5:
fixed by #6358

Issue #5175, opened 2022-05-17
More keyboard shortcuts


kidonng:
This is a bad title, but it get the message out. Here we goes:

- <kbd>q</kbd> to focus **every** search box. Despite listed under the "Global" section, it only works for posts for the moment. Wanna search on a tag or wiki page? Ask your cursor... no it should not.
- <kbd>meta</kbd> + <kbd>enter</kbd> to submit. I know there's <kbd>ctrl</kbd> + <kbd>enter</kbd> but it just feels wrong on macOS. Sounds niche, but if you would kindly accept a PR I'm glad to do it myself. The problem is that the current way of creating shortcuts doesn't seem to allow multiple combination for a single function.
- <kbd>shift</kbd> + <kbd>/</kbd> (i.e. inputing a `?`) opens the "Keyboard Shortcuts" page (`/static/keyboard_shortcuts`). This is a common trick used by many, like the very site you are looking at now. There are like hundreds of links on the "More" page impossible to go through each, and it took me several days to find a _wiki_ page introducing the shortcuts, it's not obvious as it appears to be.
- Navigation shortcuts, for links in the top column ("My Account", "Posts", "Comments" etc.). Again inspired by GitHub, something that can save numerous time jumping around.
- Shortcut to add to pool and add commentary. I believe these are actions common enough to warrant a shortcut.

With the above shortcuts added, using Danbooru will be a much better experience for keyboard lovers. Shouldn't be hard to implement (by myself) but I just want to throw this out in case you may be interested.

bipface:
for the convenience of anyone doing hotkey-related development/testing, here's a function to dump all the current registered hotkeys as tables into the console, grouped by target element:
```js
for (let tw = document.createTreeWalker(document,
	NodeFilter.SHOW_DOCUMENT|NodeFilter.SHOW_ELEMENT),
	el = tw.currentNode;
	el !== null;
	el = tw.nextNode())
{
	let infos = jQuery._data(el, 'events')?.keydown
		?.map?.(x => ({keys: x.data?.keys ?? '', name: x.namespace}));

	if (infos?.length) {
		console.group(el);
		console.table(infos);
		console.groupEnd();
	}
}
```

kidonng:
[...]

PR #6358, opened 2026-03-28
Add keyboard shortcuts for navigation, search, and post actions

steve-redka:
- q focuses search box on Tags and Wiki Pages (previously only worked on Posts despite being listed as "Global")
- Cmd+Enter (meta+return) submits forms on Mac, alongside existing Ctrl+Enter
- ? (Shift+/) opens the Keyboard Shortcuts help page
- g then X GitHub-style navigation sequences to jump to top nav links: g p Posts, g c Comments, g n Notes, g a Artists, g t - Tags, g l Pools, g w Wiki, g f Forum, g m My Account
- p on post page opens "Add to pool" dialog
- c on post page opens "Add commentary" dialog
- e on post search results opens edit page for the associated artist / wiki / pool / favorite group
- Updated Keyboard Shortcuts help page to document all new shortcuts

Related issues: #5175 #6256

### DIFF
--- a/app/javascript/src/javascripts/shortcuts.js
+++ b/app/javascript/src/javascripts/shortcuts.js
@@ -110,7 +110,17 @@ Shortcuts.hide_tooltips = function() {
 }
 
 Shortcuts.keyboard_shortcuts = function() {
-  window.location.href = "/static/keyboard_shortcuts";
+  $.get("/static/keyboard_shortcuts", function(html) {
+    let content = $(html).find("#a-keyboard-shortcuts");
+    $("<div>").append(content.children()).dialog({
+      title: "Keyboard Shortcuts",
+      width: 800,
+      modal: true,
+      close: function() {
+        $(this).dialog("destroy");
+      },
+    });
+  });
 }
 
 $(document).ready(function() {


TypeA2:
I'm not really a fan of the idea of having those navigation sequences on such "normal" keys, nor am I convinced they're really needed at all.

ghost:
> I'm not really a fan of the idea of having those navigation sequences on such "normal" keys, nor am I convinced they're really needed at all.

Yeah, they seem like they'd cause confusion for the average user. Especially if there's no feedback that this is a hotkey. Shift+/ should open in a dialog instead of redirecting you, that's also poor UX.

steve-redka:
Fair point about discoverability. That said, these navigation sequences are specifically a power-user feature - someone uploading hundreds of images benefits a lot from quick navigation without reaching for the mouse. GitHub uses the same "g + X" pattern for the same reason, so there's existing precedent users might already know.

> Shift+/ should open in a dialog instead of redirecting you, that's also poor UX.

Reworked that.

ghost:
> someone uploading hundreds of images benefits a lot from quick navigation without reaching for the mouse.

Power uploaders don't visit any page other than the upload pages. Even other power users don't regularly visit comments, notes, tags, or pools, and artists are usually edited directly from the artist page or create new artist link on the upload page.

steve-redka:
Ok.
Navigation hotkeys are in separate commit, so it can be dropped.

evazion:
I think the `e` to edit shortcut on the post index is likely to cause confusion when someone presses `e` by accident and ends up editing a wiki page instead. Using `data-shortcut-when` to only activate it when the wiki tab is open would be better. Or `Shift+E` so it can't be pressed by accident.

The nav shortcuts are unlikely to be used much but otherwise fine. Although it would be better to change `data-shortcut` so you could say `data-shortcut="g p"` and not have to special case these.

I've never really liked the `q` to search shortcut. It should be `/` or `Ctrl+K` instead. That's what's used by other sites. The only problem is that it conflicts with Firefox shortcuts.

evazion:
Just realized there are conflicts with the nav shortcuts on the post show page. If you press `g p`, that shows the favgroup dialog, the add to pool dialog, and navigates to the posts page.

steve-redka:
> Just realized there are conflicts with the nav shortcuts on the post show page. If you press `g p`, that shows the favgroup dialog, the add to pool dialog, and navigates to the posts page.

Damn

I'll fix it later

steve-redka:
Oh, they just use the same key for different things. This isn't fixable as is.
I guess there are two ways to move forward
- remove the g+x navigation
- Assign a different key (Shift+g, or \`)
Since two other maintainers were against adding g+x navigation in the first place, I'm leaning towards former.

Issue #5827, opened 2024-09-27
Allow to hide uploads


nonamethanks:
This has been one of the most requested features since the redesign of the upload system.
We could put them in a third tab here so that they're still available but removed from the main view.
![image](https://github.com/user-attachments/assets/53ad93dd-b33c-4abf-affd-4a248e35e6c5)

nottalulah:
#5070 related.

Zygarbodor:
This is genuinely my most-wanted feature for the site. I have so much clutter I want to get rid of, but I can't, since it's stuff I don't want to post. This includes:
- Accidental AI uploads
- Accidental banned artist uploads
- Duplicates
- Multi-image sets where I only intended on posting some of the images, not all
- Stuff I originally wanted to post, but changed my mind

I know there's the "hide uploads" unserscript, but that only makes the item(s) invisible, rather than completely removed. I've had to switch to displaying 200 items per pages just so I don't have to look through countless blank pages to find something. To put things in perspective, as of the time I'm writing this, I have exactly 401 items in my uposted tab, 361 of which are hidden. That's how much unpostable garbage is clogging up my assets. It's genuinely ridiculous.

Adding a third tab in "My Uploads" definitely seems like the best way to go about this. The assets would still be on the site, but they wouldn't be in the way anymore. I really do hope this gets implemented someday, as it'd be an amazing quality-of-life feature that would completely fix my biggest issue with the site.

Issue #6028, opened 2025-06-09
Add upload folders


nottalulah:
As a potential alternate solution to #5827, #5070, #5127, etc etc consider a system to allow users to put uploads into folders, kind of like dmails have. This avoids potential confusion with "delete"/"hide" (since "deleted" uploads would still be visible in /media_assets) and is more versatile. It could be used for any organization the user wants, or for attachments in forums/comments/etc if #6008 got implemented.

Example implementation could be a `folder` column on the uploads table, default to showing only uploads with `null` folders, some JS widget to facilitate moving uploads to another/no folder.

PR #6361, opened 2026-03-28
posts: add ability to hide posts

steve-redka:
Puts hidden posts into a third tab where they are still available but removed from main view.

I'm not sure how the frontend should behave upon hiding/unhiding post - currently it stays on page and toggle switches state. I could make it disappear, but I think that would make handling misclicks more annoying.

Screenshot from 2026-03-28 21-23-34

Solves issue: #5827

### DIFF
--- a/app/controllers/upload_media_assets_controller.rb
+++ b/app/controllers/upload_media_assets_controller.rb
@@ -27,4 +27,10 @@ class UploadMediaAssetsController < ApplicationController
       respond_with(@upload_media_asset)
     end
   end
+
+  def update
+    @upload_media_asset = authorize UploadMediaAsset.find(params[:id])
+    @upload_media_asset.update(permitted_attributes(@upload_media_asset))
+    respond_with(@upload_media_asset)
+  end
 end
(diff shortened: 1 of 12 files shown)


ghost:
This implementation was proposed before, but we didn't go with it because
1. Uploads should be immutable
2. The hidden/deleted terminology is misleading, since the media asset is still publicly viewable
3. It's inflexible (#6028)

steve-redka:
> Uploads should be immutable

Is there some sort of document which lists all design guidelines? 

Also, what do you mean by "immutable"? Should it literally never change (which isn't the case - uploads have `status` column which is supposed to change), or should they just not be otherwise changeable by users?

ghost:
> > Uploads should be immutable
> 
> Is there some sort of document which lists all design guidelines?

Not really, it's just years of discussions from the forums, GitHub issues, or Discord.

> Also, what do you mean by "immutable"? Should it literally never change (which isn't the case - uploads have `status` column which is supposed to change), or should they just not be otherwise changeable by users?

After they're finished (i.e. in the `error` or `completed` state) they should not be updated further. Uploads (and upload media assets) should serve as a fixed log of uploads the user has made or attempted.

steve-redka:
Well, I'd like to help, but I don't have access to Discord. It would be useful to have a single source of truth regarding design philosophy to avoid running into such issues in the future.

PR #6380, opened 2026-04-06
comment votes: search by body instead of id

### LINKED ISSUE #6369: "comment" in comment vote search form is for the id, when it should be for the comment text
Image

I would consider this a bug because it would seem obvious that you could just type the text of the comment to find matching comments, and yet that field searches for the comment_id.

withanorange:
Fixes #6369. Replaces the `comment_id` search field with a `body_matches` search field, form order remains the same. I picked `body_matches` to keep parity with normal comment searching which also uses `body_matches` in its form.

### DIFF
--- a/app/views/comment_votes/_search.html.erb
+++ b/app/views/comment_votes/_search.html.erb
@@ -4,7 +4,7 @@
     <%= fc.input :creator_name, label: "Commenter", input_html: { value: params.dig(:search, :comment, :creator_name), "data-autocomplete": "user" } %>
     <%= fc.input :body_matches, label: "Text", input_html: { value: params.dig(:search, :comment, :body_matches ) } %>
     <%= fc.input :post_tags_match, label: "Tags", input_html: { value: params.dig(:search, :comment, :post_tags_match), "data-autocomplete": "tag-query" } %>
-    <%= fc.input :post_id, label: "Post", input_html: { value: params.dig(:search, :comment, :post_id) } %>
+    <%= fc.input :post_id, label: "Post ID", input_html: { value: params.dig(:search, :comment, :post_id) } %>
   <% end %>
   <%# <%= f.input :comment_id, label: "Comment", input_html: { value: params[:search][:comment_id] } %>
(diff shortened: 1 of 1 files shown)


withanorange:
Converting this into a draft and will do a bit more wordsmithing on some more forms to go along with it, just so it can all ship at once instead of picking them off in parts every now and again.

nonamethanks:
I would recommend just submitting pulls for smaller fixes one by one, if you try to fix everything at once you'll just be swamped (i made this mistake many times). Can you open a pull specifically to fix that issue instead?

withanorange:
I think what I could do instead then is just convert the PR information back to its original state and take it out of draft. I remember I expanded the scope mainly from a recommendation to do it together but I kind of forgot to continue after puzzling over some choices and just left this to die.

app/views/comment_votes/_search.html.erb
```diff
@@ -2,10 +2,11 @@
   <%= f.input :user_name, label: "Voter", input_html: { value: params[:search][:user_name], "data-autocomplete": "user" } %>
   <%= f.simple_fields_for :comment do |fc| %>
     <%= fc.input :creator_name, label: "Commenter", input_html: { value: params.dig(:search, :comment, :creator_name), "data-autocomplete": "user" } %>
+    <%= fc.input :body_matches, label: "Text", input_html: { value: params.dig(:search, :comment, :body_matches ) } %>
     <%= fc.input :post_tags_match, label: "Tags", input_html: { value: params.dig(:search, :comment, :post_tags_match), "data-autocomplete": "tag-query" } %>
-    <%= fc.input :post_id, label: "Post", input_html: { value: params.dig(:search, :comment, :post_id) } %>
+    <%= fc.input :post_id, label: "Post ID", input_html: { value: params.dig(:search, :comment, :post_id) } %>
   <% end %>
-  <%= f.input :comment_id, label: "Comment", input_html: { value: params[:search][:comment_id] } %>
+  <%# <%= f.input :comment_id, label: "Comment", input_html: { value: params[:search][:comment_id] } %>
```
nonamethanks: Just remove the comment, no point leaving it.

Issue #5240, opened 2022-08-22
Prune/Scheduled Tasks Causing Issues on Testbooru With User Permissions


BrokenEagle:
Users given the approval privilege on Testbooru quickly lose it as it's not possible to approve 30 posts every 45 days, nor should it be required on a test platform like Testbooru.

https://testbooru.donmai.us/user_feedbacks?commit=Search&search%5Bbody_matches%5D=Lost+approval+privileges

That and perhaps other scheduled tasks should be able to be configured so that they don't run when not needed or desired. I propose 2 solutions below which could be used to address this.

### 1. Database table

Adding scheduled tasks to a database table would allow one to disable/enable tasks as needed. This could even be an Admin privilege, and given an HTML/JSON endpoint to query/update them. I do something similar on my Prebooru instance.

![image](https://user-images.githubusercontent.com/21149935/185948825-667c28d3-e2ca-4a3c-a90c-8e6a46a84e71.png)

### 2. Config file

Another option would be to add something to the config files which would enable/disable a scheduled task. This would be less flexible than the prior solution, however it would be much easier to implement.

PR #6381, opened 2026-04-07
Add config option to disable approver pruning.

nottalulah:
Mainly for smaller boorus or things like testbooru to avoid people being demoted in cases where they wouldn't be expected to approve a lot.

I also added word_wrap to the warning message because currently it looks bad, though the demotion message doesn't wrap at all so I didn't change that one.


nottalulah:
Forgot to mention but would've closed #5240.

PR #6392, opened 2026-04-11
Notes: Replace Blackletter font with Manufacturing Consent.

### LINKED ISSUE #5189: Replace Current Blackletter Font for Better Readability
The current blackletter font is mostly unreadable, which is why it's rarely being used right now. Therefore, the old blackletter should be replaced with a different font. After searching around some and polling other users ([forum #209043](https://danbooru.donmai.us/forum_posts/209043)), I found a few better blackletter fonts which are readable that can be used to replace the current one.

- https://www.1001freefonts.com/alpine.font
- https://www.1001freefonts.com/broken-woodtypes.font
- https://www.1001freefonts.com/fenwick-woodtype.font

I'm more partial to the alpine font due to it's increased readability, but as another user pointed out, the latter two are more like a traditional blackletter font. There's also several other candidate fonts listed on the forum post that was listed above.

**Note::** All of the fonts listed above and on the forum post are 100% free to use.

nottalulah:
Replace the existing UnifrakturMaguntia font with Manufacturing Consent for legibility. Some of the glyphs in Unifraktur look nothing like the letter they're meant to represent which makes it difficult to read notes using the font.

This replacement was suggested after discussion on Discord and [the forum](https://danbooru.donmai.us/forum_posts/430241).

Old:
image

New:
image

Fixes #5189.

### DIFF
--- a/app/javascript/src/styles/common/fonts.css
+++ b/app/javascript/src/styles/common/fonts.css
@@ -302,14 +302,14 @@
     url("@fontsource/petit-formal-script/files/petit-formal-script-latin-400-normal.woff") format("woff");
 }
 
-/* https://fonts.google.com/specimen/UnifrakturMaguntia */
+/* https://fonts.google.com/specimen/Manufacturing+Consent */
 @font-face {
   font-family: "Blackletter";
   font-display: swap;
   src:
-    local("UnifrakturMaguntia"),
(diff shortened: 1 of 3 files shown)


evazion:
This will break the formatting of all existing notes using this font because the letter sizes will be different. All notes using this font will have to be checked and updated if necessary.

https://danbooru.donmai.us/notes?search[body_matches]=*blackletter*&search[group_by]=note

PR #6398, opened 2026-04-17
Use pwnedpasswords for password security

steve-redka:
Adds a validator that checks new passwords against the HaveIBeenPwned PwnedPasswords API. If a password has appeared in a known breach, signup or password change is rejected with the error "has appeared in a data breach and can't be used".

The API uses k-anonymity: we SHA1 the password, send only the first 5 characters of the hash, and the API returns all matching suffixes. The full password and full hash never leave the server. No API key is required.

Responses are cached for 1 hour. If the API is unreachable or slow, the validator fails open so outages don't block signups.

Related issue: #6608

### DIFF
--- a/app/logical/pwned_password_validator.rb
+++ b/app/logical/pwned_password_validator.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+# Validates that a password has not appeared in known data breaches using the
+# HaveIBeenPwned PwnedPasswords API.
+#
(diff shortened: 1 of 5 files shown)


nottalulah:
This should probably be disable-able via config since it makes web requests. I can imagine some self-hosted boorus would not want that.

steve-redka:
I made it disableable, but wouldn't it make it more sense to make it off by default?

PR #6399, opened 2026-04-17
Use zxcvbn for password security#6398

steve-redka:
Adds a validator that estimates password strength using [zxcvbn](https://github.com/envato/zxcvbn-ruby), a pattern-matching password strength estimator originally from Dropbox. It catches common words, keyboard patterns ("qwerty"), short sequences ("abc123"), dates, and l33t speak that simple length rules miss. Passwords scoring below 2 on the 0-4 scale are rejected with an error that includes zxcvbn's own feedback ("This is a top-10 common password", "Add another word or two", etc).

Default minimum score is 2 - "protect from unthrottled online attacks". Can be overridden per-use-case via validates :password, zxcvbn_password: { min_score: 3 } for stricter checking.

Many tests used passwords like "password", "xxxxx1", "12345", "hunter2" etc. that no longer pass validation. These were replaced with strong fixtures. Two unrelated users_controller_test.rb tests that were already failing on master (DNS lookup for email deliverability) remain failing - not introduced by this PR.

Related issue: #6608

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -59,6 +59,7 @@ gem "bcrypt_pbkdf"
 gem "connection_pool"
 gem "rotp"
 gem "rqrcode"
+gem "zxcvbn-ruby", require: "zxcvbn"
 gem "kramdown"
 gem "abbrev"
 gem "rubyzip", require: false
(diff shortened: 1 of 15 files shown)


steve-redka:
This one runs locally, so I'm not sure if you are supposed to be able to disable this.

Issue #6363, opened 2026-03-29
Make rate limits page prettier


nonamethanks:
<img width="1101" height="790" alt="Image" src="https://github.com/user-attachments/assets/155e9b67-423e-4534-b5bb-061958a0b8b9" />

Right now it doesn't explain much for non-technical users. it could be englicized so that each entry says something like "you have made 1 out of 200 edits in a hour for Posts", or "you have 199 edits left for posts in the next hour", or even fix the table so that it has "used actions", "actions left", "total actions", and the timeframe.

It should also be linked from the user profile in a place that's visible.

steve-redka:
I looked into this. The rate limits page exists at /rate_limits but it's not linked from anywhere - you have to know the URL. So that's also a problem.

The main challenge with showing "used/remaining/total" is that the rate_limits table only stores the current points balance. The burst (max capacity) and rate (refill speed) values aren't persisted - they're computed dynamically per-request from the Pundit policies, where they vary based on user level, account age, and activity history.

To show something like "5 of 200 used" or "195 remaining", we'd need to store burst and rate in the table. They're already passed into RateLimit.create_or_update!, so it would just be a migration adding two float columns and updating the upsert SQL to persist them.

Is adding burst and rate columns to rate_limits acceptable, or would you prefer the no-migration approach?

ghost:
IMO we should prefer without migrations where it's reasonable to do so. I don't think persisting the burst/rate values would be necessary for performance, so the values should be calculable from the policy class data.

PR #6400, opened 2026-04-17
Make rate limits page readable for non-technical users

steve-redka:
The `/rate_limits` page currently shows raw technical columns (`action`, `key`, `points`, `limited?`) and isn't linked from anywhere visible, so most users never find it and can't make sense of it if they do.

This PR:

- Adds a Rate Limits row to the user profile (next to the API Key row) so the page is actually reachable.
- Replaces the raw columns with: humanized action, usage (`X of Y left`), refill rate, status pill (OK / Limited), and last used.
- The owner still sees the raw `key` as a subtext under each action for debugging.

The `rate_limits` table only stores `points`, not the `burst`/`rate` the bucket was configured with. Per maintainer guidance I avoided a migration and instead resolve these at render time: for each row, `RateLimit#limit_config(user)` looks up the policy class by controller name, invokes the matching `rate_limit_for_*` method, and returns `{burst:, rate:}`.

Caveats worth knowing:

- For policies that branch on `record.invalid?` or specific record state, I pass a fresh `.new` record. This may not match the branch that was actually in effect when the bucket was last updated, so the displayed burst/rate is "what applies right now for a typical action of this type" rather than a historical replay.
- When the policy method raises or the policy/record class can't be resolved, the column falls back to `~N left` (just `points`) and the Refills column shows a dash.

Screenshot from 2026-04-17 17-49-53

#6363

### DIFF
--- a/app/models/rate_limit.rb
+++ b/app/models/rate_limit.rb
@@ -76,4 +76,66 @@ class RateLimit < ApplicationRecord
 
     RateLimit.find_by_sql([sql, sql_params])
   end
+
+  # A human-readable version of the `action` string. Record-specific suffixes
+  # like ":post-123" or ":wiki-page-45" are stripped; non-id modifiers like
+  # ":invalid" or ":large" are preserved as parenthesized qualifiers.
+  def humanized_action
+    return "Unknown" if action.blank?
+
+    controller, action_name, modifier = parsed_action
+    return action.humanize if controller.blank? || action_name.blank?
+
+    label = "#{controller.humanize}: #{action_name.humanize(capitalize: false)}"
+    label += " (#{modifier.humanize(capitalize: false)})" if modifier.present? && !record_suffix?(modifier)
+    label
+  end
+
+  # Returns the `{ burst:, rate: }` values currently applicable to this rate
+  # limit's action when invoked by `user`, by calling into the corresponding
+  # Pundit policy. Returns nil if the policy/action can't be resolved or the
+  # method raises (e.g. record-dependent branches with no available record).
+  def limit_config(user)
+    controller, action_name, = parsed_action
+    policy_class = resolve_policy_class(controller)
+    return nil unless policy_class && action_name.present?
+
+    record = build_dummy_record(controller)
+    policy = policy_class.new(user, record)
+    method = :"rate_limit_for_#{action_name}"
+    return nil unless policy.respond_to?(method)
+
+    config = policy.send(method, request: nil)
+    return nil if config.blank? || config[:burst].nil? || config[:rate].nil?
+
+    { burst: config[:burst].to_f, rate: config[:rate].to_f }
+  rescue StandardError
+    nil
+  end
+
+  private
+
+  def parsed_action
+    parts = action.to_s.split(":")
+    [parts[0], parts[1], parts[2..]&.join(":")]
+  end
+
+  def record_suffix?(modifier)
+    modifier.match?(/\A[a-z-]+-\d+\z/)
+  end
+
+  def resolve_policy_class(controller)
+    return nil if controller.blank?
+
+    ["#{controller.classify}Policy", "#{controller.camelize}Policy"].each do |name|
+      return name.constantize if Object.const_defined?(name)
+    end
+    nil
+  end
+
+  def build_dummy_record(controller)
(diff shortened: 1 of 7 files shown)


app/models/rate_limit.rb
```diff
@@ -76,4 +76,66 @@ def self.create_or_update!(action:, keys:, cost:, rate:, burst:, minimum_points:
...
+  rescue StandardError
+    nil
+  end
+
+  private
+
+  def parsed_action
+    parts = action.to_s.split(":")
+    [parts[0], parts[1], parts[2..]&.join(":")]
+  end
```
nottalulah, line 121: This can just be `action.to_s.split(":", 3)`. The second parameter limits how many things to return. Also the third item seems to do nothing.

app/views/rate_limits/index.html.erb
```diff
@@ -1,17 +1,46 @@
...
   <div id="a-index">
+    <p class="prose mb-4">
+      Rate limits cap how many times you can perform certain actions within a given period. Each bucket refills over time, so unused capacity returns if you take a break.
+    </p>
+
     <%= table_for @rate_limits, class: "striped autofit" do |t| %>
-      <% t.column :action %>
+      <% t.column "Action" do |rate_limit| %>
+        <%= rate_limit.humanized_action %>
+        <% if CurrentUser.user.is_owner? %>
```
nottalulah, line 10: I don't think this needs to be limited to owner. It would also make it unclear why `Posts: create` is showing up three times. Maybe they could be deduped instead.

app/models/rate_limit.rb
```diff
@@ -76,4 +76,66 @@ def self.create_or_update!(action:, keys:, cost:, rate:, burst:, minimum_points:
...
+  # limit's action when invoked by `user`, by calling into the corresponding
+  # Pundit policy. Returns nil if the policy/action can't be resolved or the
+  # method raises (e.g. record-dependent branches with no available record).
+  def limit_config(user)
+    controller, action_name, = parsed_action
+    policy_class = resolve_policy_class(controller)
+    return nil unless policy_class && action_name.present?
+
+    record = build_dummy_record(controller)
+    policy = policy_class.new(user, record)
```
nottalulah, line 104: These extra helper functions just seem to be a roundabout way of doing `safe_constantize`:
```suggestion
    controller_name, action_name = action.to_s.split(":", 3)
    return nil unless controller_name.present? && action_name.present?

    dummy_record = controller_name.classify.safe_constantize&.new
    policy = dummy_record&.policy(user)
```
`Pundit.policy` seems to already be capable of handling `MediaMetadata` which is the only problematic class name that requires the difference between `classify` and `camelize`.

(also I don't think `action` can be `nil`, so `to_s` may be redundant).

app/views/rate_limits/index.html.erb
```diff
@@ -1,17 +1,46 @@
...
+          <%= "#{remaining} of #{burst} left" %>
+        <% else %>
+          <%= "~#{[rate_limit.points, 0].max.round} left" %>
+        <% end %>
+      <% end %>
 
-      <% t.column :points do |rate_limit| %>
-        <%= rate_limit.points.round(2) %>
+      <% t.column "Refills" do |rate_limit| %>
+        <% config = rate_limit.limit_config(CurrentUser.user) %>
```
nottalulah, line 27: Doing `config = ...` multiple times is bad. Hoist it to the top of the loop if it's needed.

app/models/rate_limit.rb
```diff
@@ -76,4 +76,66 @@ def self.create_or_update!(action:, keys:, cost:, rate:, burst:, minimum_points:
...
+    method = :"rate_limit_for_#{action_name}"
+    return nil unless policy.respond_to?(method)
+
+    config = policy.send(method, request: nil)
```
nottalulah, line 108:
```suggestion
    config = policy.try(method, request: nil)
```
Should be equivalent.

nottalulah (changes requested):

nottalulah:
TBH kinda having doubts whether no-migration route was optimal given it seems without it half the routes will be broken anyway.

steve-redka:
I addressed the issues in the review. But regarding migrations - I think it should be a separate issue, if it's really that important.

PR #6435, opened 2026-05-24
Yachiyo's room extractor

TypeA2:
http://yachiyo-room.com/

It's an oekaki-oriented site specifically for the Cho Kaguya-hime! copyright. It's mostly pretty basic but there's some worthwhile stuff in there, I think.

The site is pretty barebones and doesn't use credentials so it should be pretty low-upkeep too.


evazion:
Run `rubocop` on this, it has various formatting issues. Indentation should be two spaces not four.

TypeA2:
My bad, forgot to install it in my new dev environemnt, should be fixed now

PR #6455, opened 2026-06-07
Modularize BUR logic

nonamethanks:
This is a very early WIP to move BUR commands to their own classes to ease future development.

There's still a lot of convoluted logic due to how it was all under the same class, so there's a lot of simplification to be done yet.


app/logical/bulk_update_request/command/mass_update.rb
```diff
@@ -0,0 +1,45 @@
...
+  def to_query
+    PostQuery.new(@second_search)
+  end
+
+  def affected_tags
+    from_query.tag_names + to_query.tag_names
+  end
+
+  def process!(**)
+    BulkUpdateRequest::Processor.mass_update(@first_search, @second_search)
```
ghost: I don't think this should be implemented in the Processor class. It would make more sense to do `BulkUpdateRequest::Command::MassUpdate.new(x, y).process!` elsewhere, or have `BulkUpdateRequest::Processor.mass_update` call `BulkUpdateRequest::Processor::MassUpdate`.

app/logical/bulk_update_request/command.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def process!(**)
+    raise NotImplementedError
+  end
+
+  # @param line [String] a single line from a BUR script
+  # @return [BulkUpdateRequest::Command] a command instance, or InvalidLine if the line doesn't match
+  def self.parse(line)
+    commands.each do |kommand|
+      match = kommand.regex.match(line)
```
ghost: Why is this `kommand` instead of `command`?

app/logical/bulk_update_request/processor.rb
```diff
@@ -0,0 +1,137 @@
+# frozen_string_literal: true
+
+# Process a bulk update request. Parses the request and applies each line in
+# sequence.
+class BulkUpdateRequest::Processor
+  # Maximum tag size allowed by the rename command before an alias must be used.
+  MAXIMUM_RENAME_COUNT = 200
```
ghost: This should be moved to the Rename command class since it's only relevant there.

app/logical/bulk_update_request/processor.rb
```diff
@@ -0,0 +1,137 @@
...
+  # * Both tags have less than 200 posts.
+  def self.is_tag_move_allowed?(antecedent_name, consequent_name)
+    antecedent_tag = Tag.find_by_name(Tag.normalize_name(antecedent_name))
+    consequent_tag = Tag.find_by_name(Tag.normalize_name(consequent_name))
+
+    antecedent_allowed = antecedent_tag.present? && antecedent_tag.artist? && antecedent_tag.post_count < MAXIMUM_BUILDER_MOVE_COUNT
+    consequent_allowed = consequent_tag.nil? || consequent_tag.empty? || (consequent_tag.artist? && consequent_tag.post_count < MAXIMUM_BUILDER_MOVE_COUNT)
+
+    antecedent_allowed && consequent_allowed
+  end
```
ghost: We should have a `BulkUpdateRequest::Command#approval_level` or such function instead, where each command can independently compute a level required. This would let us expand it easier, say if we want to add long-video-upload BURs or replacements approvable by mods or approvers respectively, or something else with lower-than-admin permissions.

Then we can just do something like `commands.max_by(&:approval_level)` to check whether the user is allowed to approve the BUR.

app/logical/bulk_update_request/command/mass_update.rb
nonamethanks: We have TagMover for tag aliases, so I was thinking that we could maybe have a MassUpdater or MassPostMover or something for mass updates, since it's used in multiple places, so that we could keep Processor free of BUR logic and only let it deal with command ordering.

app/logical/bulk_update_request/command.rb
nonamethanks: Because it's still a wip, I'll do a general adjustment of variable names once it's ready for review.

app/logical/bulk_update_request/processor.rb
nonamethanks: Hm, so a method on each command? I didn't do that  at first because it's just duplicate code for rename & alias, but the problem is that if we want a BUR to be approvable only if multiple commands pass an inclusive checklist, it becomes harder.

ghost: Yeah, it would be duplicated for BURs with similar requirements. I'm just thinking that we've had previous mentions of things like approvers being able to handle replacement BURs, having long video upload BURs, or having builders able to create implications from costumes to their base character tag. In those cases we'd have to have some function that returns whether they're approvable by a user or not for rendering the Approve button, so having a function that returns the minimum level required to approve it would probably be simplest.

nonamethanks:
I'll give this another pass in a few days to see what else is missing, but it should be close to finish.

PR #6475, opened 2026-06-28
Fix NoMethodError on xhs uploads.

nottalulah:
I can't run tests (I don't have a XHS account), so I don't know if this fixes all problems with XHS, but they seem to originate from here anyway.


nonamethanks:
Closed in favor of pull #6483.

PR #6478, opened 2026-06-30
Fix #5675: Feature Request for Private Community Boards

### LINKED ISSUE #5675: Feature Request for Private Community Boards
A way to make a private Booru that is invitation-based or with a register token without allowing users who are not logged in to see posts

Describe the solution you'd like

1. When registering, enter a token that can be set in the admin settings
2. That guests without an account cannot see posts

phsm:
Implemented two config knobs:
- force_authenticated: only authenticated users can access any page (except ones related to login, password reset, 2fa verify etc.)
- signups_restricted_to_admin: turn off the ability to self-register. Only admin (or higher) can register users.

Both of these knobs are false by default, so the original behavior is preserved.

How it is done:
- The application controller gets a new gate `force_authenticated` that is executed for every page. With the `force_authenticated` config knob set to false, it is skipped. 

- To bypass the restriction for certain pages (health checks, password reset, 2fa verification), `skip_force_authenticated` is introduced and used in those views.

- UserPolicy is adjusted to either allow the anonymous users to sign up, or limit it to admin users, depending on the `signups_restricted_to_admin` config value.

- The templates mentioning Sign Up link are edited not to show it when `signups_restricted_to_admin` is set.

I am open to feedback and suggestions. 

Disclaimer: I had not worked with RoR before (I occasionally write in ruby, though), so I leveraged an LLM to produce this. I have read and understood every line in the change though. 

Upd: if this feature feels too big to merge (I understand that Danbooru itself has no plans to use it), then reverting the template changes might also be an option.

### DIFF
--- a/app/policies/ai_tag_policy.rb
+++ b/app/policies/ai_tag_policy.rb
@@ -1,10 +1,6 @@
 # frozen_string_literal: true
 
 class AITagPolicy < ApplicationPolicy
-  def index?
-    true
-  end
-
   def tag?
     unbanned? && record.post.present? && policy(record.post).update?
   end
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -9,6 +9,7 @@ class ApplicationPolicy
   end
 
   def index?
+    return false if Danbooru.config.force_authenticated? && user.is_anonymous?
     true
   end
 
--- a/app/policies/background_job_policy.rb
+++ b/app/policies/background_job_policy.rb
@@ -1,10 +1,6 @@
 # frozen_string_literal: true
 
 class BackgroundJobPolicy < ApplicationPolicy
-  def index?
-    true
-  end
-
   def update?
     user.is_admin?
   end
--- a/app/policies/email_notification_policy.rb
+++ b/app/policies/email_notification_policy.rb
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 
 class EmailNotificationPolicy < ApplicationPolicy
+  # Override the parent method from ApplicationPolicy to ensure that users can always access the unsubscribe links.
+  # This is a public-facing endpoint that needs to remain accessible regardless of the force_authenticated setting.
+  def show?
+    true
+  end
+
   def destroy?
     true
   end
--- a/app/policies/forum_post_policy.rb
+++ b/app/policies/forum_post_policy.rb
@@ -1,10 +1,6 @@
 # frozen_string_literal: true
 
 class ForumPostPolicy < ApplicationPolicy
-  def index?
-    true
-  end
-
   def show?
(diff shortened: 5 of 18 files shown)


notWolfxd:
`force_authenticated` duplicates the logic for policies? 

You can probably just modify `app/policies/application_policy.rb` index method. Policies for sessions, 2fa, password resets, etc already have explicit sets to anonymous users.

phsm:
Thanks for the suggestion: indeed using policies is much cleaner (no need for "skip" override). 

I updated PR with the following:
- Switched to use application policy
- Reverted almost every template change: since all the pages show 403 anyway, why hide the links. The only template change proposed is to show "Sign Up" link under `/users` URL for the admin users when `signups_restricted_to_admin` is set.
- Tests are written.

I ran the entire test suite, got 89 failed tests, all are related to external sites (extractors tests). I guess this is expected.

phsm:
Update: 

During my tests I've found some pages that were not affected by the `ApplicationPolicy` index method change. 
All of them had `ApplicationPolicy` as a parent class but had their index method redefined with unconditional `return true`. 
I removed the index method from them, so now they use the parent method (which by default returns true). 

The list of policies I updated:
- `ai_tag_policy.rb`
- `background_job_policy.rb`
- `forum_post_policy.rb`
- `forum_topic_policy.rb`
- `media_asset_policy.rb`
- `media_metadata_policy.rb`
- `rate_limit_policy.rb`
- `user_event_policy.rb`

Maybe I don't understand something there, and those methods were redefined for a reason? Any advice is appreciated.

notWolfxd:
Look around the policies for them. There's redundant `index` methods in their policies, which can probably be stripped since they also duplicate application policy's index method.

phsm:
Yes, that's what I did after noticing the first such URL - I grepped for the `index?` method in `app/policies?`. Some files did have logic in the index method, and I did not modify them. The others, that I listed above, only had an empty method, those files are now modified. 

I updated the pull request with the newest changes.

nonamethanks:
Before going through the pull: the main problem with merging something like this is that we basically have to support it in perpetuity. 

This is the kind of feature that is very easily breakable by new endpoint addition, it's not a matter of when but if, because we are not going to be focused on mantaining features that the main site will never use. 

Also, you're trying to introduce two pretty big features at once in this pull, and neither will ever be used by the main site. These two features work in tandem, so why are they separate? And the workflow for new user creation does not make much sense to me. Are you expecting admins to manually create users and then give account passwords away? I don't think that's acceptable.

If you want a completely locked system, it would probably make more sense to restrict vision to verified users (members) rather than registered users. Basically, all new accounts start as restricted, and the only way to view stuff is be promoted to member by an admin. That way you avoid direct password handling. But again, that's something that you'd have in your own fork, we're not really interested in having that kind of logic in the main repo because it would just be a feature we'd have to mantain forever despite never using it.

Issue #4519, opened 2020-06-17
Add "n/m characters remaining" character counter to the appeal reason


luni3359:
The site only warns you after you have sent the reason, it should let you know beforehand.
![Reason](https://cdn.discordapp.com/attachments/710192137878241372/722855795636830238/unknown.png)

BrokenEagle:
Related to #4429.

lllusion3469:
There's e.g. HTMLs `<textarea required maxlength="140">` / `<input required maxlength="140">`
(`minlength` only applies if not empty)

This could be combined with the [`:valid`](https://developer.mozilla.org/en-US/docs/Web/CSS/:valid) / `:not(:valid)` pseudo-class and maybe [`::after`](https://developer.mozilla.org/en-US/docs/Web/CSS/::after) if you want to show text, though this wouldn't show the current length (which was requested).

One problem, though, is that HTML/JS count in "UTF-16 code units" whereas ruby counts in Unicode code points, so if characters outside the [BMP](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane) are used it might get rejected even if it's valid in ruby. If you want the number of code points in JS, you have to do it [like this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length#Unicode)

evazion:
HTML5 client-side validations like `maxlength` are preferable to writing our own Javascript widgets. See https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation.

SimpleForm is supposed to generate client side validations by default, but we had to disable it before because a bug in SimpleForm caused it to generate wrong `minlength` validations on everything (https://github.com/heartcombo/simple_form/issues/1496). I think that bug has been fixed, so we could probably enable it again.

As for calculating string length, technically we should count grapheme clusters rather than bytes or code points. See https://hsivonen.fi/string-length/:

```
# ruby
"🤦🏼‍♂️".grapheme_clusters.size # => 1
"🤦🏼‍♂️".codepoints.size # => 5
"🤦🏼‍♂️".bytes.size # => 17
"🤦🏼‍♂️".size # => 5

# javascript
"🤦🏼‍♂️".length # => 7
[..."🤦🏼‍♂️"].length # => 5
```

I don't think anyone - browsers, Javascript, or Rails - counts characters correctly by default.

lllusion3469:
[...]

PR #6485, opened 2026-07-14
Fix #5914: shrink oversized Edit Appeal reason text box

### LINKED ISSUE #5914: Shrink size of "Edit Appeal" text box
The size of the Reason text box on the Edit Appeal page is very big, but appeals only allow 140 characters:

Image

Consider also adding a live character counter so limit isn't unexpectedly exceeded: #5913

steve-redka:
## Summary

The Edit Appeal reason field renders as a single-line `<input type="text">` (`PostAppeal` declares `dtext_attribute :reason, inline: true`), and the site's global simple_form CSS forces all text inputs to `width: 100%`. Since the Edit Appeal page is full-width, the 140-character reason field stretched across the whole page.

This constrains the field's wrapper to `max-w-480px` so it renders at a reasonable size instead.

This does not add a live character counter. That's tracked separately in #4519.

### DIFF
--- a/app/views/post_appeals/edit.html.erb
+++ b/app/views/post_appeals/edit.html.erb
@@ -7,7 +7,7 @@
     </p>
 
     <%= edit_form_for(@post_appeal) do |f| %>
-      <%= f.input :reason, as: :dtext, input_html: { autofocus: true } %>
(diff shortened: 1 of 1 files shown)


nonamethanks:
Doesn't apply anymore.

PR #6418, opened 2026-05-09
Fix version history diff rendering

NekoAria:
Fixes #4788

`diff_body_html` treated every CJK/Hangul character as its own token (old regex only matched `\w+`), making diffs on Japanese/Korean/Chinese text unreadable. `diff_name_html` diffed the entire string character-by-character instead of isolating the changed portion.

- Tokenize CJK/Katakana/Hiragana/Hangul as atomic tokens
- Trim common prefix/suffix from names before diffing
- Skip the expensive diff when texts are >85% different (with a 5k-char cap to keep Levenshtein fast)
- Extract `format_body_text` / `levenshtein_similarity` helpers

Merge to betabooru first for testing against real data.

#### 1. CJK / Hangul tokenization

PixPin_2026-05-09_23-57-36
PixPin_2026-05-09_23-57-41
PixPin_2026-05-09_23-58-25
PixPin_2026-05-09_23-58-32
PixPin_2026-05-10_00-01-31
PixPin_2026-05-10_00-01-37

#### 2. Long name with small change

PixPin_2026-05-10_00-00-15
PixPin_2026-05-10_00-00-20
PixPin_2026-05-10_00-00-58
PixPin_2026-05-10_00-01-07

#### 3. Very different long bodies (wholesale replace)

PixPin_2026-05-09_23-57-23
PixPin_2026-05-09_23-57-30
PixPin_2026-05-09_23-59-28
PixPin_2026-05-09_23-59-34

PR #6508, opened 2026-08-13
Fix unreadable version diffs for unrelated text

### LINKED ISSUE #4788: Note history diff comparisons are broken
The diffs in note history doesn't appear to properly show what was removed and what was added now. Best demonstrated by [this post.](https://danbooru.donmai.us/posts/4370005)

demo

NekoAria:
This PR supersedes #6418 with a rewrite that fixes #4788 while preserving detailed diffs for related edits.

- Render name changes by Unicode grapheme cluster.
- Tokenize body text semantically and keep common surrounding context.
- Replace unrelated changed sections as a whole instead of matching incidental characters.
- Bound `Diff::LCS` work and use a linear fallback for oversized changes.
- Escape all input consistently, including strings already marked as HTML safe.

Differences from #6418:

#6418 adds CJK-specific token classes and a full-string Levenshtein pass before running LCS. The CJK classes retain the existing character-level behavior, while Levenshtein adds quadratic work.

This implementation instead:

- uses language-independent grapheme and content coverage rules;
- runs LCS at most once and only within deterministic resource limits;
- derives relevance from the resulting ordered matches;
- preserves formatting-only diffs without LCS when over budget;
- exposes only explicit `name_html` and `body_html` operations.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -21,17 +21,16 @@ module ApplicationHelper
   end
 
   def diff_name_html(this_name, other_name)
-    DiffBuilder.new(this_name, other_name, /./).build
+    DiffBuilder.new(old_text: other_name, new_text: this_name).name_html
   end
 
   def diff_body_html(record, other, field)
     if record.blank? || other.blank?
-      diff_record = other.presence || record
-      return h(diff_record[field]).gsub(/\r?\n/, '<span class="paragraph-mark">¶</span><br>').html_safe
+      body = (other.presence || record)&.[](field)
+      return DiffBuilder.new(old_text: body, new_text: body).body_html
     end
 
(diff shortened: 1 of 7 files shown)


nonamethanks:
This pull is basically too complex to understand. The code is unreadable by a human. It needs docstrings everywhere, and additional comments explaining why the most complex passages are the way they are, otherwise it's unmaintenable.

There's a lot of questionable choices like `compare_content` running before `within_lcs_budget?` despite not being used inside that block, and that's just at a glance, but I can't go further because nothing in the code is explained. How do you expect anyone to be able to review this?

NekoAria:
Thanks for the review.

I simplified the flow, added clarifying comments, and fixed the rendering bug that could place inserted text inside a `<del>` block.

PR #6555, opened 2026-08-24
BURs: allow users to filter by whether they can approve

nonamethanks:


app/logical/bulk_update_request/command/create_alias.rb
```diff
@@ -42,7 +42,10 @@ def new_tag
     @new_tag ||= Tag.find_by_name(@new_name)
   end
 
-  def approval_level
+  def approval_level(tags: nil)
+    old_tag = tags.present? ? tags.find { |tag| tag.name == @old_name } : old_tag()
+    new_tag = tags.present? ? tags.find { |tag| tag.name == @new_name } : new_tag()
```
evazion: `old_tag()` is bad style in Ruby, it would be `self.old_tag`. Even better would be to pass the `tags` list in the constructor then have the `old_tag` method do this lookup itself. Or better yet, pass in a reference to the BUR itself in the constructor when building the command, so the command can do `bur.tags` to get the list of tags in the BUR.

app/models/bulk_update_request.rb
```diff
@@ -47,8 +47,20 @@ module SearchMethods
...
+      statuses = params[:status].to_s.split(",")
+
+      if statuses.present?
+        q = q.where(status: statuses)
+      end
+
+      if params[:can_approve].present? && statuses.present? && statuses.exclude?("pending")
+        q = q.none
+      elsif params[:can_approve].present?
+        can_approve = ActiveModel::Type::Boolean.new.cast(params[:can_approve])
```
evazion: Should use our own `params[:can_approve].to_s.truthy?` here instead of `ActiveModel::Type::Boolean.new.cast`

PR #6560, opened 2026-08-25
Fix #6099: allow editing disapprovals

### LINKED ISSUE #6099: posts: allow editing disapprovals
A side effect of cbeeb9d is that it is no longer possible to edit disapprovals by selecting a different reason or using the detailed reason form on the post page.

ghost:
Changes disapprovals to update the existing one if you try to make a new one. This means you can just use the regular mod buttons to change your disapproval reason. I didn't change the UI of the mod buttons, so the "Skip" button will still be disabled if you have disapproved the post already, otherwise it wouldn't be clear whether you've disapproved it or not. You can still change from a non-disinterest to disinterest via the detailed rejection UI.

I opted to do this instead of calling PUT /post_disapprovals/:id because it simplifies the template logic.

### DIFF
--- a/app/controllers/post_disapprovals_controller.rb
+++ b/app/controllers/post_disapprovals_controller.rb
@@ -24,8 +24,8 @@ class PostDisapprovalsController < ApplicationController
   end
 
   def create
-    @post_disapproval = authorize PostDisapproval.new(user: CurrentUser.user, **permitted_attributes(PostDisapproval))
-    @post_disapproval.save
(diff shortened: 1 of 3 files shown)


nonamethanks:
Reminder to self: one thing that appears to be broken is the filling in if you've already disapproved, but it was already broken before this commit.

ghost:
Do you mean pre-populating the rejection dialog box with the current reason/message?

nonamethanks:
Yeah, it's a bug I noticed while testing this PR. I'll fix it when I merge this, it was just a note so that I don't forget.

PR #6578, opened 2026-08-28
Users: add superadmin level

nonamethanks:
Superadmins are a step above admins: they can perform dangerous and potentially irreversible database actions like user deactivation, and sensitive actions such as viewing all dmails and private favorite groups. They can also promote and demote other admins.

This is a role for site staff that already has some degree of access to destructive backend functions and analytics. In the future, this role may be able to perform backend actions such as launching database migrations or metadata regeneration scripts from the site interface, so downstream boorus should be extremely conservative in assigning this role.

Site owners retain the exclusive ability to handle legally sensitive actions, such as those related to DMCA processing or payments/refunds.


nonamethanks:
If merged this would make pull #6577 be a superadmin-only function.

PR #6586, opened 2026-08-31
Tests: refactor e2e tests

### LINKED ISSUE #6584: Implement proper system tests
- [ ] if we don't want browser binaries due to docker image size inflation, add a script to install them on demand/to spin up selenium containers on demand (for both devbooru and on the github test actions) 
- [ ] test on both Firefox and Chrome latest versions. Possibly on last supported versions too? 
- [ ] Also need to test for mobile pages, maybe just by reducing viewport & pointer type (touch) 
- [ ] investigate using playwright webkit for safari testing 
- [ ] test most `onclick:` stuff

### LINKED ISSUE #6589: Move storage manager tests to docker compose
Those tests are really flaky and slow because they bring up a docker container for every test. 

Github actions support custom commands for services now (since ~2025), so we should just change them to rely on an existing container, like it's already being done for pull #6586.

nonamethanks:
Very early WIP. 

Replaces selenium with playwright, which allows us to test Chrome, Firefox and Safari at the same time with a single extra container.

The container is optional in the dev environment and started with `bin/dev --profile test up`.

These tests are ancient, and they never worked reliably (at least for me), so we're not actually testing anything meaningful at the moment aside from maybe the autocomplete, which does not even have extensive tests.

Will need a lot of work to test pretty much every dropdown, modal, shortcut and javascript function. 

Playwright has [experimental support for android](https://playwright.dev/docs/api/class-android) so we might be able to test mobile more meaningfully than just changing viewport too.

Fixes #6584, fixes #6589.


nonamethanks:
Should be finished now. 

I opted not to make them by default on github. They are simply too flaky. They could be made more reliable, but it would take a lot of work, which would be better spent on other tasks.
They can be easily run locally now, just do `bin/dev --profile test up -d` and it'll bring up the optional containers for the full test suite: playwright and sftp. The latter is for proper storage tests, which were failing way too often before because they were bringing up a docker container up and down at every step.


I also changed the push pipeline because I noticed that the build cache was polluting the danbooru package. Now they get pushed to a separate package, to reduce noise.

